PR #22, opened 2016-05-19
Don't hardcode screenshot directory (still hardcodes filename)

rcrit:
Obviously it would be nice to generate a unique name for the screenshot but this at least removes the hardcoded directory.


danhedron:
Thanks! It would be nice if this were cross platform, but there are a few other places where unix is assumed so this can be cleaned up along with them later.

PR #33, opened 2016-05-20
OS X port of OpenRW

christoph-heiss:
Hi,
beside some small compilation errors, this also fixes `findPathRealCase` in `rwengine/src/engine/GameData.cpp` on OS X.

On a side note, it does not correctly render yet. (first of, the shaders do not even compile because of core/compatibility profile issues on OS X)

(This was tested on 10.11.5)


danhedron:
Thanks for this. I'm keen to get a GL 2 path working for OS X support.

PR #34, opened 2016-05-21
Feature compile under mac

Hockenberry:
More changes to compile under macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,15 +48,20 @@ ENDIF()
 IF(${ENABLE_SCRIPT_DEBUG})
 	add_definitions(-DSCM_DEBUG_INSTRUCTIONS)
 ENDIF()
-
-# Find dependancies
-IF(APPLE)
-	set(OPENRW_PLATFORM_LIBS iconv)
-ENDIF()
 find_package(OpenGL REQUIRED)
 find_package(Bullet REQUIRED)
 find_package(SFML 2 COMPONENTS system window audio graphics network REQUIRED)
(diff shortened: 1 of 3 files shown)


christoph-heiss:
Regarding the change in `rwgame/GameConfig.cpp`:
`$HOME/.config` does/should not exist in OS X (some applications use it anyway, but that's just wrong).

In OS X, configuration files (usually .plist's) are stored in `$HOME/Library/Preferences/<appname>`,
not `$HOME/.config/<appname>` like under Linux.

I forgot yesterday to include [this change](https://github.com/christoph-heiss/openrw/commit/5e6a422fa0d12eef8ba0de1e76ea1b4fde0be451) in my previous PR.

That's just my two cents.

CMakeLists.txt
```diff
@@ -48,15 +48,20 @@ ENDIF()
...
+find_package(SFML 2 COMPONENTS system window audio graphics network REQUIRED)
+find_package(MAD REQUIRED)
+find_package(GLM REQUIRED)
+
+
 
 # Find dependancies
 IF(APPLE)
 	set(OPENRW_PLATFORM_LIBS iconv)
+	include_directories("${GLM_INCLUDE_DIRS}")
```
danhedron: This is technically required on all platforms, It would be best to move this line after the find_package directives and do it on all platforms.

danhedron:
Thanks for this!

As @christoph-heiss says it would be nice to use the correct path on OS X.
What does the .clang-format file do?

Issue #41, opened 2016-05-21
Make fails with 'Could not find a package configuration file provided by "GLM"'


TacoDeBoss:
As of this hour, compilation now fails on my Linux system, something relating to GLM. libglm-dev is installed, but `make` results in

-- Found SFML 2.1.0 in /usr/include
-- Found MAD in /usr/include
CMake Error at CMakeLists.txt:61 (find_package):
  By not providing "FindGLM.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "GLM", but
  CMake did not find one.

  Could not find a package configuration file provided by "GLM" with any of
  the following names:

    GLMConfig.cmake
    glm-config.cmake

  Add the installation prefix of "GLM" to CMAKE_PREFIX_PATH or set "GLM_DIR"
  to a directory containing one of the above files.  If "GLM" provides a
  separate development package or SDK, be sure it has been installed.

-- Configuring incomplete, errors occurred!
See also "/home/taco/stuff/openrw-git/openrw/build/CMakeFiles/CMakeOutput.log".
make: *** [cmake_check_build_system] Error 1

danhedron:
Looks like FindGLM was deprecated and replaced by the new packaging system that I'm not familiar with in https://github.com/g-truc/glm/commit/62a7daddcf082f754000fc5e42d7bcdf93c895f7

TacoDeBoss:
I fixed it by copying a FindGLM.cmake into the /cmake/modules folder.
I found the FindGLM.cmake at https://github.com/lighttransport/nanogi/blob/master/cmake/FindGLM.cmake
So I'm good, but what about others having this problem?

TacoDeBoss:
From the terminal output before FindGLM.cmake was placed
`Add the installation prefix of "GLM" to CMAKE_PREFIX_PATH or set "GLM_DIR"
  to a directory containing one of the above files.  If "GLM" provides a
  separate development package or SDK, be sure it has been installed`
I'm honestly not a programmer, but if we can direct whatever cmake file to /usr/include/glm as the GLM directory, it would replace the need for the FindGLM file, yes?
Apologies for not being able to help to a greater degree.

neuromancer:
> So I'm good, but what about others having this problem?

I had the same problem. Your workaround solved it.

PR #42, opened 2016-05-21
Add FindGLM.cmake to prevent make error

TacoDeBoss:
We can add FindGLM.cmake to close issue #41


LeonardKoenig:
The better way to handle this is to use Config mode. PRQ here: https://github.com/danhedron/openrw/pull/43

TacoDeBoss:
@LeonardKoenig Hey, I'm not even a programmer, so if you say so.. Haha

danhedron:
Handled by #43

PR #43, opened 2016-05-21
[CMake] Use find_package config mode for GLM

LeonardKoenig:
GLM requires using the new config mode as it only ships a
glmConfig.cmake.

more info:
http://stackoverflow.com/q/34634850/4453524
https://cmake.org/cmake/help/latest/command/find_package.html

fixes https://github.com/danhedron/openrw/issues/41

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ find_package(OpenGL REQUIRED)
 find_package(Bullet REQUIRED)
 find_package(SFML 2 COMPONENTS system window audio graphics network REQUIRED)
 find_package(MAD REQUIRED)
-find_package(GLM REQUIRED)
+find_package(glm REQUIRED CONFIG)
 
 include_directories("${GLM_INCLUDE_DIRS}")
 


LeonardKoenig:
I've also added Config mode for Bullet, since it does support it too.

LeonardKoenig:
We could also remove `cmake/modules/FindSFML.cmake` as it is already provided by upstream:
https://github.com/SFML/SFML/blob/master/CMakeLists.txt#L297

LeonardKoenig:
Just noticed that Bullet, although providing a Config cmake seems not to build properly that way. I'll investigate.

LeonardKoenig:
Okay, deleted the commit. `FindBullet.cmake` is already provided by CMake itself, so we gonna use that. The Config way it doesn't set up include paths properly.

Hockenberry:
Using patch (fe0c93f) under ubuntu 16.04 LTS does work with a glm rebuild from source.

danhedron:
Thanks.

PR #66, opened 2016-05-23
Added how to install Dependencies on ubuntu to readme

Korons:
I added a oneliner to install the needed Dependencies on ubuntu so people wouldn't have to look around for them

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ Dependencies:
 * libmad
 * Boost Test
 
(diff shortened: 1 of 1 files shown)


danhedron:
Thanks, but we're moving the instructions onto the wiki where they're easier to edit: [Getting Started](https://github.com/rwengine/openrw/wiki/Getting-Started)
We should have the readme updated shortly.

PR #86, opened 2016-05-25
Add FreeBSD support

AMDmi3:
For now, there's only single OS-dependent bit of code, it it should be handled on FreeBSD just like on Linux. While here, change macro testing from #if XXX to #if defined(XXX), this is clener and not prone to
"undefined macro" errors


AMDmi3:
FYI, I was able to compile it with another little change (added libiconv to target_link_libraries; not PR yet as it needs to be properly detected first), but haven't runtested it yet.

danhedron:
Thanks. I am a fan of the more explicit ``#if XXX`` but as platforms get added this is easier to manage.

PR #87, opened 2016-05-25
Use inih as git submodule

AMDmi3:
ExternalProject use is discouraged.
- Most package build systems do not allow any fetching on
  configure/build stage
- It's insecure the way it's used now, as it does not specify commit
  hash, so it may fetch whatever malicous code and it'll come
  unoticed

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "external/inih"]
+	path = external/inih
+	url = https://github.com/benhoyt/inih.git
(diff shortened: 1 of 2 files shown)


danhedron:
Thanks, much better.

parkerlreed:
Is this going to require doing a git submodule init when we clone?

AMDmi3:
> Is this going to require doing a git submodule init when we clone?

Yes. I myself usually add a check to CMakeLists.txt to ensure that submodule directory is non-empty and suggest a user to to ```git submodule init && git submodule update``` otherwise.

PR #104, opened 2016-05-29
Attempt to remove compiler warning //??

SamStudio8:


paulsapps:
What about checking RW_CHECK to do the RW_UNUSED in release/whatever option causes it to do nothing?

rwengine/src/objects/PickupObject.cpp
```diff
@@ -89,8 +89,10 @@ PickupObject::PickupObject(GameWorld *world, const glm::vec3 &position, int mode
 	m_corona->particle.texture = engine->data->findTexture("coronacircle");
 
 	auto flags = behaviourFlags(m_type);
+	RW_UNUSED(flags);
+	
 	RW_CHECK((flags & PickupInVehicle)==0, "In Vehicle pickup not implemented yet");
```
Hockenberry: Something like:

`RW_CHECK((behaviourFlags(m_type) & PickupInVehicle)==0, "In Vehicle pickup not implemented yet");`

JayFoxRox:
Pretty bikesheddish.

Either solution is ugly for reasons (1. it's only unused without RW_CHECK, 2. it hides the missing feature behind logic).
However, I guess this could be squash-merged just so we don't keep this PR dangling forever.

paulsapps:
Maybe RW_UNIMPLEMENTED that gets compiled in debug/release? Easy to search for and will kill the warnings.

JayFoxRox:
Good idea. yeah.. just use

```
if (flags & PickupInVehicle == 0) {
  RW_UNIMPLEMENTED("PickupObject if PickupInVehicle is set");
}
```

[Eventually we should probably attach the function name in the macro (which also still misses a space). It's a different issue though]

PR #108, opened 2016-05-29
Add function leaveVehicle in CharacterObject

aldoxtor:
I added function leaveVehicle in CharacterObject.


JayFoxRox:
Kind of unecessary imho.
I think EnterVehicle should be renamed to SetVehicle or something, but other than that I don't see why we should change it / add a new function.

You let characters enter and exit vehicles through activities anyway.

Other possible names btw: `exitVehicle`, `ejectFromVehicle`

danhedron:
I agree with @JayFoxRox that the existing function should just be renamed to `setVehicle` over `enterVehicle`.

danhedron:
Thanks for raising this. You can open another PR to re-name this method.

PR #111, opened 2016-05-30
Replace SFML Audio with libsndfile + OpenAL

tsjost:
This is a "quick fix" to make the game not completely break for many people. The entire audio system will have to be reworked soon.
Let there be SDL now!


rwengine/src/audio/MADStream.cpp
```diff
@@ -58,6 +58,7 @@ mad_flow MADStream::ms_output(void* user, mad_header const* header, mad_pcm* pcm
 	if ( ! self->numFreeBuffers) {
 		ALint buffersProcessed;
 		do {
+			std::this_thread::sleep_for(std::chrono::milliseconds(20));
```
JayFoxRox: Should add a note here why this is necessary

JayFoxRox: Should add a note here why this is necessary

JayFoxRox:
@rwengine: Reminder that rwengine/openrw/ still has a fix/audio branch which can probably be deleted now?
Slightly related there is also an outdated rwng-deferred-rendering which should be moved to a private repo ( @danhedron ? ) too.

tsjost:
Thanks, fixed! The deferred rendering is mine; I guess I'll have to rebase it before putting it in my repo.

PR #119, opened 2016-06-01
Vehicle Physics Improvements [do not merge]

danhedron:
This cleans up some usage of bullet within vehicles and instances, and introduces a new vehicle physics implementation in VehicleDynamics.

VehicleDynamics replaces btRaycastVehicle giving finer control over the behaviour. It's still far from being correct and seems way too bouncy, but we can iterate on this with something to compare it against.

To Fix:
- [ ] Everything is wobbly for some people
- [ ] Exit a vehicle & accelerate, stop accelerating -> brakes will still be applied when entering.

### DIFF
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -1,6 +1,7 @@
 #include <objects/VehicleObject.hpp>
 #include <objects/CharacterObject.hpp>
 #include <engine/GameWorld.hpp>
+#include <dynamics/CollisionInstance.hpp>
 #include <BulletDynamics/Vehicle/btRaycastVehicle.h>
 #include <dynamics/RaycastCallbacks.hpp>
 #include <data/CollisionModel.hpp>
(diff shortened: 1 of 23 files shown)


JayFoxRox:
I'd like to test this before merging.

I think if the original didn't use 'ackermann steering' we should also hold back on adding that.
Sorry for bringing it up in the first place :(

Did you somehow test this implementation against behaviour of the game? Should we?
(I think this should give us more control than btRayCast so it's probably better than before; it might also be too early to get vehicle physics right, but we should have an issue about it and try to match the original as close as possible in the future)

danhedron:
I didn't do any quantitative testing. I'm hoping to tune it a little more before merging to get it at least qualitatively better than it was before, then doing some more analysis in future.

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -161,47 +118,78 @@ void VehicleObject::tickPhysics(float dt)
...
-		// todo: a real engine function
-		float velFac = info->handling.maxVelocity;
-		float engineForce = info->handling.acceleration * throttle * velFac;
-		if( fabs(engineForce) >= 0.001f ) physBody->activate(true);
+		float currentVelocity = (getVelocity()*60*60)/1000.f;
+		float velFac = 1.f;
+		if (currentVelocity >= info->handling.maxVelocity)
+		{
+			velFac = 0.f;
+		}
```
JayFoxRox: Should probably add a comment (@todo) here with a [link to cerbera who figured out that the max velocity is actually mostly depending on the model shape + acceleration instead.](http://projectcerbera.com/gta/3-vc/tutorials/acceleration)

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -29,6 +29,16 @@ VehicleObject::VehicleObject(GameWorld* engine, const glm::vec3& pos, const glm:
 {
 	collision = new CollisionInstance;
 	if( collision->createPhysicsBody(this, data->modelName, nullptr, &info->handling) ) {
+
+		// calculate the intertia from the handling information
+		auto& dim = info->handling.dimensions;
+		auto mass = info->handling.mass;
+		btVector3 handlingInertia(
+		    1.f / ((mass / 12) * (dim.y * dim.y + dim.z * dim.z)),
```
JayFoxRox: "12..! because why not 12..?"

In all honesty.. why 12?

danhedron: https://en.wikipedia.org/wiki/List_of_moments_of_inertia your guess is as good as mine :8ball: 

I'll leave a note.

danhedron:
Some of the changes in this PR are now in #158 . This PR will be re-based once those are in so this PR can focus on vehicle specific changes.

danhedron:
(Closing this for now since it needs to be re-based and more work)

Issue #107, opened 2016-05-29
Garages do not open


JayFoxRox:
There should be a garage class.

At least the following garages must be supported in order to be able to play the game:

* Pay'n'Spray; car drives in, player looses control, spraypaint effect happens, script can check if the specific garage just resprayed something (ResprayGarage)
* Import/Export; compares vehicle id to from a list and strikes through the entry on the door, sets a bit in a bitstring which is also stored in savegames. Script has access to said bitstring (IEGarage)
* Hideouts, any vehicle inside the garage is stored in the savegame (SaveGarage)
* Mission garages, controllable by script, can only open and close / no special behaviour (StandardGarage)
* Bombshop Garage, similar to respray but installs a specific type of bomb in the car (BombshopGarage, maybe we should have a class ShopGarage for BombshopGarage and ResprayGarage which takes control away from the player etc once inside the garage?)

[You can find a full list of garge types here](http://www.gtamodding.com/wiki/Garage#GTA_III)

It's currently not known how the game knows which object is the garage door.
The garages are created from the script or a savegame. A garage is created as a specific type, and each type can have multiple garages associated with it.
The script also controls how each garage opens (swing door / slide door). It's a miracle how the game can know which garage number (which is independent from the type!) controls which door object.

[There is a list which links different garages to their respective door model](http://www.gtamodding.com/wiki/Garage#GTA_III_2). However, each door model appears multiple times and the game somehow figures out which door is the correct one.

It's not known if the garage door objects might be hardcoded independent of the garages.
We should test if garage doors open, even if no garage is created from the script / savegame.

From the script the garages are identified by a number (the order they are created in the SCM, starting at index 1, index 0 is some unused (broken?) garage I believe). The index of each garage is then stored in the global variables listed on the list above.

---

It's not known if the garage code also handles opening / closing the sliding fences / gates or if that is done by the script.
Example: The cartel sliding gate does not exist in the OpenRW test mode, and seems to be created by the script. Yet, it does not seem to open properly (tried parking a yakuza stinger and cartel cruiser in front)?

Standard doors (such as the one to mistys house) are created as instance object by script or savegame.
They are opened and closed by setting the heading / z_angle using the script.

JayFoxRox:
I'm trying to reach @spaceeinstein at the moment [as he said I misinterpreted some of the garage stuff of his](http://gtaforums.com/topic/856343-openrw-engine-not-to-be-confused-with-reditecs-tool/?p=1068810163).
[...]

PR #120, opened 2016-06-02
Basic garage functionality in script

JayFoxRox:
I've confirmed that the game handles garages using indices from 0 to $garageCount-1 (value returned from `create_garage`).
In my savegames, there is one additional garage in slot 0 which is probably created (or controlled) by the [crusher crane from opcode 02FB](http://www.gtamodding.com/wiki/02FB) (The gtamodding wiki article [2nd of june '16] doesn't say that a garage is created; I think this is the major difference to [opcode 01EE](http://www.gtamodding.com/wiki/01EE) ).

It's quite possible that 02FB only controls the crusher garage or that the garage is hardcoded to element 0 anyway. However, my implementation (02FB creating the crusher garage) should work fine with the standard mission script from GTA3 which only contains 1 crusher anyway.
As garages + global variables (which hold garage handles / indices) are both part of the savegame this shouldn't be able to break anything either.
If there are any problems in the future we can still change it.

Besides crusher crane garage creation this also adds 2 opcodes to check if a specific car is in a garage and another one to check if any car is inside a garage (presumably). These functions only check if the car hits the bounding box of the garage to be considered "inside garage" - consider it a stub.
Another opcode stub (to only open garages for a certain car) was added because we'll need it in LUIGI2 soon anyway and it was mission garage related.

(As a hack) the garages can be "opened" by a debug menu function as we don't have generic code for handling garages (and their doors) yet. Also the rules for opening garages and finding garage doors are not really known yet (to us).
@spaceeinstein seems to know more about this but I think some reverse engineering still has to be done to get them right.

---

I think the opcodes could be named better + they are pretty hacky.
However, this is enough to stub them for now.
I added '@todo's and we should probably verify the behaviour in the future.

We'll also need seperate code / class for the garages in the future as requested per #107 .
I'll try to get more details about that before implementing that (sometime in the future) though.

---

I consider this PR ready for merge


rwgame/debugstate.cpp
```diff
@@ -129,6 +130,53 @@ DebugState::DebugState(RWGame* game, const glm::vec3& vp, const glm::quat& vd)
 	m->addEntry(Menu::lambda("Cull Here", [=] {
 		game->getRenderer()->setCullOverride(true, _debugCam);
 	}, entryHeight));
+	m->addEntry(Menu::lambda("Unsolid garage doors", [=] {
+
+		std::vector<std::string> garageDoorModels {
```
paulsapps: you could probably use const static std::array of const char* here to avoid doing memory allocations/for a bit more performance (if it matters)

JayFoxRox: It's only a debug feature (called only if you choose to do so) and std::array doesn't support initializing from an unknown number of elements afaik?
The compiler will most certainly optimize everything of that away anyway.

paulsapps: Not sure if it can figure out N or not - if its debug then I guess its moot point either way :)

rwengine/src/objects/InstanceObject.cpp
```diff
@@ -174,4 +174,20 @@ bool InstanceObject::takeDamage(const GameObject::DamageInfo& dmg)
 	return false;
 }
 
+void InstanceObject::setSolid(bool solid)
+{
+	// Early out in case we don't have a collision body
+	if (body == nullptr || body->body == nullptr) {
```
danhedron: body is not public in the physics changes, I'll rebase & update this.

danhedron:
Looks good to me. +1.

PR #125, opened 2016-06-03
Replace SFML with SDL2

christoph-heiss:
Hi!

So, I *think* this should be ready!
I rebased and recommited all changes on top of master (original history can be found at the `sdl-gl33-port` branch).

**But, before merging, please test it somewhat extensively!**

I noticed a weird bug while testing, where the game basically goes into an slowmo-like state, but it occurs very, very rarely (it has yet to occur a third time).

I could not find out why this is/was happening, although I could imagine it has something to do with commit f6bf1f066c30f0c5c0aa79b002744493809d2205.

Everything else should work fine.

Christoph


CMakeLists.txt
```diff
@@ -67,13 +67,19 @@ ENDIF()
 find_package(OpenGL REQUIRED)
 find_package(OpenAL REQUIRED)
 find_package(Bullet REQUIRED)
-find_package(SFML 2 COMPONENTS system window graphics network REQUIRED)
 find_package(MAD REQUIRED)
 find_package(GLM REQUIRED)
 find_package(LibSndFile REQUIRED)
+find_package(SDL2 REQUIRED)
```
danhedron: Just to note: Ubuntu 16.04 includes sdl2-config.cmake but earlier releases don't, they'll need to build SDL from source.

rwgame/GameWindow.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+	auto size = getSize();
+	png_set_IHDR(png_ptr, info_ptr, size.x, size.y, 8, PNG_COLOR_TYPE_RGB,
+		PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_BASE, PNG_FILTER_TYPE_BASE);
+
+	png_write_info(png_ptr, info_ptr);
+
+	const unsigned char* buffer = renderer->getRenderer()->readPixels(size);
+	for (int y = size.y-1; y >= 0; y--)
+		png_write_row(png_ptr, &buffer[y * size.x * 3]);
```
danhedron: Ubuntu 14.04:
```
/home/dan/dev/openrw/rwgame/GameWindow.cpp: In member function ‘void GameWindow::captureToFile(const string&, GameRenderer*)’:
/home/dan/dev/openrw/rwgame/GameWindow.cpp:86:48: error: invalid conversion from ‘const unsigned char*’ to ‘png_bytep {aka unsigned char*}’ [-fpermissive]
   png_write_row(png_ptr, &buffer[y * size.x * 3]);
```

parkerlreed:
I have had the slow-mo happen to me on that SDL port at least a few times.

danhedron:
Upgrading libpng resolves the compilation error. 

The camera in the debug mode (accessible via the M key) seems to be taking the absolute position as relative.

rwgame/State.cpp
```diff
@@ -4,6 +4,39 @@
...
+	case SDL_MOUSEBUTTONUP:
+		if (e.button.button == SDL_BUTTON_LEFT)
+			m->click(e.button.x, e.button.y);
+		break;
+
+	case SDL_MOUSEMOTION:
+		m->hover(e.motion.x, e.motion.y);
+		break;
+
+	case SDL_KEYUP:
```
tsjost: This makes it no longer possible to hold the arrow keys to navigate the menus quickly. Is there any particular reason to not use KEYDOWN?

README.md
```diff
@@ -43,3 +43,7 @@ SFML is licensed under the zlib license
 Bullet Physics is licensed under the zlib license
 
 * http://bulletphysics.org/mediawiki-1.5.8/index.php/LICENSE
+
```
JayFoxRox: You forgot to remove SFML from the README?

rwgame/debug/TcpSocket.cpp
```diff
@@ -0,0 +1,67 @@
...
+	client.sock = accept(sock, reinterpret_cast<sockaddr*>(&client.addr), &len);
+}
+
+
+void TcpSocket::recv(std::string& out, size_t len)
+{
+	std::vector<char> buffer(len);
+	read(sock, buffer.data(), len);
+
+	out = { buffer.begin(), buffer.end() };
```
JayFoxRox: `/// @todo must check how many bytes were actually received`

rwgame/debug/TcpSocket.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+void TcpSocket::send(const std::string& str)
+{
+	write(sock, str.c_str(), str.length());
```
JayFoxRox: `/// @todo must check how many bytes were actually sent`

rwgame/debug/TcpSocket.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+void TcpSocket::bind(short port)
+{
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = htonl(INADDR_ANY);
+	addr.sin_port = htons(port);
+
+	::bind(sock, reinterpret_cast<sockaddr*>(&addr), sizeof(addr));
```
JayFoxRox: `/// @todo must check if socket exists or check for errors in general`

rwgame/debug/TcpSocket.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void TcpSocket::send(const std::string& str)
+{
+	write(sock, str.c_str(), str.length());
+}
+
+
+void TcpSocket::disconnect()
+{
+	close(sock);
```
JayFoxRox: `/// @todo must check if socket exists or check for errors in general`

also you might want to set `sock = -1` again or something?

rwgame/debugstate.cpp
```diff
@@ -147,6 +147,7 @@ DebugState::DebugState(RWGame* game, const glm::vec3& vp, const glm::quat& vd)
 
 void DebugState::enter()
 {
+	getWindow().hideCursor();
```
JayFoxRox: actually I think we'd like to show the mouse cursor in DebugState unless the free flight Camera is in use?
[should be a function somewhere else: make the mouse cursor invisible and center it OR show the cursor but release it]

christoph-heiss:
Hi, please review the changes!

I reworked the changes to include all your suggestion, I hope everything is fine now.
Also, I snuck in a commit for fixing a build regression on OS X.

(@JayFoxRox: The error handling I added is somewhat sneaky. I purely relies upon that passing `-1` as sock will result in `errno == EBADF`)

Christoph

JayFoxRox:
re errno: I guess that's okay for now but in the long term we should avoid errors by checking / fixing inputs.

I've just compiled and run this and it seems to work fine.
However, one major problem I have with this that the mouse speed / sensitivity is *FAR* lower than with SMFL. I only have a ThinkPad trackpoint to use and that makes it nearly impossible for me to navigate.

There are also a handful of other nits left, but I think this is another case of "merge-first, fix later".

I only played a bit of the first mission.
\- I did not test the mousewheel stuff, did not pay attention to when the cursor is visible and also did not test the http server (which I never used before either).

(I also had the feeling the camera position was slightly different, but maybe I'm also just a little too sleepy :sleeping:  )

---

If you fix the mouse thing this gets LGTM from me.

rwgame/debugstate.cpp
```diff
@@ -235,27 +237,10 @@ void DebugState::tick(float dt)
 		}
 	}*/
 
-	if( _freeLook ) {
```
JayFoxRox: Add this check back please.
The debug menu is spinning around even if freelook is not selected.

JayFoxRox:
Tested some more:

(In windowed mode) it takes 6 (!) seconds to do a full 360 degree spin in the debugstate freelook if I pull my trackpoint as hard as possible. around 4 seconds to do it on foot to spin the player.

Also if possible the resize and maximize buttons should return before this is merged (or very shortly afterwards, but I don't like rushed hotfixes) because it makes debugging way harder. The window is too large for my small monitor so I can hardly see gdb.

rwgame/GameWindow.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+
+}
+
+
+void GameWindow::create(size_t w, size_t h, bool fullscreen)
+{
+	uint32_t style = SDL_WINDOW_OPENGL/* | SDL_WINDOW_ALLOW_HIGHDPI*/;
+	if (fullscreen)
+		style |= SDL_WINDOW_FULLSCREEN;
```
danhedron: The default style should include `SDL_WINDOW_RESIZABLE`, since that is very useful for debugging (and there's no technical reason why we can't resize the window).

`SDL_WINDOW_FULLSCREEN_DESKTOP` would be nice to use also, but that will mean the window is always at desktop resolution, we'll need some internal framebuffer scaling to match the requested resolution, so this can be done later.

JayFoxRox: While you are at it: `/* | SDL_WINDOW_ALLOW_HIGHDPI*/` dead code? Why was this there in the first place?

If you want to keep it: document why this is commented out and when it should be reintroduced. (Even in that case, just throw it away and create a issue about it if it's actually important)
If you don't want to keep it: throw it away.

JayFoxRox:
Hi, what's the status?

Remaining issues:

- [Insanely slow mouse sensitivity (for me only?)](https://github.com/rwengine/openrw/pull/125#issuecomment-225381829) In the worst case we'd merge first, fix it later
- [Requires `SDL_WINDOW_RESIZABLE`](https://github.com/rwengine/openrw/pull/125#discussion_r66719725)
- [Why `/* | SDL_WINDOW_ALLOW_HIGHDPI*/` ?](https://github.com/rwengine/openrw/pull/125#discussion_r66719787)
- [Freelook check](https://github.com/rwengine/openrw/pull/125#discussion_r66709738)
- Needs rebase


I did not review the socket error handling yet. I'd also prefer that in a seperate PR too as it's an entirely different feature (hence my remark making it a `@todo`). Transitioning context creation is risky as is. However, error-handling can stay in this PR if you want, I'll probably review it then.

christoph-heiss:
First of, sorry for the big delays, but I'm kinda busy currently :/

Anyway, I addressed all remaining issues and rebased it again on `master`. (I hope I haven't overlooked anything!)
Please review!

JayFoxRox: Regarding your comment on 2fd3b685f50e73d51912daa9bede5452edb6966b:
It isn't a proper solution, but for now it should be enough. I also tested on 4K and spinning slows down considerably, but I don't think anyone really plays at 4K yet.
On the other side, if you remove the divide by `screenSize`, it has an just insanely high sensitivity.

I think this can be done later on by an extra PR.

tests/test_globals.hpp
```diff
@@ -60,7 +58,12 @@ class Global
 #endif
 
 	Global() {
-		wnd.create(sf::VideoMode(640, 360), "Testing");
+		if (SDL_Init(SDL_INIT_VIDEO) < 0)
+			throw std::runtime_error("Failed to initialize SDL2!");
+
+		window.create(800, 600, false);
```
JayFoxRox: nit: I think test mode should always run at 640x480 because that's what GTA3 seems to use as it's reference coordinate system?
(At least we should keep the intended 4:3 AR)

JayFoxRox:
I've tested this and all original issues seem to be fixed.
Mouse speed is still too slow for me but it's fast enough to be merged now. (I think I calculated a factor of about 20.0 instead of 2.5 some weeks ago on IRC; however, that would be way too fast for others - it seems to be an SDL / window environment issue)

---

I don't particularily like modifying the SDL event either but it's okay for now.
We'll probably replace the mouse sensitivity stuff soon again anyway

I skimmed over the code and it also LGTM.

danhedron:
This looks good to me too. The remaining issues can be fixed separately.

parkerlreed:
Compiling now to test

parkerlreed:
Loads and plays fine in Arch.

christoph-heiss:
Thanks for merging, dan!

gbudny:
@christoph-heiss

I compiled Openrw on Mac OS X 10.10.5, but it doesn't work.

Did you try to run it on this version of Mac OS X?

christoph-heiss:
@gbudny I'm currently on 10.11.5, so no.
But what do you mean by "not working"?

gbudny:
**@christoph-heiss**

I can see a window of openrw when I try to run it, but it is completely black:

https://gist.github.com/gbudny/908b506a106a73bfa7158217f27f5018

JayFoxRox:
What version of OSX are you running? @gbudny

gbudny:
**@JayFoxRox**

I use Mac OS X Yosemite - 10.10.5

PR #136, opened 2016-06-05
Add Contributing guidelines

danhedron:
These are the rules that have been on the wiki, opening this up for a bit of further input here before merging it.


JayFoxRox:
There should probably be something about "Testing OpenRW" / "Posting Issues" and "Documenting the original GTA3 behaviour".
However, some of that might also be wiki material.

These are all nits though.
- We should get this merged first, we can always fix this later

---

LGTM

PR #137, opened 2016-06-08
File loading improvements

haphzd:


JayFoxRox:
Nice work :)

There are various other places in the code where binary files (script for example) are opened without that flag, could you attempt to change all of them in your PR?

We should probably discuss wether we also want to open "*.dat" files as binary too because the stupid newline stuff windows does is just cancerous behaviour..
We'd never be guaranteed that our code is portable.

danhedron:
Is there any progress on this?

haphzd:
Yes, don't merge yet.

rwgame/RWGame.cpp
```diff
@@ -111,7 +111,6 @@ RWGame::RWGame(int argc, char* argv[])
 	data->loadIMG("/models/gta3");
 	//engine->data.loadIMG("/models/txd");
 	data->loadIMG("/anim/cuts");
-	data->loadTXD("/models/hud.txd");
```
JayFoxRox: might be a stupid question, but where is it loaded now?

danhedron: `GameData::load()` should load that file

JayFoxRox:
I don't really like that file abstraction we currently have (because we only seem to have one instance to look in every folder which (luckily!) works for GTA3). We'll need an issue for that soon.
Your current changes are fine though.

@haphzd: Can you still grep for the ifstream stuff and add `, std::ios::binary` everywhere so we can get this merged?

haphzd:
Enough for now, I think.

danhedron:
Looks good to me, tested locally and it works fine.

tests/test_SaveGame.cpp
```diff
@@ -55,7 +55,7 @@ BOOST_AUTO_TEST_CASE(test_load_game)
 	GameState state;
 	GameWorld world();
 	SCMOpcodes s;
-	auto file = Global::get().d->loadSCM("data/main.scm");
+	auto file = Global::get().d->loadSCM("main.scm");
```
JayFoxRox: Why is this part of this commit?

rwlib/source/platform/FileIndex.cpp
```diff
@@ -155,7 +155,7 @@ FileHandle FileIndex::openFile(const std::string& filename)
 	}
 	else
 	{
-		std::ifstream dfile(fsName.c_str());
+		std::ifstream dfile(fsName.c_str(), std::ios::binary);
```
JayFoxRox: Please merge this with the other ifstream changes.

JayFoxRox:
Tested fine locally. I just think the commit structure has to improve a little.
Especially the SCM change breaks tests for one commit when it's really easy to avoid it.

So if you could fix what I mentioned this LGTM. Thanks!

danhedron:
Thanks for fixing this. Looks good to me now.

rwengine/src/engine/GameData.cpp
```diff
@@ -341,22 +341,12 @@ void GameData::loadHandling(const std::string& path)
...
-
-	if(! f.is_open() ) return nullptr;
-
-	f.seekg(0, std::ios_base::end);
-	unsigned int size = f.tellg();
-	f.seekg(0);
-
-	char* buff = new char[size];
-	f.read(buff, size);
+	auto scm_h = openFile(name);
```
JayFoxRox: nit: Dan likes to camelCase variable names.

(as this is the only thing I can nit about this is fine to me, we'll probably catch it later when loading SCM from the proper path with a better IO abstraction)

JayFoxRox:
- [x] Game still works
- [x] Doesn't contain malware
- [x] Follows our code style

LGTM unless you still want to change anything @haphzd  :)

danhedron:
Thanks. I appreciate the work in getting things working on Windows.

Issue #132, opened 2016-06-04
Script: 032B (Create Weapon Pickup) is poorly stubbed


JayFoxRox:
```
bindUnimplemented( 0x032B, game_create_weapon_pickup, 7, "Create Weapon Pickup" );
```

Due to this code there is no exception when the opcode is called, however, this opcode expects a weapon pickup handle to be returned to args[6].globalInteger (which does not happen).
As no pickup is created the handle is likely invalid and will cause crashes in other opcodes (because of invalid pointers).

This should either be stubbed correctly OR this should be implemented.

(Note that we already support 0x0213 which can create weapons too?)

JayFoxRox:
This makes it impossible to enable 03DD (which was also not tested because of that).
It's used in LUIGI4 (Pump Action Pimp)

JayFoxRox:
This has been fixed by #147

PR #140, opened 2016-06-08
Partially Fix LUIGI4

JayFoxRox:
This partially fixes the mission LUIGI4 (Pump Action Pimp).
I consider it merge-ready.

I've had success finishing this mission by throwing grenades and molotov cocktails at the car.
(The car will disappear, the 2 gang members will die)

This is also affected by #132 which makes it impossible to enable 03DD (which was not tested because of that).

---

This is mostly stubbed. The car won't drive, gang members won't fight back.


JayFoxRox:
Causes `E [Script] Illegal Instruction 01cc encountered at offset 35c98 on thread LUIGI4` when trying to steal the car.
I'll probably fix this

JayFoxRox:
Rebased and fixed problem with stealing the "Diablo"

Issue #133, opened 2016-06-05
Openrw crashes after the first cutscene


gbudny:
Hi

I want to inform you that Openrw crashes after this cutscene:

https://www.youtube.com/watch?v=Mzbl6neJ4uo&list=PL7BCD02E7D8C6D564&index=2

00:57

```
#  gdb ./rwgame GNU gdb (Ubuntu 7.11-0ubuntu1) 7.11 Copyright (C) 2016 Free Software Foundation, Inc. License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html> This is free software: you are free to change and redistribute it. There is NO WARRANTY, to the extent permitted by law.  Type "show copying" and "show warranty" for details. This GDB was configured as "x86_64-linux-gnu". Type "show configuration" for configuration details. For bug reporting instructions, please see: <http://www.gnu.org/software/gdb/bugs/>. Find the GDB manual and other documentation resources online at: <http://www.gnu.org/software/gdb/documentation/>. For help, type "help". Type "apropos word" to search for commands related to "word"...```
Reading symbols from ./rwgame...(no debugging symbols found)...done.
(gdb) run
Starting program: /home/gbudny/Desktop/openrw-build/rwgame/rwgame 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffefa54700 (LWP 1700)]
Warning: The created OpenGL context does not fully meet the settings that were requested
Requested: version = 1.1 ; depth bits = 32 ; stencil bits = 8 ; AA level = 0 ; core = false ; debug = false
Created: version = 3.3 ; depth bits = 24 ; stencil bits = 8 ; AA level = 0 ; core = false ; debug = false
I [Game] Game directory: /home/gbudny/GTAIII
Max UBO Size: 65536
UBO Alignment: 256
Max batch size: 512
I [Renderer] OpenGL Renderer Version: 3.3.0 NVIDIA 340.96 (GLSL 3.30 NVIDIA via Cg compiler) Vendor: NVIDIA Corporation
[New Thread 0x7fffeb6ec700 (LWP 1704)]
[Thread 0x7fffeb6ec700 (LWP 1704) exited]
[New Thread 0x7fffeb6ec700 (LWP 1705)]
[New Thread 0x7fffe6eea700 (LWP 1706)]
[...]

PR #141, opened 2016-06-09
Fix opcodes 0121 & 0395

danhedron:


JayFoxRox:
I think it's not the player itself which is ignored by `clear_area` but script chars in general.
However, this'll have to do for now. We can figure out the details later.

Tested and working. Fixes #133. Also quickly skimmed over the code.

LGTM

PR #142, opened 2016-06-14
Implement stealing vehicle

danhedron:
This change adds the ability to forcibly eject characters from vehicles.

### DIFF
--- a/rwengine/include/ai/CharacterController.hpp
+++ b/rwengine/include/ai/CharacterController.hpp
@@ -190,7 +190,10 @@ namespace Activities {
 	struct ExitVehicle : public CharacterController::Activity {
 		DECL_ACTIVITY( ExitVehicle )
 
-		ExitVehicle( )
+		const bool jacked;
+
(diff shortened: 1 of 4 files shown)


JayFoxRox:
There is a bug if you press "F" again while already trying to enter the car. The player will spin to where the camera is looking and repeat the animation, eventually freezing until you jump away.
Except for that, this looks good to me.

Once this is merged we should create a new issue about missing steal features:
- Quickjack is not implemented
- It's not possible to cancel stealing at the moment (I think?)
- Stealing from the passenger side is not working (also needs the shuffle animation if not entering as passenger)
- Stealing for boats not tested
- Stealing for special animation vehicles (such as the "Coach") not tested

[This is a very good source of informations](http://gtaforums.com/topic/847721-forcing-specific-car-jacking-animation/) which should be added to the issue.

danhedron:
Looks like that is related to activity handling in general, I can reproduce it on master.

PR #147, opened 2016-06-19
Fix opcodes 0x0213 & 0x032B

danhedron:
This provides a proper implementation for 0x032B and cleans up the implementation of 0x0213.

- Move the creation of Pickups into `GameWorld`, where it belongs.
- Improve the Integer value conversion of `SCMOpcodeParameter` and use it instead.
- Use `PickupObject` for non-weapon pickups.
- use `createPickup` in rwgame.

### DIFF
--- a/rwengine/include/script/ScriptTypes.hpp
+++ b/rwengine/include/script/ScriptTypes.hpp
@@ -1,6 +1,8 @@
 #pragma once
 #ifndef _SCRIPTTYPES_HPP_
 #define _SCRIPTTYPES_HPP_
+#include <rw/defines.hpp>
+
 #include <cstdint>
 #include <map>
 #include <string>
@@ -56,23 +58,28 @@ static SCMTypeInfoTable typeData = {
 struct SCMOpcodeParameter {
 	SCMType type;
 	union {
-		uint32_t integer;
+		int32_t integer;
 		float real;
 		char string[8];
 		void* globalPtr;
-		uint32_t* globalInteger;
+		int32_t* globalInteger;
 		float* globalReal;
 	};
 
 	int integerValue() const 
 	{
-		if ( type == TGlobal )
+		switch (type)
 		{
-			return * globalInteger;
-		}
-		else
-		{
+		case TGlobal:
+		case TLocal:
+			return *globalInteger;
+		case TInt8:
+		case TInt16:
+		case TInt32:
 			return integer;
+		default:
+			RW_ERROR("Unhandled type");
+			return 0;
 		}
 	}
 };
--- a/rwengine/include/script/ScriptTypes.hpp
+++ b/rwengine/include/script/ScriptTypes.hpp
@@ -108,6 +108,8 @@ public:
 		return parameters->at(arg);
 	}
 
+	int getModel(unsigned int arg) const;
+
 	template <class T>
 	GameObject* getObject(unsigned int arg) const;
 
--- a/rwengine/src/engine/GameWorld.cpp
+++ b/rwengine/src/engine/GameWorld.cpp
@@ -19,6 +19,7 @@
 #include <objects/InstanceObject.hpp>
 #include <objects/VehicleObject.hpp>
 #include <objects/CutsceneObject.hpp>
+#include <objects/ItemPickup.hpp>
 
 #include <data/CutsceneData.hpp>
 #include <loaders/LoaderCutsceneDAT.hpp>
@@ -456,6 +457,36 @@ CharacterObject* GameWorld::createPlayer(const glm::vec3& pos, const glm::quat& 
 		}
 	}
 	return nullptr;
+}
+
+PickupObject* GameWorld::createPickup(const glm::vec3& pos, int id, int type)
+{
+	auto modelInfo = data->findObjectType<ObjectData>(id);
+
+	data->loadDFF(modelInfo->modelName + ".dff");
+	data->loadTXD(modelInfo->textureName + ".txd");
+
+	PickupObject* pickup = nullptr;
+	auto pickuptype = (PickupObject::PickupType)type;
+
+	// Attempt to find an InventoryItem associated with this model
+	auto it = std::find_if(
+	    inventoryItems.begin(), inventoryItems.end(),
+	    [=](InventoryItem* itm) { return itm->getModelID() == id; });
+
+	// If nothing, create a generic pickup instead of an item pickup
+	if (it != inventoryItems.end()) {
+		pickup = new ItemPickup(this, pos, pickuptype, *it);
+	}
+	else {
+		RW_UNIMPLEMENTED("Non-weapon pickups");
+		pickup = new PickupObject(this, pos, id, pickuptype);
+	}
+
+	pickupPool.insert(pickup);
+	allObjects.push_back(pickup);
+
+	return pickup;
 }
 
 void GameWorld::ObjectPool::insert(GameObject* object)
--- a/rwengine/src/script/ScriptMachine.cpp
+++ b/rwengine/src/script/ScriptMachine.cpp
@@ -81,12 +81,12 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
 				hasExtraParameters = false;
 				break;
 			case TInt8:
-                parameters.back().integer = _file->read<std::uint8_t>(pc);
-                pc += sizeof(SCMByte);
+				parameters.back().integer = _file->read<std::int8_t>(pc);
+				pc += sizeof(SCMByte);
 				break;
 			case TInt16:
-                parameters.back().integer = _file->read<std::int16_t>(pc);
-                pc += sizeof(SCMByte) * 2;
+				parameters.back().integer = _file->read<std::int16_t>(pc);
+				pc += sizeof(SCMByte) * 2;
 				break;
 			case TGlobal: {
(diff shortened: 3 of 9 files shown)


rwengine/src/script/modules/ObjectModule.cpp
```diff
@@ -29,6 +29,30 @@
...
+/**
+ * @brief resolveModelID Resolves script & global model IDs to model id
+ * @param args
+ * @param argument
+ * @return
+ *
+ * Converts negative IDs into the model ID for the object that matches the model
+ * name at that index in the SCM's second section.
+ */
+int resolveModelID(const ScriptArguments& args, int argument)
```
JayFoxRox: nit: I think I would have overloaded this for script args and only supported int arguments natively. But this works

rwengine/include/script/ScriptTypes.hpp
```diff
@@ -66,13 +68,19 @@ struct SCMOpcodeParameter {
...
 		{
-			return integer;
+		case TGlobal:
+			return *globalInteger;
+		case TInt8:
+			return (uint8_t)integer;
+		case TInt16:
+			return (uint16_t)integer;
+		case TInt32:
+			return (uint32_t)integer;
```
JayFoxRox: [Looks like I changed these to signed ints in my commit](https://github.com/JayFoxRox/openrw/pull/11/commits/5455d2912675419032292442732c6c759813b393)

I can't remember where it mattered specifically, but I probably did it for a reason.
I didn't see any forced signed comparisions in SCM yet either so it's probably signed by default.

danhedron: Yeah, these should be signed.

JayFoxRox:
Weapon behind ammunation in "Pump-Action-Pimp" doesn't spawn for me

I believe this is also supposed to fix #132 which affects opcode 03DD.
So if this PR is working properly we should also enable 03DD.

rwengine/include/script/ScriptTypes.hpp
```diff
@@ -66,13 +68,20 @@ struct SCMOpcodeParameter {
...
-		}
-		else
+		switch (type)
 		{
-			return integer;
+		case TGlobal:
+		case TLocal:
+			return *globalInteger;
+		case TInt8:
+			return (int8_t)integer;
```
JayFoxRox: Is this correct? I'm not sure about the internals but I'd assume that `integer` is garbage because it's popping too much data from the stack?
Casting would depend on the endianess then?

(I don't really mind, but we should add a comment about this maybe?)

rwengine/include/script/ScriptTypes.hpp
```diff
@@ -66,13 +68,20 @@ struct SCMOpcodeParameter {
...
-		{
-			return * globalInteger;
-		}
-		else
+		switch (type)
 		{
-			return integer;
+		case TGlobal:
+		case TLocal:
+			return *globalInteger;
```
JayFoxRox: This is still `uint32_t*` above. I think this is also a `int32_t*`?

JayFoxRox:
I only reviewed the final product too, a lot of code in this one is replaced later. You might want to squash a little? (it's up to you though)

Check my diff comments and also re-enable 03DD (should be part of this PR as the blocking part is resolved).
Other than that it LGTM but I'd like to do a final review pass + test once 03DD is included.

rwengine/include/script/ScriptTypes.hpp
danhedron: Yes. This still isn't quite right. I'll fix it further up in the VM so it doesn't need casting here.

rwgame/MenuSystem.hpp
```diff
@@ -147,7 +147,13 @@ class Menu
...
 	void move(int movement)
 	{
-		activeEntry = std::min<const int>(entries.size()-1, std::max<const int>(0, activeEntry + movement));
+		activeEntry += movement;
+		if (activeEntry >= int(entries.size())) {
+			activeEntry = 0;
+		}
+		else if (activeEntry < 0) {
+			activeEntry = entries.size() - 1;
+		}
```
JayFoxRox: Why is this in this PR?!

rwengine/src/script/modules/ObjectModule.cpp
```diff
@@ -1129,6 +1072,20 @@ bool game_character_in_range(const ScriptArguments& args)
...
+	int type = args[1].integer;
+	int ammo = args[2].integer;
+	std::cout << "weapon: " << glm::to_string(pos) << " " << id << std::endl;
+	RW_UNUSED(ammo);
+	RW_UNIMPLEMENTED("game_create_weapon_pickup(): ammo count");
+
+	PickupObject* pickup = args.getWorld()->createPickup(pos, id, type);
+	*args[6].globalInteger = pickup->getGameObjectID();
+}
+
```
JayFoxRox: Maybe this should be a seperate commit? (one for improving the old opcode, one for adding the new one)

JayFoxRox:
Looks very good already. I've mentioned one thing on the diff.
More importantly: Still missing 03DD.

I'll run this once 03DD is included for a final review

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -81,7 +81,7 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
 				hasExtraParameters = false;
 				break;
 			case TInt8:
-                parameters.back().integer = _file->read<std::uint8_t>(pc);
+                parameters.back().integer = _file->read<std::int8_t>(pc);
```
JayFoxRox: while you are at it: fix indentation?

rwengine/src/engine/GameWorld.cpp
```diff
@@ -458,6 +459,36 @@ CharacterObject* GameWorld::createPlayer(const glm::vec3& pos, const glm::quat&
...
+	data->loadDFF(modelInfo->modelName+".dff");
+	data->loadTXD(modelInfo->textureName+".txd");
+
+	PickupObject* pickup = nullptr;
+	auto pickuptype = (PickupObject::PickupType)type;
+
+	// Attempt to find an InventoryItem associated with this model
+	auto it = std::find_if(inventoryItems.begin(),
+						   inventoryItems.end(),
+						   [=](InventoryItem* itm) { return itm->getModelID() == id; });
```
JayFoxRox: indentation. tabs.. *rolls eyes*

rwengine/src/script/modules/ObjectModule.cpp
JayFoxRox: Still my opinion

rwgame/ingamestate.cpp
```diff
@@ -41,16 +41,11 @@ void IngameState::startTest()
...
-		ItemPickup* pickup =
-					new ItemPickup(
-						getWorld(),
-						itemspawn,
-						PickupObject::OnStreet,
-						getWorld()->getInventoryItem(i));
-		getWorld()->pickupPool.insert(pickup);
-		getWorld()->allObjects.push_back(pickup);
+	glm::vec3 itemspawn(276.5f, -609.f, 36.5f);
+	for (int i = 1; i < maxInventorySlots; ++i) {
```
JayFoxRox: nit: don't we use `i++` normally in for loops?!

rwengine/src/script/modules/GameModule.cpp
```diff
@@ -1385,7 +1385,7 @@ GameModule::GameModule()
 	bindUnimplemented( 0x03DA, game_set_garage_follow_player, 1, "Set Garage Camera Follows Player" );
 	
 	bindFunction(0x03DC, game_add_object_blip<PickupObject>, 2, "Add blip for pickup");
-	bindUnimplemented(0x03DD, game_add_object_sprite_blip<PickupObject>, 3, "Add Sprite Blip for Pickup"); /// @todo crashes LUIGI4 because of missing 032B
+	bindFunction(0x03DD, game_add_object_sprite_blip<PickupObject>, 3, "Add Sprite Blip for Pickup");
```
JayFoxRox: nit: for the future: I'd prefer if the commit name somehow explained that it's script related

danhedron: How do you mean?

rwengine/src/script/ScriptMachine.cpp
danhedron: Good catch. I've been trying to sneak clang-format changes here and there, this is a good candidate.

rwengine/src/script/modules/GameModule.cpp
JayFoxRox: At the time of writing the commit is called: "0x3DD: Re-enable as 0x32B is now implemented"
This could mean various things. It's really ambiguous. It could be mistaken for flag values or version numbers of some file format.
Something like "Script: enable 0x03DD because it depended on recently fixed 0x32B" is more descriptive. Other buzz-words such as "opcode" "SCM" etc would have worked too (note that I also highlighted the dependency).

danhedron: I get you. I'll finish tidying this up tomorrow.

JayFoxRox:
Tested and reviewed, couple of nits but we should avoid bike-shedding simple fixes like this.

Thanks and LGTM

PR #148, opened 2016-06-19
Minor Improvements: Missing space, debug menu addition & cleanup

danhedron:
Also:

- Fix camera for tall vehicles being too high
- Debug menu additions
  - Add Weapon debug menu for weapons
  - Add kill all peds
- Add wrap-around to the menu system

### DIFF
--- a/rwgame/debugstate.cpp
+++ b/rwgame/debugstate.cpp
@@ -5,6 +5,8 @@
 #include <objects/InstanceObject.hpp>
 #include <objects/VehicleObject.hpp>
 #include <engine/GameState.hpp>
+#include <items/InventoryItem.hpp>
+#include <data/WeaponData.hpp>
 #include <sstream>
 #include <glm/gtc/quaternion.hpp>
 #include <glm/gtx/string_cast.hpp>
@@ -94,6 +96,9 @@ Menu* DebugState::createDebugMenu()
 	}, kDebugEntryHeight));
 	m->addEntry(Menu::lambda("-AI", [=] {
 		this->enterMenu(createAIMenu());
+	}, kDebugEntryHeight));
+	m->addEntry(Menu::lambda("-Weapons", [=] {
+		this->enterMenu(createWeaponMenu());
 	}, kDebugEntryHeight));
 
 	m->addEntry(Menu::lambda("Set Super Jump", [=] {
@@ -287,6 +292,26 @@ Menu* DebugState::createAIMenu()
 	return m;
 }
 
+Menu*DebugState::createWeaponMenu()
+{
+	Menu* m = new Menu(2);
+	m->offset = kDebugMenuOffset;
+
+	m->addEntry(Menu::lambda("Back", [=] {
+		this->enterMenu(createDebugMenu());
+	}, kDebugEntryHeight));
+
+	for (int i = 1; i < maxInventorySlots; ++i) {
+		auto item = getWorld()->getInventoryItem(i);
+		auto& name = getWorld()->data->weaponData[i]->name;
+		m->addEntry(Menu::lambda(name, [=] {
+			giveItem(item);
+		}, kDebugEntryHeight));
+	}
+
+	return m;
+}
+
 DebugState::DebugState(RWGame* game, const glm::vec3& vp, const glm::quat& vd)
 	: State(game)
 	, _freeLook( false )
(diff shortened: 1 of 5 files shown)


rwgame/debugstate.cpp
```diff
@@ -435,12 +432,29 @@ void DebugState::spawnVehicle(unsigned int id)
 	glm::vec3 fwd = ch->rotation * glm::vec3(0.f, 1.f, 0.f);
 
 	glm::vec3 hit, normal;
-	if(game->hitWorldRay(ch->position + (fwd * 5.f), {0.f, 0.f, -2.f}, hit, normal)) {
+	if(game->hitWorldRay(ch->position + (fwd * 10.f), {0.f, 0.f, -2.f}, hit, normal)) {
```
JayFoxRox: Can you use the code from SPAWN_VEHICLE?
([My spawner commit also had the improved spawnVehicle - you can just steal that](https://github.com/JayFoxRox/openrw/pull/11/commits/8f4cc97f149f4ca6d57ee16057f487ee78a7c244#diff-bd6ee77bba6402560d5fb3194f3760adR345))

The rotation of the vehicle will be different so that it's easier / faster to enter the car or place it in narrow spots.

danhedron: Done. I cherry picked the relevant part of your commit

rwgame/debugstate.cpp
```diff
@@ -261,12 +261,9 @@ Menu* DebugState::createAIMenu()
 
 #define SPAWN_FOLLOWER(name, id) \
 	m->addEntry(Menu::lambda("Add " name " Follower", [=] { \
-		auto spawnPos = player->getPosition() + player->getRotation() * glm::vec3(0.f, 1.f, 0.f); \
-		auto follower = game->getWorld()->createPedestrian(id, spawnPos); \
-		jumpCharacter(game, follower, spawnPos); \
-		follower->controller->setGoal(CharacterController::FollowLeader); \
-		follower->controller->setTargetCharacter(player); \
+		spawnFollower(id); \
```
JayFoxRox: With this only being a single line you should call this directly. Alternatively I think you should consider using a lambda function.
C++ and macros are meh..

rwlib/source/rw/defines.hpp
```diff
@@ -19,6 +19,6 @@
 	(void) var
 
 #define RW_UNIMPLEMENTED(msg) \
-	RW_MESSAGE("Unimplemented" << msg)
+	RW_MESSAGE("Unimplemented: " << msg)
```
JayFoxRox: nit: more descriptive commit name next time so we know if this is to fight a regression OR a style fix OR just a visual fix

rwgame/debugstate.cpp
```diff
@@ -451,18 +451,17 @@ void DebugState::spawnVehicle(unsigned int id)
...
-		jumpCharacter(game, follower, spawnPos);
-		follower->controller->setGoal(CharacterController::FollowLeader);
-		follower->controller->setTargetCharacter(ch);
-	}
+	auto playerRot = ch->getRotation();
+	auto spawnPos = ch->getPosition();
+	spawnPos += playerRot * glm::vec3(0.f, 3.f, 0.f);
+	auto spawnRot = glm::quat(
+	    glm::vec3(0.f, 0.f, glm::roll(playerRot) + glm::half_pi<float>()));
+	auto follower = game->getWorld()->createPedestrian(id, spawnPos, spawnRot);
```
JayFoxRox: I think the old logic was better here.

It makes sense to rotate a car because large vehicles like the "Coach" are very bad to place properly if it's always facing north. Also it's useful to spawn vehicles in the air/water if the vehicle is used for testing purposes (or a boat later!)

However, a character will fall to his death if not spawned on ground level.
I don't care where a character is facing because AI will take over + it's small enough that rotation doesn't have an impact (especially not with the current capsule collider shape).

rwgame/debugstate.cpp
```diff
@@ -285,6 +290,34 @@ Menu* DebugState::createAIMenu()
 	return m;
 }
 
+Menu*DebugState::createWeaponMenu()
+{
+	CharacterObject* player = nullptr;
+	if (game->getPlayer()) {
```
JayFoxRox: if this condition is false player will be `nullptr` in the lambda function.
I think you mean to do this check in the lambda function below.

rwgame/debugstate.cpp
```diff
@@ -285,6 +290,34 @@ Menu* DebugState::createAIMenu()
...
+	m->addEntry(Menu::lambda("Back", [=] {
+		this->enterMenu(createDebugMenu());
+	}, kDebugEntryHeight));
+
+	for(int i = 1; i < maxInventorySlots; ++i) {
+		auto item = getWorld()->getInventoryItem(i);
+		auto name = getWorld()->data->weaponData[i]->name;
+		m->addEntry(Menu::lambda(name, [=] {
+			player->addToInventory(item);
+		auto& wep = player->getCurrentState().weapons[item->getInventorySlot()];
```
JayFoxRox: indentation bad

rwgame/debugstate.cpp
danhedron: Good point. I overlooked that behaviour in jumpCharacter

rwgame/debugstate.cpp
tsjost: getPlayer() might be null, but if not, getCharacter() has to be called on it, and the result of which is used (and checked) in the lambda. It could all be done inside the lambda, but it might be unnecessary to do this vtable parkour in a loop. However, is the CharacterObject guaranteed to continue existing during the DebugState's lifetime? Otherwise the lambda might be poking at invalid memory?

danhedron: I've cleaned this up to use another function to give the player the weapon.

PR #149, opened 2016-06-21
Copy License to install

danhedron:
This only covers Linux and FreeBSD.

Anyone who knows what the right thing to do for windows and macOS is invited to tell me what to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,3 +107,7 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux|FreeBSD")
 			DESTINATION share/openrw
 			)
 endif()
+# And copy to build directory for CI
+file(COPY COPYING
+	 DESTINATION .
+	 )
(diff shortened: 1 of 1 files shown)


JayFoxRox:
What I said earlier on IRC mostly referred to including it with the build at all.
This is important if we have artifacts for download from appveyor or travis.

However, this is still fine I guess. But it's really up to how each distro manages it.
My understanding is that some distros have extra room for the license file in their package description.

I believe on Windows it's common practice to either zip up the project and have a license.txt OR show it in the installer - there is no designated folder to my knowledge.
There are some common GUI styling guides by Apple and Microsoft though and those might mention where to display such information.
The Windows one mentions the license terms for the Setup for example: https://msdn.microsoft.com/en-us/library/windows/desktop/dn742461(v=vs.85).aspx

danhedron:
I've added the file to the build directory too.

re the install: A lot of packages provide a LICENSE/COPYING file along with their license metadata, it seems reasonable to do.

We probably won't have an installer for some time, so copying to the build directory is fine by me for now.

PR #150, opened 2016-06-21
Add screenshots to readme

danhedron:


README.md
```diff
@@ -14,6 +14,7 @@ Without this data it will not be possible to run OpenRW.
 
  * [Getting Started](https://github.com/rwengine/openrw/wiki/Getting-Started)
  * [FAQ](https://github.com/rwengine/openrw/wiki/FAQ)
+ * [Screenshots & Videos](https://github.com/rwengine/openrw/wiki/Media)
```
JayFoxRox: How is this "Getting Started" ? Shouldn't this get its own subsection?
Also the FAQ is also linked to in the section just before this one.

imo this section should be giving hints what's contained in the getting started page only, then link to it. (possibly contain the gist of it in case someone is still not always-online in 2016)

Also there should just be a ## Screenshots section here which show the image right on github (only 1 or 2 screens should be fine; `OpenRW running GTA3`). You can put the link there

#bikeshedding-with-style

danhedron: I'll tidy up the semantics of the section, but I'm hesitant to include images directly in the read-me at this point.

danhedron:
How are these @JayFoxRox, nothing special but they get the point across that the game is in a semi-working state.

PR #151, opened 2016-06-22
Use RGBA for screenshots to use 4-byte row alignment

danhedron:
On my screen resolution `1366x768` the default 4-byte alignment of `GL_PACK_ALIGNMENT` causes glReadPixels to not work as expected for RGB output. Save RGBA instead to ensure that rows are always 4-byte aligned,


danhedron:
There are better ways to get screenshots. I'd rather not try to keep this not very useful path alive.

PR #152, opened 2016-06-22
Replace SFML by SDL2 for Travis and scripttool

JayFoxRox:
I was originally only going to fix the `.travis.yml` but then noticed that the scripttool also still linked against SFML.

Let's see if Travis works now :)

//Edit: FindSDL2.cmake has been stolen from citra-emu/citra which works on OSX, Linux and Windows in MSVC, Clang and GCC.


danhedron:
This looks good to me

Hockenberry:
I could compile and run this branch under macosx 10.11.5.

PR #153, opened 2016-06-23
Mission vehicle base

JayFoxRox:
PR extends `master` towards being able to add vehicle missions.
This has been extracted from a larger PR which will follow soon.
There is nothing too interesting here.

However, I'd like to draw your attention to the input hack which maps the `player->isRunning()` return value to the script button-state-getter-opcodes sub-mission key.
This works because OpenRW always updates `isRunning` even if the player is in a car. It also means you can start the vehicle missions by pressing shift (or whatever the sprint key may be) while in a mission vehicle.
I'm aware that this is a cruel hack, but OpenRW doesn't maintain a input state yet. IMHO this is fine as a temporary fix. I'll create an issue about this once this PR is merged too, to remind us of this ugly piece.

None of this was tested much and might contain bugs. However, we can improve that later (after merge).
Consider the implementations to be stubs and unverified.

### DIFF
--- a/rwengine/include/script/ScriptTypes.hpp
+++ b/rwengine/include/script/ScriptTypes.hpp
@@ -82,6 +82,22 @@ struct SCMOpcodeParameter {
 			return 0;
 		}
 	}
+
+	float realValue() const
+	{
+		switch (type)
+		{
+		case TGlobal:
+		case TLocal:
+			return *globalReal;
+		case TFloat16:
+			return real;
+		default:
+			RW_ERROR("Unhandled type");
+			return 0;
+		}
+	}
+
 };
 
 typedef std::vector<SCMOpcodeParameter> SCMParams;
--- a/rwengine/src/script/modules/GameModule.cpp
+++ b/rwengine/src/script/modules/GameModule.cpp
@@ -66,6 +66,20 @@ void game_print_now(const ScriptArguments& args)
 					));
 }
 
+void game_print_soon(const ScriptArguments& args)
+{
+	const auto& world = args.getWorld();
+
+	std::string id(args[0].string);
+	int time = args[1].integerValue();
+	unsigned short style = args[2].integerValue();
+
+	auto str = world->data->texts.text(id);
+
+	auto textEntry = ScreenTextEntry::makeBig(id, str, style, time);
+	world->state->text.addText<ScreenTextType::BigLowPriority>(textEntry);
+}
+
 void game_clear_prints(const ScriptArguments& args)
 {
 	args.getWorld()->state->text.clear<ScreenTextType::Big>();
@@ -1162,6 +1176,7 @@ GameModule::GameModule()
 	
 	bindFunction(0x00BA, game_print_big, 3, "Print big" );
 	bindFunction(0x00BC, game_print_now, 3, "Print Message Now" );
+	bindFunction(0x00BD, game_print_soon, 3, "Print Message Soon" );
 	
 	bindFunction(0x00BE, game_clear_prints, 0, "Clear Message Prints" );
 	bindFunction(0x00BF, game_get_time, 2, "Get Time of Day" );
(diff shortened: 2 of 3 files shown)


rwengine/src/script/modules/VMModule.cpp
```diff
@@ -52,6 +52,15 @@ void vm_dec_global_float(const ScriptArguments& args)
 	*args[0].globalReal -= args[1].real;
 }
 
+void vm_mul_global_int(const ScriptArguments& args)
+{
+	*args[0].globalInteger /= args[1].integer;
```
danhedron: Should this be a multiplication?

JayFoxRox:
Good catch! That would have been a very weird way to do multiplication..
Fixed.

rwengine/src/script/modules/GameModule.cpp
```diff
@@ -1162,6 +1174,7 @@ GameModule::GameModule()
 	
 	bindFunction(0x00BA, game_print_big, 3, "Print big" );
 	bindFunction(0x00BC, game_print_now, 3, "Print Message Now" );
+	bindFunction(0x00BD, game_print_soon, 3, "Print Message Soon" );
```
tsjost: Space shouldn't appear before )

rwengine/src/script/modules/GameModule.cpp
```diff
@@ -66,6 +66,18 @@ void game_print_now(const ScriptArguments& args)
...
+	std::string id(args[0].string);
+	std::string str = args.getWorld()->data->texts.text(id);
+	int time = args[1].integer;
+	unsigned short style = args[2].integerValue(); // same as flags?!
+	args.getWorld()->state->text.addText<ScreenTextType::BigLowPriority>(
+				ScreenTextEntry::makeBig(
+					id, str, style, time
+					));
+}
+
```
tsjost: Weird indentation.

We don't have a style standard for this type of thing, I think. Maybe something like:
```
args.getWorld()->state->text.addText<ScreenTextType::BigLowPriority>(
	ScreenTextEntry::makeBig(
		id, str, style, time
	)
);
```
or because the parameters are short
```
args.getWorld()->state->text.addText<ScreenTextType::BigLowPriority>(
	ScreenTextEntry::makeBig(id, str, style, time)
);
```

rwengine/src/script/modules/GameModule.cpp
```diff
@@ -903,6 +925,22 @@ void game_start_chase_scene(const ScriptArguments& args)
...
+				args.getWorld()->data->texts.text(id),
+				formatValue(args[1]),
+				formatValue(args[2]));
+	int time = args[3].integer;
+	unsigned short style = args[4].integer;
+	args.getWorld()->state->text.addText<ScreenTextType::Big>(
+				ScreenTextEntry::makeBig(
+					id, str, style, time
+					));
+}
```
tsjost: Weird indentation like earlier.

rwengine/src/script/modules/GameModule.cpp
JayFoxRox: Kept the style of surrounding code, didn't want to touch more than necessary. I think this is a good candidate for a clang-format commit.
I also might fix it with my upcoming text PR which touches almost all of these opcodes.

rwengine/include/script/ScriptTypes.hpp
```diff
@@ -82,6 +82,16 @@ struct SCMOpcodeParameter {
 			return 0;
 		}
 	}
+
+	float realValue() const
+	{
+		if (type == TGlobal) {
+			return *globalReal;
+		} else {
```
danhedron: Should this include TLocal for completeness?

JayFoxRox: Yes, this is a bug!

danhedron:
This is looking good to me, how many more opcodes are fundamental for vehicle missions?

JayFoxRox:
There'll be seperate PRs for Taxi and Ambulance after this. (Taxi will be 14 commits probably, I'm assuming about 10 commits for the other one).
I also plan a larger PR to fix text rendering as vehicle missions use various text message styles at once (it's low priority though).

We shouldn't work on vigilante (vehicle damage and AI missing) and firefighter yet (fire + water cannon missing).

PR #155, opened 2016-06-23
Implemented proper RW-style shading for world objects, in particular …

aap:
…vertex colours

Statically lit geometry is now shaded as I described here: http://www.gtamodding.com/wiki/Rendering_with_RenderWare#Default_rendering

Screenshot:
Screenshot of proper vertex colours

(Markup fixed by JayFoxRox)

### DIFF
--- a/rwengine/src/render/GameShaders.cpp
+++ b/rwengine/src/render/GameShaders.cpp
@@ -236,12 +236,14 @@ float alphaThreshold = (1.0/255.0);
 void main()
 {
 	// Only the visibility parameter invokes the screen door.
-	vec4 diffuse  = texture(tex, TexCoords);
+	vec4 diffuse = Colour;
+	diffuse.rgb += ambient.rgb*ambientfac;
+	diffuse *= colour;
+	diffuse *= texture(tex, TexCoords);
 	if(visibility <= filterMatrix[int(gl_FragCoord.x)%4][int(gl_FragCoord.y)%4]) discard;
 	if(diffuse.a <= alphaThreshold) discard;
-	vec4 colour_dyn = diffuse * colour;
 	float fog = 1.0 - clamp( (fogEnd-WorldSpace.w)/(fogEnd-fogStart), 0.0, 1.0 );
-	fragOut = vec4(mix(colour_dyn.rgb * dynamic.rgb + ambient.rgb * ambient.a, fogColor.rgb, fog), colour_dyn.a);
+	fragOut = vec4(mix(diffuse.rgb, fogColor.rgb, fog), diffuse.a);
 })";
 
 const char* Particle::FragmentShader = R"(


danhedron:
Thanks. This looks good to me.

rwengine/src/render/GameShaders.cpp
```diff
@@ -236,12 +236,14 @@ float alphaThreshold = (1.0/255.0);
 void main()
 {
 	// Only the visibility parameter invokes the screen door.
-	vec4 diffuse  = texture(tex, TexCoords);
+	vec4 diffuse = Colour;
+	diffuse.rgb += ambient.rgb*ambientfac;
```
JayFoxRox: nit: spaces between operator and var

rwengine/src/render/GameShaders.cpp
```diff
@@ -236,12 +236,14 @@ float alphaThreshold = (1.0/255.0);
...
+	diffuse *= colour;
+	diffuse *= texture(tex, TexCoords);
 	if(visibility <= filterMatrix[int(gl_FragCoord.x)%4][int(gl_FragCoord.y)%4]) discard;
 	if(diffuse.a <= alphaThreshold) discard;
-	vec4 colour_dyn = diffuse * colour;
 	float fog = 1.0 - clamp( (fogEnd-WorldSpace.w)/(fogEnd-fogStart), 0.0, 1.0 );
-	fragOut = vec4(mix(colour_dyn.rgb * dynamic.rgb + ambient.rgb * ambient.a, fogColor.rgb, fog), colour_dyn.a);
+	fragOut = vec4(mix(diffuse.rgb, fogColor.rgb, fog), diffuse.a);
```
JayFoxRox: I felt the colors were too strong (at least compared to what I remember).

So I looked into it:
```
out = (ambientLight*ambientProperty + diffuseLight*diffuseProperty + prelighting)*materialColor
```
(given this from your reference - which is also by you, but it's fine)

Assumption:
```
ambientLight = ambient.rgb
ambientProperty = ambientfac
diffuseLight = Colour
diffuseProperty = colour
prelighting = 0.0 <- not supported yet?!
materialColor = texture(tex, TexCoords)
```

Wouldn't the proposed shader do:
```
out_intended = (ambientLight*ambientProperty + diffuseLight*diffuseProperty + prelighting)*materialColor
out_shader   = (ambientLight*ambientProperty * diffuseLight*diffuseProperty + prelighting)*materialColor
                                             ^--- Bug?
```

aap:
Wrong assumption, the variable names in the shader are a bit unfortunate:

```
ambientLight = ambient.rgb
ambientProperty = ambientfac
diffuseLight = dynamic.rgb but unused as world geometry has no normals
diffuseProperty = diffusefac but unused as well
prelighting = Colour
materialColor = colour
```

This is the *lighting* equation, which happens traditionally at the vertex stage. In the pixel stage the resulting color is just modulated with the texture.

JayFoxRox:
Okay, LGTM then.
(I'm probably mistaken about colors being too vibrant here or it's due to missing diffuse lighting)

---

It's not scope of this PR to improve the shader code.
However, we should have an issue about this after merging: use the variables name we have now used in our assumptions because they are somewhat more descriptive + move stuff to the VS for performance.

---

( @aap genral github etiquette: respond using another "line note" in diff comments - otherwise your response won't be shown in the diff view which can be used to see the context / code of the discussion)

PR #156, opened 2016-06-23
Remove screenshot functionality, as it's broken.

danhedron:


christoph-heiss:
Is it because of the alignment issue?
This could be solved by `glPixelStorei(GL_UNPACK_ALIGNMENT, 1);`

Otherwise, what is broken?

danhedron:
Yes; that is the issue. However there are plenty of ways to take a screenshot, using kscreenshot for example. I'd rather remove the path than try and beat these tools.

JayFoxRox:
I guess there are a couple of reasons: Ugly pointer interface, lots of hardly used code ( #151 : there are better ways to take screenshots on all platforms) and the icing is the current alignment issue.

\- Unless GTA does it we don't have to replicate it and it shouldn't be our task to implement screenshotting which is to be expected on any recent OS.

Even if we need screenshots for the Camera in SA we can still recover this code in a couple of months when SA even partially works.
Eventually we'll have very different requirements and environment then though.

I appreciate this change.

---

Still compiles, game still works, F12 does nothing. Code also looks fine and grep doesn't find any mentions of "screenshot" either.

LGTM.

(We might have to change documentation after merging this)

christoph-heiss:
Well, that's completely reasonable.

PR #157, opened 2016-06-24
Include travis build status into readme.md

Hockenberry:


PureTryOut:
Most projects have it next to the project title, so line 1. Any reason for not doing the same?

JayFoxRox:
I agree with @PureTryOut.
The current position somehow isn't very logical

JayFoxRox:
LGTM.

(though, I'm guessing this will always link to the latest build, not the latest master build. This is common practice but very confusing imo because it shows "Passing" but once you click it you might arrive on a "Failed" build. However, that's fine because this is also how it works everywhere else)

PR #158, opened 2016-06-25
MotionStates & physics cleanup

### LINKED ISSUE #72: Physics objects are invisible
Detached car doors, street lights, and probably others are invisible.
screenshot_2016-05-23_22-34-05

danhedron:
This is a subset of changes included in #119 that aren't related to vehicles. Vehicle physics requires a bit more work so these changes can be fast-tracked since they're complete.

- Remove defunct code
- Use bullet motion states to update object transformations
- make getPosition/getRotation non-virtual 
- Instances (lamposts, boxes etc) will now move along with their rigid body

Fixes #72 (doors will be fixed in the next PR)

### DIFF
--- a/rwengine/include/objects/CharacterObject.hpp
+++ b/rwengine/include/objects/CharacterObject.hpp
@@ -158,14 +158,12 @@ public:
 
 	virtual void setPosition(const glm::vec3& pos);
 
-	virtual glm::vec3 getPosition() const;
-
-	virtual glm::quat getRotation() const;
-
 	bool isAlive() const;
 	bool takeDamage(const DamageInfo& damage) override;
 
 	bool enterVehicle(VehicleObject* vehicle, size_t seat);
+
+	bool isEnteringOrExitingVehicle() const;
 
 	/**
 	 * @brief isStopped
--- a/rwengine/include/objects/GameObject.hpp
+++ b/rwengine/include/objects/GameObject.hpp
@@ -98,10 +98,10 @@ public:
 	
 	virtual void setPosition(const glm::vec3& pos);
 
-	virtual glm::vec3 getPosition() const { return position; }
+	const glm::vec3& getPosition() const { return position; }
 	const glm::vec3& getLastPosition() const { return _lastPosition; }
 
-	virtual glm::quat getRotation() const;
+	const glm::quat& getRotation() const { return rotation; }
 	virtual void setRotation(const glm::quat &orientation);
 
 	float getHeading() const;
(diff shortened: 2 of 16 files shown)


rwengine/src/objects/InstanceObject.cpp
```diff
@@ -43,20 +43,17 @@ InstanceObject::~InstanceObject()
...
-				body->body->setMassProps(dynamics->mass, inert);
-				engine->dynamicsWorld->addRigidBody(body->body);
+				body->changeMass(dynamics->mass);
 			}
 		}
 		
 		_updateLastTransform();
 
-		auto _bws = body->body->getWorldTransform().getOrigin();
+#warning replace with position from motionstate
```
JayFoxRox: That's quite an odd compile-time warning.
Isn't this supposed to be a `/// @todo` + issue instead?

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -166,7 +202,10 @@ void VehicleObject::tickPhysics(float dt)
 		// todo: a real engine function
 		float velFac = info->handling.maxVelocity;
 		float engineForce = info->handling.acceleration * throttle * velFac;
-		if( fabs(engineForce) >= 0.001f ) physBody->activate(true);
+		if( fabs(engineForce) >= 0.001f )
+		{
```
JayFoxRox: nit: style wise another instance in this file uses a single line for `if ( condition ) {`. this one has a newline before the brace

rwengine/include/objects/GameObject.hpp
```diff
@@ -185,9 +185,18 @@ class GameObject
...
+	{
+		_lastPosition = position;
+		_lastRotation = rotation;
+		position = pos;
+		rotation = rot;
+	}
+
 private:
 	ObjectLifetime lifetime;
+
```
JayFoxRox: nit: useless newline

rwengine/include/objects/CharacterObject.hpp
```diff
@@ -158,15 +158,13 @@ class CharacterObject : public GameObject
 
 	virtual void setPosition(const glm::vec3& pos);
 
-	virtual glm::vec3 getPosition() const;
```
JayFoxRox: nit: commit name has punctation as last symbol

JayFoxRox:
I haven't really checked the code for logic errors yet (and not sure if I want to tbh :P )

However, during testing I noticed that driving a car (bobcat) into the water will cause it to rock back and forth faster and faster with this PR (eventually slowing down to speed up again). I've tried to reproduce it in master and while the car seems to vibrate a little the water seems to damp the motion instead of adding to it.

Aside from that everything seems to work fine.

danhedron:
Since water behaviour needs another look-at in future, that can be filed as an issue.

danhedron:
Now includes a fix for the crash that has become more common since #159

PR #159, opened 2016-06-25
Parked traffic and out-of-view spawning

### LINKED ISSUE #16: Utilize Bullet MotionStates correctly
Currently not used in a sensible way, leaking memory.

danhedron:
This adds:
- spawning of vehicles from vehicle generators
-  limits spawning to outside of the view near the player (with exceptions)

Fixes #16 

Issues:
- [x] The time limit on vehicle generators is not respected
- [x] Many vehicle generators are at -100.f and never get used
- [ ] Vehicle can spawn in the ground
- [x] Save game data isn't used

### DIFF
--- a/rwengine/include/ai/TrafficDirector.hpp
+++ b/rwengine/include/ai/TrafficDirector.hpp
@@ -9,6 +9,7 @@
 
 class GameObject;
 class GameWorld;
+class ViewCamera;
 
 class TrafficDirector
 {
@@ -16,17 +17,17 @@ public:
 	
 	TrafficDirector(AIGraph* graph, GameWorld* world);
 	
-	std::vector< AIGraphNode* > findAvailableNodes(AIGraphNode::NodeType type, const glm::vec3& near, float radius);
+	std::vector< AIGraphNode* > findAvailableNodes(AIGraphNode::NodeType type, const ViewCamera& camera, float radius);
 	
 	void setDensity(AIGraphNode::NodeType type, float density);
 
 	/**
 	 * Creates new traffic at available locations.
-	 * @param center The area to spawn around
+	 * @param camera The camera to spawn around
 	 * @param radius the maximum distance to spawn in
 	 * @param max The maximum number of traffic to create.
 	 */
-	std::vector<GameObject*> populateNearby( const glm::vec3& center, float radius, int maxSpawn = -1 );
+	std::vector<GameObject*> populateNearby( const ViewCamera& camera, float radius, int maxSpawn = -1 );
 
 	/**
 	 * Sets the maximum number of pedestrians and cars in the traffic system
--- a/rwengine/include/data/VehicleGenerator.hpp
+++ b/rwengine/include/data/VehicleGenerator.hpp
@@ -1,0 +1,35 @@
+#ifndef RWENGINE_VEHICLEGENERATOR_HPP
+#define RWENGINE_VEHICLEGENERATOR_HPP
+
+/**
+ * Stores information about where the game can generate vehicles.
+ */
+struct VehicleGenerator
+{
+	glm::vec3 position;
+	float heading;
+	/// ID of the vehicle to spawn, or -1 for random.
+	int vehicleID;
+	/// @todo not yet used
+	int colourFG;
+	/// @todo not yet used
+	int colourBG;
+	bool alwaysSpawn;
+	/// @todo not yet used
+	short alarmThreshold;
+	/// @todo not yet used
+	short lockedThreshold;
+
+	int minDelay;
+	/// @todo not yet used
+	int maxDelay;
+	int lastSpawnTime;
+
+	/**
+	 * Number of vehicles left to spawn 0-100, 101 = never decrement.
+	 * Intentionally disabled to match behaviour
+	*/
+	int remainingSpawns;
+};
+
+#endif
--- a/rwengine/include/data/VehicleGenerator.hpp
+++ b/rwengine/include/data/VehicleGenerator.hpp
@@ -1,5 +1,6 @@
 #ifndef RWENGINE_VEHICLEGENERATOR_HPP
 #define RWENGINE_VEHICLEGENERATOR_HPP
+#include <glm/glm.hpp>
 
 /**
  * Stores information about where the game can generate vehicles.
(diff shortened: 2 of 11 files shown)


rwengine/src/engine/GameWorld.cpp
```diff
@@ -946,3 +958,57 @@ bool GameWorld::isPaused() const
...
+
+	bool clear = true;
+
+	if (gen.remainingSpawns <= 0) {
+		clear = false;
+	}
+
+	if (gen.lastSpawnTime + gen.minDelay > state->basic.timeMS) {
+		clear = false;
+	}
```
JayFoxRox: while `clear` is a good name to check if the area is `clear` I feel like this (and the instance just before) is abuse of the variable. Maybe just `return nullptr` here or call the variable differently?

Early-out feels like a good idea in general for this function.

rwengine/src/engine/GameWorld.cpp
```diff
@@ -946,3 +958,57 @@ bool GameWorld::isPaused() const
...
+
+	// Ensure there's no existing vehicles near our spawn point
+	if (clear) {
+		for (auto& v : vehiclePool.objects)
+		{
+			if (glm::distance2(gen.position, v.second->getPosition())
+					< kMinClearRadius * kMinClearRadius) {
+				clear = false;
+				break;
```
JayFoxRox: Maybe just early out here (out of the function)? instead of using a variable?

rwengine/include/data/VehicleGenerator.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+	glm::vec3 position;
+	float heading;
+	/** ID of the vehicle to spawn, or -1 for random. */
+	int vehicleID;
+	int colourFG;
+	int colourBG;
+	bool alwaysSpawn;
+	short alarmThreshold;
+	short lockedThreshold;
```
JayFoxRox: Please add a /// @todo to those variables we currently don't use (and possibly create an issue after merge!).
There will be hard to detect bugs later otherwise!

rwengine/include/data/VehicleGenerator.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+	int minDelay;
+	int maxDelay;
+	/** Incrementing timer (in ms), will only spawn a vehicle when minDelay < spawnTimer and will always spawn if maxDelay < spawnTimer */
+	int lastSpawnTime;
```
JayFoxRox: `maxDelay` is not checked yet! This makes the comment wrong.
We need a `/// @todo` here.

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -72,6 +74,18 @@ std::vector<GameObject*> TrafficDirector::populateNearby(const glm::vec3& center
 
 	std::vector<GameObject*> created;
 
+	// Spawn vehicles at vehicle generators
+	/// @todo avoid spawning in view frustum unless alwaysSpawn == true
+
```
JayFoxRox: nit: useless empty line

rwengine/include/ai/TrafficDirector.hpp
```diff
@@ -9,24 +9,25 @@
...
 
 	/**
 	 * Creates new traffic at available locations.
-	 * @param center The area to spawn around
+	 * @param camera The camera to spawn around
 	 * @param radius the maximum distance to spawn in
 	 * @param max The maximum number of traffic to create.
 	 */
-	std::vector<GameObject*> populateNearby( const glm::vec3& center, float radius, int maxSpawn = -1 );
+	std::vector<GameObject*> populateNearby( const ViewCamera& center, float radius, int maxSpawn = -1 );
```
JayFoxRox: Parameter is still named `center`. Should be `camera`

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -18,36 +19,39 @@ maximumPedestrians(20), maximumCars(10)
...
 	float density = type == AIGraphNode::Vehicle ? carDensity : pedDensity;
 	float minDist = (10.f / density) * (10.f / density);
+	float halfRadius2 = std::pow(radius / 2.f, 2.f);
 
-	// Determine if anything in the open set is blocked
-	for ( auto it = available.begin(); it != available.end(); )
-	{
+	// Check if any of the nearby nodes are blocked by a pedestrian standing on it
+	// or because it's inside the view frustum
+	for (auto it = available.begin(); it != available.end();) {
```
JayFoxRox: indentation

JayFoxRox:
- I've had one crash with this branch (but no debugger attached so I'm not sure if this PR is to blame)

- I noticed that some cars are still missing (such as the firetruck in Portland). Got any explanation for that?

- Also some cars like the Mission Patriot near the docks can be seen jumping out of the ground when it spawns. I think cars might be spawning too low?

- Most importantly: The car generators don't work in savegames yet. I think this should be fixed before a merge.

I'll do another test + review once you have addressed these issues.
(Aside from that this PR is a huge step for OpenRW imo. Good job!)

rwengine/src/engine/GameWorld.cpp
```diff
@@ -998,6 +998,9 @@ VehicleObject* GameWorld::tryToSpawnVehicle(VehicleGenerator& gen)
 		/// @todo use zone information to decide vehicle id
 	}
 
+	/// @todo is this sufficient? should dimensions be used?
+	position.z += 1.f;
```
JayFoxRox: Where does the 1.0f come from?
Maybe the game always sends a ray downwards to find the proper spot? What about spawning on slopes?

(Either add more comments or, probably better, make sure to open an issue about these things and explain where values come from)

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -85,8 +85,10 @@ std::vector<GameObject*> TrafficDirector::populateNearby(const ViewCamera& camer
 	float halfRadius2 = std::pow(radius/ 2.f, 2.f);
 
 	// Spawn vehicles at vehicle generators
+	auto camera2D = glm::vec2(camera.position);
 	for (auto& gen : world->state->vehicleGenerators) {
-		float dist2 = glm::distance2(camera.position, gen.position);
+		auto gen2D = glm::vec2(gen.position);
```
JayFoxRox: Unless you know this for certain, add a note (`/// @todo`) that we are not sure how the check is done yet.

(Maybe ground Z is detected first and then a 3D distance is calculated? It probably doesn't matter much for the stock game, but we might have mods at some point. Also some people might be interested in figuring this stuff out for us)

rwengine/src/engine/SaveGame.cpp
```diff
@@ -1189,6 +1189,24 @@ bool SaveGame::loadGame(GameState& state, const std::string& file)
 		vehicle->setSecondaryColour(car.colorBG);
 	}
 
+	for (int g = 0; g < carGenerators.size(); ++g) {
+		auto& gen = carGenerators[g];
+		state.vehicleGenerators.push_back({
```
JayFoxRox: isn't this what emplace_back is for?

JayFoxRox: Actually.. can't you just add a function to return a proper VG from the SaveGame VG type? (Up to you)

JayFoxRox:
(Comments above are nits only)

---

While the patriot spawns fine now other cars drop now.
Remove the last commit and create an issue about it instead (doesn't have to be fixed now and who knows if we'll ever fix it properly if we have that ugly hack in place already).

Tested and reviewed. LGTM otherwise.

danhedron:
This does make it easier to reproduce a vehicle related crash, so I'll nip that with the physics PR.

PR #164, opened 2016-06-28
Stub Taxi-Mission

JayFoxRox:
This implements (stubs) the 'Taxi Driver' mission.

---

To test this, go to the last island (teleport to airport for example), enter a taxi, drive near passengers and press "Shift" (or whatever your Sprint key is)

---

**Known issues (all of which can be fixed after merge):**

- Peds like to spawn in very bad spots (Unrelated, this is  a problem of the traffic spawner)
- Peds are "flickering" when selecting them (probably a problem with the traffic spawner)
- Always claims you earned "$-1" (no idea why)
- Peds don't wait for your taxi and don't hail the taxi (stubbed only)
- Taxi lights stay off (stubbed only)
- Only works on last island (airport etc.) because we currently don't support the script function to check if an island is loaded

---

I intend to create an issue about the "random character in zone" finder because it does always grab the first character it can find. This has to be verified with game behaviour; also the code is not very clean in that function and it should be fixed later
(however, compatibility should be more important for now, so we can use this ugly version for now)

As always: Reviews + Feedback welcome

### DIFF
--- a/rwengine/include/objects/VehicleObject.hpp
+++ b/rwengine/include/objects/VehicleObject.hpp
@@ -132,6 +132,12 @@ public:
 	void setPrimaryColour(uint8_t color);
 	void setSecondaryColour(uint8_t color);
 
+	/**
+	* @brief isStopped
+	* @return True if the vehicle isn't moving
+	*/
+	bool isStopped() const;
+
 private:
 
 	void registerPart(ModelFrame* mf);
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -467,8 +467,11 @@ bool CharacterObject::enterVehicle(VehicleObject* vehicle, size_t seat)
 
 bool CharacterObject::isStopped() const
 {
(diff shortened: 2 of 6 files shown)


danhedron:
Tried it out. Looks to be working. LGTM

JayFoxRox:
I've now turned `sqrtf` into `std::sqrt()` and `rand()` into `std::rand()`

Also the random char is actually random now (which might be wrong, I have not tested game behavior yet)

---

I'll create an issue about the random char being possibly wrong after the merge.

Also we should figure out when the game removes the traffic goals from a char:
The taxi script seems to pick a random char, then makes them idle first, tells them to ignore everybody, then hail the taxi; then enter the taxi; then leave the taxi; then sets them to wander again.
\- So the question is: does the idle state completly remove the traffic behavior already? or is there something to stop the idle, so it will automaticly result in traffic behavior again?
I believe game_character_make_idle should clear the goal + maybe also activity (?) in OpenRW, but we should discuss this in a seperate issue, which I'll create after merge.

PR #165, opened 2016-06-28
CMake: Add BIN_DIR and DOC_DIR

orbea:
This will add `BIN_DIR` and `DOC_DIR` to the cmake build options so that anyone compiling openrw from source can choose custom locations for these files when installed. Additionally it will move `COPYING` from `share/openrw` to the correct standard documentation directory `share/doc/openrw`.

These options will be documented if you try this command in the openrw source directory.
`mkdir build; cd build; cmake -LAH .. | less`
```
// BIN_DIR
BIN_DIR:PATH=bin

// DOC_DIR
DOC_DIR:PATH=share/doc/openrw
```

This is useful for distros like Slackware which may not use the standard locations, for example with this commit I can build openrw with: `-DBIN_DIR=games  -DDOC_DIR=doc/$PRGNAM-$VERSION`


danhedron:
Thanks. This looks good to me.

orbea:
Almost...I made a small typo which I fixed now...

PR #166, opened 2016-06-29
Seperate Input state from binding

danhedron:
This separates the SDL key bindings from the logic of those events, and also makes input state part of the engine code. This will make it easier to implement certain opcodes.

- [x] Record logical input state in the engine code
- [x] Use logical input state for controlling the player
- [x] Add player sprint, as we have proper bindings now.

### DIFF
--- a/rwengine/include/engine/GameInputState.hpp
+++ b/rwengine/include/engine/GameInputState.hpp
@@ -3,6 +3,8 @@
 
 struct GameInputState
 {
+	static constexpr float kButtonOnThreshold = 0.1f;
+
 	/// Inputs that can be controlled
 	/// @todo find any sensible values
 	enum Control {
@@ -55,6 +57,16 @@ struct GameInputState
 	 * For buttons, this will result in either 0 or 1.
 	 */
 	float currentLevels[_MaxControls] = { };
+
+	float operator[] (Control c) const
+	{
+		return currentLevels[c];
+	}
+
+	bool pressed(Control c) const
+	{
+		return currentLevels[c] > kButtonOnThreshold;
+	}
 };
 
 #endif
--- a/rwengine/include/engine/GameInputState.hpp
+++ b/rwengine/include/engine/GameInputState.hpp
@@ -1,0 +1,60 @@
+#ifndef RWENGINE_GAMEINPUTSTATE_HPP
+#define RWENGINE_GAMEINPUTSTATE_HPP
+
+struct GameInputState
+{
+	/// Inputs that can be controlled
+	/// @todo find any sensible values
+	enum Control {
+		/* On Foot */
+		FireWeapon = 0,
+		NextWeapon,
+		NextTarget = NextWeapon,
+		LastWeapon,
+		LastTarget = LastWeapon,
+		GoForward,
+		GoBackwards,
+		GoLeft,
+		GoRight,
+		ZoomIn,
+		ZoomOut,
+		EnterExitVehicle,
+		ChangeCamera,
+		Jump,
+		Sprint,
+		AimWeapon,
+		Crouch,
+		Walk,
+		LookBehind,
+
+		/* In Vehicle */
+		VehicleFireWeapon = FireWeapon,
+		VehicleAccelerate,
+		VehicleBrake,
+		VehicleLeft = GoLeft,
+		VehicleRight = GoRight,
+		VehicleDown,
+		VehicleUp,
+		ChangeRadio,
+		Horn,
+		Submission,
+		Handbrake,
+		LookLeft,
+		LookRight,
+		VehicleAimLeft,
+		VehicleAimRight,
+		VehicleAimUp,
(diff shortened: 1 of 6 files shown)


rwgame/ingamestate.cpp
```diff
@@ -15,12 +15,54 @@
...
 const float kMaxRotationRate = glm::half_pi<float>();
 const float kCameraPitchLimit = glm::quarter_pi<float>() * 0.5f;
 const float kVehicleCameraPitch = glm::half_pi<float>() - glm::quarter_pi<float>() * 0.25f;
 
+// Hardcoded Controls SDLK_* -> GameInputState::Control
+const std::unordered_multimap<int, GameInputState::Control> kDefaultControls = {
+	/* On Foot */
+	{ SDLK_LCTRL,     GameInputState::FireWeapon },
+	{ SDLK_KP_ENTER,  GameInputState::NextWeapon },
+	{ SDLK_KP_PERIOD, GameInputState::LastWeapon },
```
JayFoxRox: Problematic on notebooks without numpad. Not sure if we should change it if it was the default in GTA3 though.

danhedron: There are more alternate buttons for those actions, [according to this](http://gta.wikia.com/wiki/Controls_for_GTA_III). How you're supposed to change weapon if you have no numpad or mouse wheel remains to be seen.

rwengine/include/engine/GameInputState.hpp
```diff
@@ -0,0 +1,58 @@
...
+#define RWENGINE_GAMEINPUTSTATE_HPP
+
+struct GameInputState
+{
+	/// Inputs that can be controlled
+	/// @todo find any sensible values
+	enum Control {
+		/* On Foot */
+		FireWeapon = 0,
+		NextWeapon,  // Also Next Target
```
JayFoxRox: soo.. why not  `NextTarget = NextWeapon,` ?

rwengine/include/engine/GameInputState.hpp
```diff
@@ -0,0 +1,58 @@
...
+		ChangeCamera,
+		Jump,
+		Sprint,
+		AimWeapon,
+		Crouch,
+		Walk,
+		LookBehind,
+
+		/* In Vehicle */
+		VehicleFireWeapon,
```
JayFoxRox: These indices should be the same as the on-foot slots?
If you don't want to do it this way you'll have to seperate onfoot and vehicle controls more cleanly (adding a seperator enum and guaranteeing the order).

The script checks the buttons by index. So I'd assume these indices are equivalent to these at all times (both, onfoot and in vehicle).

danhedron: The buttons for firing while in a vehicle are the same as on foot, so  `VehicleFireWeapon` in unneeded. Some of the directions are also the same. I'll remove the redundant ones.

danhedron:
Rebased to fix conflict

rwengine/include/engine/GameInputState.hpp
```diff
@@ -0,0 +1,60 @@
...
+		Jump,
+		Sprint,
+		AimWeapon,
+		Crouch,
+		Walk,
+		LookBehind,
+
+		/* In Vehicle */
+		VehicleFireWeapon = FireWeapon,
+		VehicleAccelerate,
```
JayFoxRox: Sorry to bug about these again...
I think you missed my point last time (or I missed yours?).

I was asking about aliases like "VehicleFireWeapon = FireWeapon," for *all* enum members.
At the moment VehicleBrake would be a different index than ZoomIn and Jump. However, in the script these [all map to the very same button](http://www.gtamodding.com/wiki/00E1) (because R* used the PS2 layout as the default).

We'll have to decide wether we want an enum for each game function OR if we want to do it like R* (array of 20 inputs).
I proposed the change, thinking that currentLevels would map to the table linked above.

rwgame/ingamestate.cpp
JayFoxRox: Seeing how this maps the original default mapping, shouldn't we also map the alternative buttons here?
(If you don't want to do it now, we should create an issue about it)

danhedron: There's no reason not to add them now.

rwengine/include/engine/GameInputState.hpp
danhedron: I was going to consider reversing the buttons pressed from the current player state and logical input:
```
case CROSS:
    if (player->inVehicle())
        return input.pressed(Accelerate);
    else
        return input.pressed(Sprint);
```
Since the current game input handling assumes that those controls are not both activated.

It'd be possible to only track the physical input against the logical state too, just need to update our input handling to take into account the current player state.

danhedron: On IRC we came to the conclusion that mapping the logical state to the correct button state for the script was the way to go.

PR #169, opened 2016-07-02
Misc Build cleanup

danhedron:
- List source files in rwengine's CMakeLists.txt
- Move RWGame state classes into `states/` directory for organisation.
- Remove the include directory from rwengine as nothing else used that structure.


JayFoxRox:
This should also remove https://github.com/rwengine/openrw/blob/master/rwlib/source/a.out

Can you merge #169 soon? That would also help with appveyor because I can't easily remove files selectively as they are globbed

JayFoxRox:
LGTM

PR #179, opened 2016-07-31
Rework how exiting and cleanup is handled.

darkf:
This involves a few changes. The first changes involve
allocating GameWindow and WorkContext on the heap, so that
RWGame still owns them but chooses when they're freed.

The work queue is given a method to stop the worker thread
without destroying the work context, so that subsystems
relying on the work context may still function to shut down.

Then RWGame is rearranged to cleanup separate subsystems
in an order that does not conflict (i.e., stop the work queue,
shut down other subsystems, then the renderer, *then* the window.)

The window needs to be cleaned up *after* the renderer because it
owns the OpenGL context.

In addition, exiting is now performed by emptying the state queue, instead of
RWGame relying on polling the GameWindow.

### DIFF
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -35,6 +35,7 @@ StdOutReciever logPrinter;
 RWGame::RWGame(int argc, char* argv[])
 	: config("openrw.ini")
 	, state(nullptr), world(nullptr), renderer(nullptr), script(nullptr),
+	window(nullptr), work(nullptr),
 	debugScript(false), inFocus(true),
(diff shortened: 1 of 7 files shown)


danhedron:
Still works on Linux, looks good to me. +1

darkf:
Awesome. @JayFoxRox, looks good to merge?

I plan on doing further cleanups in the future (things like refactoring code to use more smart pointers, reducing memory leaks, moving to more idiomatic C++, reducing UB, etc.), this is one step of it.

PR #180, opened 2016-08-02
De-initialize OpenAL; open binary files in binary mode; some warning cleanup

darkf:


rwgame/RWGame.cpp
```diff
@@ -457,7 +457,7 @@ void RWGame::tick(float dt)
 		}
 		
 		// Clean up old VisualFX
-		for( int i = 0; i < world->effects.size(); ++i )
+		for( ssize_t i = 0; i < static_cast<ssize_t>(world->effects.size()); ++i )
```
JayFoxRox: The `--i` in the loop is pretty stupid. There is really no sane reason why this must be signed.

Ideally this should be done differently?
(While loop and only increment `i` if no effect had to be destroyed?)

rwgame/RWGame.hpp
```diff
@@ -27,12 +27,12 @@ class RWGame
 	// must be allocated after Logger setup.
 	GameRenderer* renderer;
     ScriptMachine* script;
+	GameWindow *window;
```
JayFoxRox: What's changed here?!

rwgame/ingamestate.cpp
```diff
@@ -101,7 +101,6 @@ IngameState::IngameState(RWGame* game, bool newgame, const std::string& save)
 void IngameState::startTest()
 {
 	auto playerChar = getWorld()->createPlayer({270.f, -605.f, 40.f});
-	auto player = new PlayerController(playerChar);
```
JayFoxRox: What's happening in the controller contructor?

This feels like it's an accidental change.

rwengine/src/audio/SoundManager.cpp
```diff
@@ -7,6 +7,45 @@
...
+	if(alContext) {
+		alcMakeContextCurrent(NULL);
+		alcDestroyContext(alContext);
+	}
+
+	if(alDevice)
+		alcCloseDevice(alDevice);
+}
+
+bool SoundManager::initializeOpenAL()
```
JayFoxRox: Diff beyond this point is same as original? (line endings or something?)

rwgame/ingamestate.cpp
danhedron: It's created in createPlayer

rwgame/RWGame.cpp
paulsapps: Why not ```for( autoi = 0u; i < world->effects.size(); ++i )```

rwengine/src/audio/SoundManager.cpp
darkf: @JayFoxRox I rearranged the constructor to be at the top of the file to be more consistent with the class member layout, then I added the destructor there.

rwgame/ingamestate.cpp
darkf: @JayFoxRox `player` is unused and I don't think PlayerController mutates it without calling any methods.

rwgame/RWGame.cpp
darkf: @JayFoxRox Yes, or destroyEffect should return an iterator (a la `erase()`).

But I'm far too lazy for that, please ignore it for now. ;)

@paulsapps Because the `--i` on `i=0` will overflow.

danhedron: This block isn't great, but it can be cleaned up with a more thorough pass on rwgame.

rwgame/RWGame.hpp
danhedron: Initialization order warning

danhedron:
+1 Looks good to me, still works on Linux.

A few minor things surfaced by this PR are prime candidates for future clean up.

PR #181, opened 2016-08-03
Torch the HTTP script debugger

darkf:
0-0 Wear your safety goggles, and don't breathe this!


danhedron:
+1 didn't burn down my house.

PR #182, opened 2016-08-04
Switch filesystem handling to boost::filesystem

darkf:
This moves away from using the POSIX file APIs and brings in nicer path handling, shorter code, and native compatibility with Windows.

All of the code using POSIX file code is rewritten to use `boost::filesystem`.

It works fine on my machine (running Windows), but I would ask that it be tested on Linux (or another platform with a case-sensitive filesystem). I am also unable to run the tests due to a bug with boost on MinGW.


darkf:
Not sure what is causing the build failure with Travis, if someone could see if it compiles on their Linux machine that would be nice.

paulsapps:
I think you should be using a namespace alias due to the fact that filesystem is part of the standard now and implemented in MSVC directly. Looks like the travis failure is due to a missing ```#include <iterator>```` or something, or the boost API isn't the same.

danhedron:
This patch makes it compile on my 14.04 install https://gist.github.com/danhedron/45c292e39e7981a8d82da64c894f27ac

darkf:
:+1: @danhedron for fixing FileIndex! :-1: @danhedron for breaking findPathRealCase!

Thanks, looks good now, Travis finally agrees.

danhedron:
This breaks one test case
```/home/dan/dev/openrw/tests/test_FileIndex.cpp(41): error in "test_index_archive": check data.archive == "gta3.img" failed [gta3 != gta3.img]
```
Looks like indexArchive is dropping the extension from the filename.

rwlib/source/platform/FileIndex.cpp
```diff
@@ -1,121 +1,93 @@
...
 void FileIndex::indexArchive(const std::string& archive)
 {
 	// Split directory from archive name
-	auto slash = archive.find_last_of('/');
-	auto directory = archive.substr(0, slash);
-	auto archivebasename = archive.substr(slash+1);
-	auto archivepath = directory + "/" + archivebasename;
+	path archive_path = path(archive);
+	path directory = archive_path.parent_path();
+	path archive_basename = archive_path.stem();
```
danhedron: I think this could just be archive_path.filename() to fix the test.

darkf:
@danhedron OK, done, does it pass now?

danhedron:
Looks good to me now. +1

PR #183, opened 2016-08-04
improve jump speed, fix double-jump

flatr0ze:
Just a small patch.
It prevents the character from making that double-jump and also makes the jump behave closer to how it is in the original engine.

### DIFF
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -100,11 +100,11 @@ void CharacterObject::createActor(const glm::vec2& size)
 		physObject->setCollisionShape(physShape);
 		physObject->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
 		physCharacter = new btKinematicCharacterController(physObject, physShape, 0.30f, 2);
-		physCharacter->setFallSpeed(20.f);
+		physCharacter->setFallSpeed(120.f);
 		physCharacter->setUseGhostSweepTest(true);
 		physCharacter->setVelocityForTimeInterval(btVector3(1.f, 1.f, 1.f), 1.f);
 		physCharacter->setGravity(engine->dynamicsWorld->getGravity().length());
-		physCharacter->setJumpSpeed(5.f);
+		physCharacter->setJumpSpeed(4.f);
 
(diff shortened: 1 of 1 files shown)


JayFoxRox:
To be honest, this feels bikesheddy. I'll still review it.

---

I've ran this and this, indeed, fixes the "double jump". However, if you hold the space bar, the char will still start the jump again.
So this change is incomplete? What's the behavior of the original game?

---

I'm not sure about the FallSpeed and JumpSpeed changes. I'm not entirely sure what the FallSpeed is (is it a factor on gravity somehow?).
However, I feel that the Char still falls too slow (at 30 FPS near the airport) and I can't really judge the jump height. We should probably just record the jump height from the original game instead (by logging velocity and position of a player during a jump OR proper reverse engineering) and refine this once and for all, rather than plugging magic numbers.

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -100,11 +100,11 @@ void CharacterObject::createActor(const glm::vec2& size)
 		physObject->setCollisionShape(physShape);
 		physObject->setCollisionFlags(btCollisionObject::CF_KINEMATIC_OBJECT);
 		physCharacter = new btKinematicCharacterController(physObject, physShape, 0.30f, 2);
-		physCharacter->setFallSpeed(20.f);
+		physCharacter->setFallSpeed(120.f);
```
danhedron: Magic numbers like this should be raised to constants and defined in an appropriate place (such as the top of the file).

I've updated the [coding conventions](https://github.com/rwengine/openrw/wiki/Coding-Conventions) for naming constants.

danhedron:
As @JayFoxRox says, this should be compared with the behaviour of the original game to see how it measurably improves the behaviour.

danhedron:
Is there any progress on this? I'm not concerned with fixing the jump speed of the character controller, since the bullet controller will need replacing with a custom implementation for accuracy.

Fixing jump behaviour on the other hand would be very nice, since that'll still be useful when the bullet implementation is replaced.

flatr0ze:
I can no longer observe double jump happening unless I attempt to move the character while gliding as mentioned in issue #208 
This PR may be closed as I should've probably created an issue instead of a PR since I did not have the right numbers to port the jumping precisely as it is in the original game. I apologize for that.

PR #184, opened 2016-08-05
Fix the construction of WorkContext

darkf:
Previously the worker thread was constructed before the queue was, leading to
unfortunate race conditions. This fixes that, along with unrelated minor cleanup.

### DIFF
--- a/rwlib/source/job/WorkContext.hpp
+++ b/rwlib/source/job/WorkContext.hpp
@@ -1,6 +1,4 @@
 #pragma once
-#ifndef _LOADCONTEXT_HPP_
-#define _LOADCONTEXT_HPP_
 
 #include <queue>
 #include <thread>
@@ -8,7 +6,6 @@
 #include <atomic>
 #include <memory>
 #include <functional>
-#include <fstream>
 #include <iostream>
 
 class WorkContext;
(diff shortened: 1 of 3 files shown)


rwlib/source/job/WorkContext.hpp
```diff
@@ -69,13 +63,13 @@ class GameWorld;
...
-	std::unique_ptr<LoadWorker> _worker;
+	std::mutex _inMutex;
+	std::mutex _outMutex;
 
 	std::queue<WorkJob*> _workQueue;
 	std::queue<WorkJob*> _completeQueue;
 
-	std::mutex _inMutex;
-	std::mutex _outMutex;
+	std::unique_ptr<LoadWorker> _worker;
```
JayFoxRox: Might be worth adding a comment somewhere that the order is important here.
I guess the average user wouldn't think of it.

darkf: @JayFoxRox done

JayFoxRox:
Seems to fix the crashes I've been getting lately.
Tested in Test mode and started / quit repeatedly. Not a single crash :)
Code also reviewed.

LGTM

PR #185, opened 2016-08-05
Travis: Build all tools

JayFoxRox:
This causes the script tool and data viewer to be compiled on Travis CI.
This hopefully avoid breaking some of the non-game components of OpenRW in future-PRs.

It causes slight slowdowns due to the dependency on Qt which is a rather large package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,18 @@ addons:
       - libglm-dev
       - libsndfile-dev
       - libopenal-dev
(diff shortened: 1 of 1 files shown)


JayFoxRox:
This conflicts with changes in #182 which uses travis internal package options from the travis.yml, rather than manually upgrading packages. I'll update this.

**tl;dr: DON'T MERGE THIS**

PR #186, opened 2016-08-08
Generate Script Functions and script cleanup

danhedron:
To fix:
- [x] Strange man appears in kurma during intro
- [x] Fading doesn't work correctly
- [x] Door doesn't rotate
- [x] Script stops after re-entering kurma in cutscene
- [x] Double kurma in the introduction mission due to a problem ``00d8``
- [x] Misty never enters the car
- [x] Misty's radar blip is at the origin

Issues:
- Object replacement is removed, as it was hacky originally.


rwengine/src/script/ScriptModule.cpp
```diff
@@ -13,7 +13,7 @@ void ScriptModule::bind(ScriptFunctionID id, ScriptFunction func, bool condition
 				desc
 			}
 		}
-	);
+				);
```
JayFoxRox: What happened here?

Oh right.. "tabs are better than spaces" happened here.

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,216 @@
...
+	{
+		functions.insert({id, {
+							  [=](const ScriptArguments& args) {
+								script_bind::do_unpacked_call(function, args);
+							  },
+							  args,
+							  false,
+							  "opcode",
+							  ""
+						  }});
```
JayFoxRox: The tab strikes again!

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,216 @@
...
+
 private:
 	const std::string name;
 	std::map<ScriptFunctionID, ScriptFunctionMeta> functions;
 };
 
+template <>
+void ScriptModule::bind<void(const ScriptArguments&)>(ScriptFunctionID id,
+		  int args,
+		  void(*function)(const ScriptArguments&));
```
JayFoxRox: Alignment must be done with spaces.

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,216 @@
...
+
 	void bind(ScriptFunctionID id,
-	          ScriptFunction func,
-	          bool conditional,
-	          int args,
-	          const std::string& name,
-	          const std::string& desc
+			  ScriptFunction func,
+			  bool conditional,
+			  int args,
```
JayFoxRox: nit: I'd prefer use of argc rather than args. We can do it later though

rwengine/src/script/modules/GameModule.cpp
```diff
@@ -803,7 +803,7 @@ void game_create_crusher_crane(const ScriptArguments& args)
 	// NOTE: This instruction also creates or controls a garage
 	/// @todo find out if this creates a garage or if it just controls garage[0]
 	args.getWorld()->state->garages.push_back({
-		min, max, garageType
+		0, min, max, garageType
```
JayFoxRox: What does the extra parameter do, and why is this part of this PR?

rwengine/src/script/ScriptModule.hpp
danhedron: oops; forgot to reconfigure clang-format after upgrading qtcreator.

JayFoxRox:
I've skimmed over the code so far and it looks mostly good except for formatting errors.
I also think commits like "Cleanup SCMFile" or "Remove scripttool" should have been done *before* this PR. I'm also not sure wether we should keep scripts and tools in the main repo.
\- However, this is mostly nit-ting and we can just change it later if we want.


I'll still have to run and review this locally before I can give better feedback.

rwengine/src/script/modules/GameModule.cpp
danhedron: It's used to determine the number to write to the script output when assigning to a ScriptGarage.

JayFoxRox:
(Note: I'm using your branch directly, not merged/rebased onto master)

First mission: Can't pick up misty. She just stands around.
Also her marker seems to be in the middle of nowhere. I believe we fixed this issue on master before?

Also, hopefully unrelated, after teleporting, the doors of the kuruma started spinning uncontrollably - that never happened to me before? - wasn't that fixed in master?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -9752,7 +9752,7 @@ void opcode_034c(const ScriptArguments& args) {
 	@arg angle1 Angle
 	@arg arg4 Boolean true/false
 */
-void opcode_034d(const ScriptArguments& args, const ScriptObject object, const ScriptFloat angle0, const ScriptFloat angle1, const ScriptBoolean arg4) {
+bool opcode_034d(const ScriptArguments& args, const ScriptObject object, const ScriptFloat angle0, const ScriptFloat angle1, const ScriptBoolean arg4) {
```
JayFoxRox: How did this happen?
Error in XML or some error in script?

PR #187, opened 2016-08-08
Fix RWViewer

JayFoxRox:
3 Simple changes to RWViewer:

- Init GL 3.3 core context (code stolen from Qt sample page; fixes error in GL code)
- Don't load models/txd as we might not have that file (fixes assert or crash, can't remember)
- And finally don't hog the CPU (My machine can't keep up with 16ms interval so events will stack up. When opening a file dialog this means a framerate of about 1 FPS for the entire process which makes it impossible to select files)

The last one is a bit hacky. Ideally we'd only redraw if necessary and sync to VSYNC somehow.
But: we can fix that later.


danhedron:
Looks good to me. +1

PR #188, opened 2016-08-09
Script cleanup and remove old tool

danhedron:
The script tool is obsolete in the face of better tools, so it can be remove since it's just dead code.

Also cleans up some code in SCMFile just because now seems like a good time.


rwengine/src/script/ScriptMachine.cpp
```diff
@@ -148,10 +148,15 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
 #if RW_SCRIPT_DEBUG
 		if (strcmp(t.name, "EIGHT") == 0)
```
JayFoxRox: Can we get rid of this or move the thread name somewhere else where it's easier to change it?

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -148,10 +148,15 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
...
-			printf("% 8s  %04x %04x % 25s", t.name, t.programCounter, opcode, code.signature.c_str());
+			printf("% 8s %01x %04x %04x % 25s", t.name, t.conditionResult, t.programCounter, opcode, code.signature.c_str());
 			for (auto& a : sca.getParameters())
 			{
-				printf(" %08x", a.integerValue());
+				if (a.type == SCMType::TString) {
+					printf(" %01x:% 8s", a.type, a.string);
+				}
+				else {
+					printf(" %01x:%08x", a.type, a.integerValue());
```
JayFoxRox: This will often print out long error messages because the type is not handled properly.
We also need something for realValue();

Would it be possible to create something for this PR already?

I had a similar commit in my cleanup but I'd drop it in favor of this one.

JayFoxRox:
Skimmed over the code which looks good.
I'll have to grep this locally and also want to test it soon.

danhedron:
I've added an environment variable switch to control the thread name, or it'll print out everything and added the real value output as well.

JayFoxRox:
Code reviewed and ran the game:

Missing script-tool removal from https://github.com/rwengine/openrw/blob/master/.travis.yml#L24

Aside from that LGTM

---

I agree that the script-tool should be removed from the main repo.
However, in the future gta3sc will become a full decompiler without opcode index output. We might want a cross-platform dis*assembler* then. We should co-ordinate with thelink that the disassembler will be a feature which will still exist later. If he does not want to do that (understandably) we should keep a patch ready (Forking from the OpenRW project is not advisable due to the xml files).

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -146,12 +146,21 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
         }
 
 #if RW_SCRIPT_DEBUG
-		if (strcmp(t.name, "EIGHT") == 0)
+		static auto sDebugThreadName = getenv("OPENRW_DEBUG_THREAD");
+		if (!sDebugThreadName || strncmp(t.name, sDebugThreadName, 8) == 0)
 		{
-			printf("% 8s  %04x %04x % 25s", t.name, t.programCounter, opcode, code.signature.c_str());
+			printf("%8s %01x %04x %04x %25s", t.name, t.conditionResult, t.programCounter, opcode, code.signature.c_str());
```
JayFoxRox: 25s is really short. we do have stuff like "game_character_in_vehicle" now which will break the alignment.
Just drop the alignment here. Left alignment is way easier to read anyway.

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -146,12 +146,21 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
...
 			{
-				printf(" %08x", a.integerValue());
+				if (a.type == SCMType::TString) {
+					printf(" %1x:%8s", a.type, a.string);
+				}
+				else if (a.type == SCMType::TFloat16) {
+					printf(" %1x:%f", a.type, a.realValue());
+				}
+				else {
+					printf(" %1x:%8x", a.type, a.integerValue());
```
JayFoxRox: Also remove the alignment on the integers OR align them using zero (eg. "%08x").

danhedron:
I fixed the alignment and dropped script tool from travis.

Also touched some of the include guards, bringing them up to the new style since this is a good a time as any.

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -146,12 +146,21 @@ void ScriptMachine::executeThread(SCMThread &t, int msPassed)
...
+				}
+				else if (a.type == SCMType::TFloat16) {
+					printf(" %1x:%f", a.type, a.realValue());
+				}
+				else {
+					printf(" %1x:%08x", a.type, a.integerValue());
+				}
 			}
 			printf("\n");
 		}
```
JayFoxRox: I just realized the indentation for the entire block is wrong

danhedron: Should be fixed now.

danhedron:
I cleaned up more of the source in ScriptMachine, there was a lot of cruft in there now.

I've run clang-format on the file as well, based on the Google style to see if this style is agreeable.

rwengine/src/script/ScriptMachine.cpp
```diff
@@ -32,7 +32,6 @@ void ScriptMachine::interuptNext()
     interupt = true;
 }
 
-#include <iostream>
```
JayFoxRox: nit: Might want to drop this commit and do a full lint / include-what-you-use pass (I've also done one for appveyor but only added new includes / not removed useless ones unless blatantly obvious)

JayFoxRox:
Code reviewed.
(gj github, thinking my diff comment is outdated.. again. Check my last one please for a nit.)

I did not check wether this removed all remaining parts of the script-tool and debugger.
I did not run tests.
I'll trust you with those.

The style was fine, nothing shady going on and generally nice cleanup :)

Code ran.
Works fine and output is a lot nicer.

This LGTM.

christoph-heiss:
Builds & runs fine on macOS, changes seems fine.

Maybe the removed `#pragma once` in `SCMFile.hpp`, `ScriptMachine.hpp` and `ScriptTypes.hpp` _could_ be left in place, as they don't hurt and can only improve compilation time.

But I think this can go as-is, so +1 from me.

PR #191, opened 2016-08-10
Cleanup and tiny features

JayFoxRox:
This is not a pure cleanup (= some commits change behaviour) as I had to fix some smaller issues and annoyances.
The reason for all changes is higher compatibility with compilers and environments, increase code-readability and preventing crashes or UB.

(Don't be scared away; most of the ~250 lines in this commit are from a FindSDL2.cmake replacement. I recommend review per commit. Thanks!)

---

* Use FindSDL2.cmake from OpenMW
Seems to work better than that from Citra. I can't remember the specifics at the moment though.

* Use bullet include path found by CMake module
We used to craft our own path here and had a windows specific hack

* Fix SDL2 and Bullet #include paths
Uses what's been suggested in the CMake modules
(bullet include dir is a mess with duplicated include files accross subdirs etc.)

* Removes empty file (0 bytes) rwengine/core/FileArchive.hpp
We should probably grep for binaries and empty files on travis? (Seperate issue)

* Include-What-You-Use
Includes the std headers if we use some of their functions (breaks MSVC otherwise)

* Replace non-std C++ code with alternatives
This is enough to make VS2015 compile. However, **I removed one function without replacement.
I'm not sure wether this was overloaded and was used anywhere?** The game still works fine for me.

* Switch default english.gxt to american.gxt
There is no point in using english.gxt as the later games use american.gxt which is essentially the same.

* Extend the previous commit by allowing language selection in openrw.ini
  Needs extension of the wiki to include:
  ```
[game]
language=american
```

* Avoid error in substr if short modelname in LOD selection
LOD selection does a substr(3) to skip the first characters 4, meaning it would crash if the string was shorter than 4 chars. Might appear in mods.

* Avoid crash if `scriptOnMissionFlag` is not linked by script
This occurs in `game_set_dead_or_arrested` which probably shouldn't depend on that flag?
We should probably fix this properly later. However, for now, this hack will be fine.
Only happens with mods, not the original game.

* Replace strcasecmp with boost::iequals which is more portable

* Don't grab mouse cursor before going ingame
Mostly to make development easier (the cursor would be released during the main menu again anyway). We probably should hide it though. We need a better API for that (Seperate issue).

* Simplify logic of player_near_point script functions 0xF5 - 0xFA and introduce missing versions of that
The old code was overly complex - these opcodes are mostly variations of 00F5

* Script: Initialize globals from SCMFile
Because.. why not? This is unverified though. We should check if this actually happens with the original game and if the code is correct or off-by-one or something (Seperate issue).

---

Please review for completeness, style consistency and **check if the new cmake/include setup works for your system**.

### DIFF
--- a/rwengine/src/script/modules/GameModule.cpp
+++ b/rwengine/src/script/modules/GameModule.cpp
@@ -116,6 +116,9 @@ bool game_is_button_pressed(const ScriptArguments& args)
 
 void game_set_dead_or_arrested(const ScriptArguments& args)
 {
+	if (args.getWorld()->state->scriptOnMissionFlag == nullptr) {
+		return;
+	}
 	*args.getWorld()->state->scriptOnMissionFlag = args[0].integer;
 }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,8 +41,6 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
 	set(OPENRW_PLATFORM_LIBS iconv)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
 	add_definitions(-DRW_WINDOWS)
-	set(BULLET_ROOT "$ENV{BULLET_ROOT}")
-	include_directories("$ENV{BULLET_ROOT}/src")
 else ()
 	message(FATAL_ERROR "Unknown platform \"${CMAKE_SYSTEM_NAME}\". please update CMakeLists.txt.")
 endif ()
@@ -99,6 +97,7 @@ include_directories(
 	${SDL2_INCLUDE_DIR}
 	${GLM_INCLUDE_DIRS}
 	${OPENGL_INCLUDE_DIR}
+	${BULLET_INCLUDE_DIRS}
 )
 
 # External-internal dependencies
--- a/cmake/modules/FindSDL2.cmake
+++ b/cmake/modules/FindSDL2.cmake
@@ -1,4 +1,4 @@
-
+# Locate SDL2 library
 # This module defines
 # SDL2_LIBRARY, the name of the library to link against
 # SDL2_FOUND, if false, do not try to link to SDL2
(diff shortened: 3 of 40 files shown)

### CODE REVIEW

**Comment by christoph-heiss** [comment]
Still compiles and runs fine on macOS (10.12).

Also, what do you imagine under a better API for mouse (un){grab,hide}ing?

**Comment by JayFoxRox** [comment]
Currently we only have:
```
void GameWindow::showCursor();
void GameWindow::hideCursor();
```

which will not only hide/show the mouse but also grab it (centering it in the OpenRW window - systemwide.. so if OpenRW grabs it the window will steal focused until you tab out).
It also switches between relative and absolute mouse mode.
A better API would group relative/absolute+mouse grab and then another command to show and hide the cursor. We'll likely draw our own mouse cursor soon so we would want to hide the cursor, but not grab it, also keep absolute position.
This should be discussed in another issue though. So I'd appreciate if someone creates one.

File: rwengine/src/script/modules/GameModule.cpp [hunk 1]
```diff
@@ -116,6 +116,9 @@ bool game_is_button_pressed(const ScriptArguments& args)
 
 void game_set_dead_or_arrested(const ScriptArguments& args)
 {
+	if (args.getWorld()->state->scriptOnMissionFlag == nullptr) {
+		return;
+	}
```
danhedron: This entire function doesn't seem correct at all. This just controls the behaviour of opcode 0112 within missions.

danhedron:
> Simplify logic of player_near_point script functions 0xF5 - 0xFA and introduce missing versions of that
The old code was overly complex - these opcodes are mostly variations of 00F5

Good cleanup. This will be replaced by some of the re-factoring done in the new generated script files though, I'll make sure that code is at least as clean.

rwengine/src/script/modules/GameModule.cpp
JayFoxRox: We'll create an issue after merge

danhedron:
+1 from me.

JayFoxRox:
Rebased due to merge conflict in CMakeLists.txt after OpenBSD was added.

Issue #59, opened 2016-05-23
Warnings during the compilation process (Mac)


gbudny:
Hi

I want to inform you that I noticed warnings during the compilation process on the Mac:

uname -a
Darwin Grzegorzs-Mac-mini.local 14.5.0 Darwin Kernel Version 14.5.0: Tue Sep 1 21:23:09 PDT 2015; root:xnu-2782.50.1~1/RELEASE_X86_64 x86_64

https://gist.github.com/gbudny/69b41f3c70585be421d499b2a1ebb3f6

JayFoxRox:
Can some mac dev shed some light on this / fix it?

paulsapps:
```
 passing 'char [14]' to parameter of type 'const GLubyte *' (aka
      'const unsigned char *') converts between pointers to integer types with
      different sign [-Wpointer-sign]
```

Exactly what it says on the tin.. casting signed char* to unsigned char*.

JayFoxRox:
Related to #195

PR #195, opened 2016-08-10
Some more cleanups

christoph-heiss:
This fixes most of the annoying warnings.

Please have a look especially at commit 95a7e73a2af7e909c6e60e02dfb80549eaf60cd8, as I'm not sure if this is the best solution for this problem.

### DIFF
--- a/rwengine/src/objects/VehicleObject.hpp
+++ b/rwengine/src/objects/VehicleObject.hpp
@@ -6,7 +6,7 @@
 #include <objects/VehicleInfo.hpp>
 
 class CollisionInstance;
-class btVehicleRaycaster;
+struct btVehicleRaycaster;
 class btRaycastVehicle;
 class btRigidBody;
 class btHingeConstraint;
--- a/rwengine/src/engine/GameState.hpp
+++ b/rwengine/src/engine/GameState.hpp
@@ -243,8 +243,9 @@ struct GarageInfo
  * Gameplay state object that holds persistent state, and references runtime
  * world state.
  */
-struct GameState
+class GameState
 {
+public:
 	/**
 	  Basic Game State
 	 */
--- a/rwengine/src/render/GameRenderer.hpp
+++ b/rwengine/src/render/GameRenderer.hpp
@@ -21,7 +21,7 @@ class ModelFrame;
 class GameWorld;
 class GameObject;
 
-class AreaIndicatorInfo;
+struct AreaIndicatorInfo;
 
 /// @todo migrate to some other way of rendering each object type.
 class CharacterObject;
--- a/rwengine/src/render/OpenGLRenderer.hpp
+++ b/rwengine/src/render/OpenGLRenderer.hpp
@@ -247,33 +247,33 @@ public:
 
 	OpenGLRenderer();
 
-	std::string getIDString() const;
+	std::string getIDString() const override;
 
-	ShaderProgram* createShader(const std::string &vert, const std::string &frag);
-	void setProgramBlockBinding(ShaderProgram* p, const std::string &name, GLint point);
-	void setUniformTexture(ShaderProgram* p, const std::string &name, GLint tex);
-	void setUniform(ShaderProgram* p, const std::string& name, const glm::mat4& m);
-	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec4& m);
-	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec3& m);
-	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec2& m);
-	void setUniform(ShaderProgram* p, const std::string& name, float f);
-	void useProgram(ShaderProgram* p);
+	ShaderProgram* createShader(const std::string &vert, const std::string &frag) override;
+	void setProgramBlockBinding(ShaderProgram* p, const std::string &name, GLint point) override;
+	void setUniformTexture(ShaderProgram* p, const std::string &name, GLint tex) override;
+	void setUniform(ShaderProgram* p, const std::string& name, const glm::mat4& m) override;
+	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec4& m) override;
+	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec3& m) override;
+	void setUniform(ShaderProgram* p, const std::string& name, const glm::vec2& m) override;
+	void setUniform(ShaderProgram* p, const std::string& name, float f) override;
+	void useProgram(ShaderProgram* p) override;
 
-	void clear(const glm::vec4 &colour, bool clearColour, bool clearDepth);
+	void clear(const glm::vec4 &colour, bool clearColour, bool clearDepth) override;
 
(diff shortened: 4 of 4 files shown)


rwlib/CMakeLists.txt
```diff
@@ -41,10 +41,9 @@ SET(RWLIB_SOURCES
 	"source/job/WorkContext.cpp"
 	)
 
-add_library(rwlib
+set_source_files_properties(source/gl/gl_core_3_3.c PROPERTIES COMPILE_FLAGS -w)
```
JayFoxRox: Instead, aren't we able to somehow generate a better source file to begin with?
If this is an issue for us, it's probably for others too. Has this been fixed in glLoadGen since?

christoph-heiss: Well, other than opening an issue at https://bitbucket.org/alfonse/glloadgen/wiki/Home, I do not really see a solution (except fixing the code by hand).

I think this is the simplest solution and should be fine for now, because this file shouldn't be touched anyway and thus dropping all warnings is harmless.

JayFoxRox: https://bitbucket.org/alfonse/glloadgen/issues/82/compiling-warning

So we'll need

```
extern "C" {
#include "foo.h"
}
```

if we don't have it yet + compile the file differently (different std flags).

I'm sure there is a proper solution for this issue, so I don't feel like settling for less.

danhedron: The cpp formats introduce a namespace. Some intermediate header that sorts out the linkage or imports the namespace seems like a reasonable solution.

christoph-heiss: glLoadGen seems unmaintained - its last commit was on 2015-11-12.

On macOS, there is also this warning:
`warning: 'NSAddImage' is deprecated: first deprecated in OS X 10.5
      [-Wdeprecated-declarations]`
<hr>
Although compiling it as C++ source is not technically correct, it works - and does not need an intermediate header.
Compiling it as C source would just introduce some more CMake setup for one file. 

As I'm not sure how to solve this properly, I can just drop f1fba2a71d544ffe0ce9036b23a413f20a3db84d for now and open an issue.
I don't think this is a high-priority issue.

JayFoxRox: This sounds like a reasonable idea. Sorry for the inconvenience :disappointed: 
There is already #59 

Maybe glloadgen is simply considered "complete"? I'll try to find out or look for alternatives. I'll also try to convince them to include styles which might not be nice, but work error-free for modern C++. I'll also still look into the cmake alternative. (but everyone else is free to do it too, as I'm super busy during the next months)

rwengine/src/ai/CharacterControllerActivities.hpp
```diff
@@ -0,0 +1,82 @@
...
+	class act_name : public CharacterController::Activity                 \
+	{                                                                     \
+	public:                                                               \
+		static constexpr auto ActivityName = #act_name;               \
+		std::string name() const override { return ActivityName; };   \
+		bool update(CharacterObject*, CharacterController*) override;
+
+#define DECL_ACTIVITY_CANSKIP(act_name)                          \
+	DECL_ACTIVITY(act_name)                                  \
+		bool canSkip(CharacterObject *) const override;
```
JayFoxRox: Urgh, I think this is really ugly, but it's fine for this PR.

Generally: I don't even like that we had the `DECL_ACTIVITY` macro to begin with.
This extends on the idea of using macros; which is something we shouldn't be doing in a modern C++ project.

Are there any alternatives to this? I guess this could be implemented using some kind of template?
(for the future, it's beyond the scope of this PR)

rwengine/src/ai/CharacterControllerActivities.hpp
```diff
@@ -0,0 +1,82 @@
...
+#include <ai/CharacterController.hpp>
+
+
+#define DECL_ACTIVITY(act_name)                                               \
+	class act_name : public CharacterController::Activity                 \
+	{                                                                     \
+	public:                                                               \
+		static constexpr auto ActivityName = #act_name;               \
+		std::string name() const override { return ActivityName; };   \
+		bool update(CharacterObject*, CharacterController*) override;
```
JayFoxRox: This opens a curly brace but doesn't close it.
It will throw of some code-highlighters and it's very confusing while reading the code.
This is not going to get an okay from me.

rwengine/src/ai/CharacterControllerActivities.hpp
```diff
@@ -0,0 +1,82 @@
...
+		const bool jacked;
+	};
+
+
+	DECL_ACTIVITY(ShootWeapon)
+		ShootWeapon(WeaponItem* item) : item(item), fired(false) { }
+
+	private:
+		WeaponItem* item;
+		bool fired;
```
JayFoxRox: I believe the authors intention for prefixing those with `_` in the old code is because they are member variables? (so `m_` would have been more appropriate)
Anyway, your solution is fine because it improves consistency within this file.

rwengine/src/ai/CharacterControllerActivities.hpp
danhedron: Strings and templates don't cooperate very well.

What I have in mind is replacing the string names with a tagged union or variant so we don't allocate an object for each activity.

rwengine/src/ai/CharacterControllerActivities.hpp
christoph-heiss: I already though so.

Well, I don't like this solution either - that's why I asked for feedback specifically for that patch.
Maybe I can come up with something better using templates (or something else - let's see).

So, please do not merge yet. :)

christoph-heiss:
I dropped both commits where the issues should be properly reworked and can be put in an separate PR, so that the remaining commits can be merged meanwhile.

danhedron:
Looks good to me on Ubuntu. +1

rwengine/src/objects/VehicleObject.hpp
```diff
@@ -6,7 +6,7 @@
 #include <objects/VehicleInfo.hpp>
 
 class CollisionInstance;
-class btVehicleRaycaster;
+struct btVehicleRaycaster;
```
JayFoxRox: These are declared in `btBulletDynamicsCommon.h` or `btVehicleRaycaster.h` etc.
I think I remember a discussion with other projects about this forward declaration / protoyping being better though?

Can someone clarify this for me or shall we include here?

JayFoxRox:
Except for what I mentioned this LGTM. I only ran the game a couple of seconds though.
(What I mentioned is out of scope for this PR though, I just wanted to mention it)

rwengine/src/objects/VehicleObject.hpp
danhedron: The intention was to avoid leaking bullet's include paths to targets that link with rwengine (however this is still the case for some headers). The btVehicleRaycaster is removed in the vehicle improvements anyway.

JayFoxRox:
Did this fix #59 or did we drop those parts?

Issue #160, opened 2016-06-26
Weird Scanlines on AMD laptop APU


hcorion:
Hi all, 
I'm getting these weird scaline type effects on rwgame. Here is two screenshots:
The title screen:
![screenshot_20160625_184040](https://cloud.githubusercontent.com/assets/9814000/16360146/f9ba5140-3b05-11e6-9a5e-57e733f49154.png)
And in game:
![screenshot_20160625_184227](https://cloud.githubusercontent.com/assets/9814000/16360147/fd38b76c-3b05-11e6-8f59-5832d5adb2a7.png)
I haven't tested on this machine previously, so I don't know if it was caused by the recent move to sdl2. I will most definitly test this when the newest mesa is released also. The log looks fine:
```
I [Game] Game directory: /opt/games/Grand Theft Auto 3/
Max UBO Size: 65536
UBO Alignment: 256
Max batch size: 512
I [Renderer] OpenGL Renderer Version: 3.3 (Core Profile) Mesa 11.2.2 (GLSL 3.30) Vendor: X.Org
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
I [Data] Loaded 42 zones from /opt/games/Grand Theft Auto 3//data/gta3.zon
I [Game] Started
AL lib: (EE) alc_cleanup: 1 device not closed
```

Here are the laptop's specs:
OS: Arch Linux 64-bit
CPU: Dual Core AMD Athlon II P340
GPU: Mobility Radeon HD 4225/4250 (integrated)
Software: 
KDE Plasma 5.6.5
Linux 4.6.2
Mesa 11.2.2
X11 version 7.7
Gallium 0.4 on AMD RS880 (DRM 2.43.0, LLVM 3.8.0)
GLX Version: 1.4
Max OpenGL core profile version: 3.3
Max OpenGL compat profile version: 3.0

hcorion:
Issue still persists on mesa 12.0.1
Linux 4.6.4
KDE 5.7.1

JayFoxRox:
Can you apitrace and bisect this issue? I'm assuming this is related to water rendering, but I'm not sure.

hcorion:
[...]

PR #197, opened 2016-08-12
Remove screen-door fade-in effect

JayFoxRox:
This applies the proposed patch from #160 to fix rendering on AMD GPUs.
We should document the game behaviour  then implement it properly (if this is not fixed/handled by librw). However, it should be discussed in a new issue, as it's beyond the scope of this PR.

For now it's better to have pop-in, than breaking compatibility with possible users / devs.

---

All credits to @danhedron for investigating and fixing this. I'm only doing the PR


danhedron:
Game still renders for me. +1.

PR #198, opened 2016-08-12
Refactor seat storage, fixes entering wrong seat

### LINKED ISSUE #73: Player sometimes enters wrong vehicle seat
Some vehicles see the player entering the back seat on the left side of the car, instead of the driver seat.
screenshot_2016-05-23_22-38-25

JayFoxRox:
I've tested this on all cars in testmode and it worked fine.
I've only tested entering the driver seat (front left).

The motivation for storing data like this has been that we are able to find front and backseats without iterating over all available seats, while still being able to get any seat easily.

The code replaces one `at()` by `[]` and overloads an `operator[]` elsewhere without bounds checking.
This is written under the assumption that we fix those kind of issues at a later point (lint etc.) or do bounds checking elsewhere.

Fixes #73


danhedron:
Works. +1

PR #203, opened 2016-08-12
Fix entering & exiting vehicle animations

danhedron:
- Don't crash if the seat doesn't have a door
- Make RHS animations count as entering or exiting animations

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -219,12 +219,13 @@ bool Activities::EnterVehicle::update(CharacterObject *character, CharacterContr
 	
 	auto entryDoor = vehicle->getSeatEntryDoor(seat);
 	auto entryPos = vehicle->getSeatEntryPositionWorld(seat);
+	auto entryPosLocal = vehicle->getSeatEntryPosition(seat);
 
 	auto anm_open = character->animations.car_open_lhs;
 	auto anm_enter = character->animations.car_getin_lhs;
 	auto anm_pullout = character->animations.car_pullout_lhs;
 
-	if( entryDoor->dummy->getDefaultTranslation().x > 0.f )
+	if ( entryPosLocal.x > 0.f )
 	{
 		anm_open = character->animations.car_open_rhs;
 		anm_enter = character->animations.car_getin_rhs;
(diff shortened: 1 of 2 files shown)


rwengine/src/objects/CharacterObject.cpp
```diff
@@ -437,8 +437,11 @@ bool CharacterObject::enterVehicle(VehicleObject* vehicle, size_t seat)
 
 bool CharacterObject::isEnteringOrExitingVehicle() const
 {
-	return animator->getAnimation(AnimIndexAction) == animations.car_getout_lhs ||
-			animator->getAnimation(AnimIndexAction) == animations.car_getin_lhs;
+	auto a = animator->getAnimation(AnimIndexAction);
+	return  a == animations.car_getout_lhs ||
+			a == animations.car_getin_lhs ||
+			a == animations.car_getout_rhs ||
+			a == animations.car_getin_rhs;
```
JayFoxRox: nit: alignment gone wrong.

JayFoxRox:
Personally I'd have preferred splitting this PR into 2 commits.
I've skimmed over the code which looks good.

I'll still do a test locally and do a full review soon.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -318,6 +320,8 @@ bool Activities::ExitVehicle::update(CharacterObject *character, CharacterContro
 {
 	RW_UNUSED(controller);
 
+	auto vehicle = character->getCurrentVehicle();
+
```
JayFoxRox: place `exitPos`, `exitPosLocal` and `isRhs` here?

rwengine/src/ai/CharacterController.cpp
```diff
@@ -353,15 +357,17 @@ bool Activities::ExitVehicle::update(CharacterObject *character, CharacterContro
...
 
-	if( character->getCurrentVehicle() == nullptr ) return true;
+	if( vehicle == nullptr ) return true;
 
-	auto vehicle = character->getCurrentVehicle();
 	auto seat = character->getCurrentSeat();
 	auto door = vehicle->getSeatEntryDoor(seat);
+	auto exitPos = vehicle->getSeatEntryPositionWorld(seat);
+	auto exitPosLocal = vehicle->getSeatEntryPosition(seat);
+	auto isRhs = exitPosLocal.x > 0.f;
```
JayFoxRox: same as in the jacked case already, move to suggested point above

rwengine/src/ai/CharacterController.cpp
```diff
@@ -329,20 +333,20 @@ bool Activities::ExitVehicle::update(CharacterObject *character, CharacterContro
...
 		else {
-			if (character->getCurrentVehicle() == nullptr) return true;
+			if (vehicle == nullptr) return true;
 
-			auto vehicle = character->getCurrentVehicle();
 			auto seat = character->getCurrentSeat();
-			auto door = vehicle->getSeatEntryDoor(seat);
+			auto exitPos = vehicle->getSeatEntryPositionWorld(seat);
+			auto exitPosLocal = vehicle->getSeatEntryPosition(seat);
+			auto isRhs = exitPosLocal.x > 0.f;
```
JayFoxRox: get rid of this?

rwengine/src/ai/CharacterController.cpp
danhedron: Vehicle can be null here.

JayFoxRox:
LGTM

Note: Entering boats is super-hard because it warps you inside, however, you still have F pressed so it will instantly cause you to leave again etc. It's beyond the scope of this PR but I felt it's related so I mention it here. Probably deserves it's own issue for boats oddities and boats physics.

danhedron:
I've noticed a few other issues with entering and exiting vehicle caused by the input changes, so that can be noted in an issue about double buffering input to detect presses.

PR #204, opened 2016-08-12
Update contributing guidelines

danhedron:
This just clears up how OpenRW is developed and improves the guidelines to make it clearer how we expect PRs to be evaluated.


JayFoxRox:
LGTM

PR #205, opened 2016-08-13
Add preliminary NetBSD support

krytarowski:
The iconv(3) function on NetBSD has a custom prototype due to old mistake in
the POSIX specification. The issue has been resolved but we keep using original
form with a constified parameter.


danhedron:
This looks good to me.

Replacing Iconv with another portable library is something that should be looked into, since it's not supported on Windows and mean we can remove this alternate path.

JayFoxRox:
Also did not forcefully install some BSD variant on my system.
PC did not catch fire.
RWGame still works.

LGTM

rwengine/src/engine/SaveGame.cpp
```diff
@@ -539,7 +539,11 @@ bool SaveGame::loadGame(GameState& state, const std::string& file)
 	auto icv = iconv_open("UTF-8", "UTF-16");
 	char* saveName = (char*)state.basic.saveName;
 
+#if defined(RW_NETBSD)
+	iconv(icv, (const char**)&saveName, &bytes, &outCur, &outSize);
```
JayFoxRox: nit: avoid C style casts, other instances too

Also: what's different here, except for the cast? Wouldn't this only cause a warning?
What's the signature in NetBSD?

rwengine/src/engine/SaveGame.cpp
```diff
@@ -1286,4 +1294,3 @@ std::vector<SaveGameInfo> SaveGame::getAllSaveGameInfo()
 
 	return infos;
 }
-
```
JayFoxRox: nit: end of file intended for tools like `cat`

rwengine/src/engine/SaveGame.cpp
krytarowski:
```
     size_t
     iconv(iconv_t cd, const char ** restrict src, size_t * restrict srcleft,
         char ** restrict dst, size_t * restrict dstleft);
```

POSIX shifted to const-less version and other OS-es (except NetBSD, ...) keep using const.

Is ther needed `const_cast<>`?

It depends on warning level - if I remember correctly (I patched it 2 days ago) it errored.

rwengine/src/engine/SaveGame.cpp
krytarowski: It's not end-of-line but trailing end-line, is it intended too?

krytarowski: Or rather, trailing empty-line.

rwengine/src/engine/SaveGame.cpp
JayFoxRox: I'm not too knowledgeable about C++, but `const_cast` would cast away the const'ness.
I think this would be a case for `static_cast`. It also works as `dynamic_cast`.
- However, don't change the code, it's just a nit and we'll hopefully replace this code soon anyway.

Also there are plenty of C style casts all over the codebase which can be cleaned up properly another time.

---

A comment / link to the manpage for the NetBSD difference would have been nice btw. However, same reason as above: don't change it, we'll replace that code with some other lib probably.

krytarowski: const_cast<> can add/remove const/volatile

PR #206, opened 2016-08-14
Remove iconv dependency

haphzd:
Test, please.

### DIFF
--- a/rwengine/src/engine/SaveGame.cpp
+++ b/rwengine/src/engine/SaveGame.cpp
@@ -11,7 +11,7 @@
 #include <ai/PlayerController.hpp>
 #include <items/WeaponItem.hpp>
 #include <cstring>
-#include <iconv.h>
+#include <data/RWUnicode.hpp>
 
 #include <boost/filesystem.hpp>
 #include <boost/range/iterator_range.hpp>
@@ -529,22 +529,8 @@ bool SaveGame::loadGame(GameState& state, const std::string& file)
 	READ_VALUE(state.basic)
 
 	// Convert utf-16 to utf-8
-	size_t bytes = 0;
-	for(;; bytes++ ) {
-		if(state.basic.saveName[bytes-1] == 0 && state.basic.saveName[bytes] == 0) break;
-	}
-	size_t outSize = 24;
-	char outBuff[48];
-	char* outCur = outBuff;
-	auto icv = iconv_open("UTF-8", "UTF-16");
-	char* saveName = (char*)state.basic.saveName;
-
-#if defined(RW_NETBSD)
-	iconv(icv, (const char**)&saveName, &bytes, &outCur, &outSize);
-#else
-	iconv(icv, &saveName, &bytes, &outCur, &outSize);
-#endif
-	strcpy(state.basic.saveName, outBuff);
+	auto saveName = RW::UnicodeConverter::utf16_to_utf8((char16_t*)state.basic.saveName, 24);
+	strcpy(state.basic.saveName, saveName.c_str());
 
 	BlockDword scriptBlockSize;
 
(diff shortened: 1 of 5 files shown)


rwengine/src/engine/SaveGame.cpp
```diff
@@ -533,17 +533,12 @@ bool SaveGame::loadGame(GameState& state, const std::string& file)
...
-
-#if defined(RW_NETBSD)
-	iconv(icv, (const char**)&saveName, &bytes, &outCur, &outSize);
-#else
-	iconv(icv, &saveName, &bytes, &outCur, &outSize);
-#endif
+	wchar_t* saveName = (wchar_t*)state.basic.saveName;
+
+	outSize = std::wcstombs(outBuff, saveName, sizeof(outBuff));
+	state.basic.saveName[outSize] = '\0';
```
haphzd: Possible crash if conversion fails. Manual null termination may be not necessary here, because source array is always null terminated?

danhedron:
Unfortunately the size of wchar_t isn't guaranteed to be 16-bits. On Linux it is 32-bits causing this conversion to fail:
```
test_text.cpp(21): error in "load_test": check texts.text("1008") == "BUSTED" failed [ != BUSTED]
```
--------------------
Converting between the UTF-16 data and the UTF-8 output we expect must be explicit. Using http://en.cppreference.com/w/cpp/locale/codecvt for example allows explicit conversion between char16_t and char, but is only available in recent versions of GCC.

JayFoxRox:
Thanks for looking into this! This is one more step towards easy MSVC support :)

---

However, I'm on Arch Linux gcc (GCC) 6.1.1 20160602 and this breaks text for me.
Due to the 2 issues danhedron just mentioned, I'd vote for using boost for this (it's a dependency anyway). Don't aim for the newest boost though: travis, appveyor and some users will be lacking behind so you can't expect the newest software :cry: 

We can rethink this once Ubuntu gets updated with a more recent compiler which doesn't suck as much.
(It's a similar situation with our std::filesystem code which also waits for widespread C++17 and uses boost for now. So probably in 2-3 years with slow migration before)

paulsapps:
Will casting to unsigned short int instead of wchar_t work?

christoph-heiss:
uchar.h doesn't seem to exist on macOS, so the build fails.

Also, if you're using the C library in C++, use the C++ header (aka. cuchar) and thus the std namespace.

haphzd:
This was expected, can't win the boilerplate way I guess.

JayFoxRox:
[This has a singleheader lib](http://stackoverflow.com/questions/6140223/c-boost-encode-decode-utf-8) for the purpose of doing that?
Note that I did not check how credible said library etc is. so some more investigation might be necessary.

I don't know how the conversion works in GTA3 (especially with russian or japanese versions), but this might actually be an okay solution? Other than that I'll remind about boost.

Thanks for trying so hard haph

christoph-heiss:
Maybe you could use [std::wstring_convert](http://www.cplusplus.com/reference/locale/wstring_convert/).

JayFoxRox:
Reminder that `std::wstring_convert` is in locale which is only available in gcc5 which may or may not be installed on every users machine.

We also just dicussed this on irc (I've shortened it)
```
JayFoxRox: openrw is running a 15 year old game
JayFoxRox: so if we can avoid dropping one of the most common OS in the target audience by a single include file lib then that's something to consider
JayFoxRox: also we have boost as alternative which is similar
danhedron: if we only need UCS-2 rather than full blown UTF-16 for the games then that's OK for the interim
JayFoxRox: + we did not even check what the game actually does with the utf16 / ucs2 (which one is it? ucs2 I think? code suggests utf-16 tho)
JayFoxRox: we did not check, so maybe the game was just stupid and all it does is indeed cast char to uint16 before saving files / uint16_t to char on load
danhedron: I'm not sure, I'm just assuming they didn't have anything outside the range of UCS-2
JayFoxRox: so before we raise the requirements for something which we might not even need we should consider this
```

So really stuff like this boils down to wether we want to support Ubuntu 12.04 (LTS still supported until April 2017). I'd vote for yes, because Ubuntu might not be the only platform affected by this.
We already have C++11 and can hop to C++14 soon (probably when Ubuntu 12.04 support runs out as a lot of services will be forced to switch to more recent releases).
The problem with expecting gcc5 *now* is that it will be hard to go back as more patches depending on gcc5 will roll in. OpenRW is still growing so excluding people now also means we might never see them again.

---

The radical option for now would be to add a big fat `// WARNING` and simply casting all uint16 chars to uint8 and asserting if anything is too large.
As we are currently not playing any custom localization or fancy string magic beyond reading the games GXT files this should work.

This might cause issues on japanese or russian games. However, we can also possibly handle those differently later. Unfortunately I don't have access to these localized versions and technical information is sparse :/

---

However, I'll leave this up to the PR author and @danhedron to decide.

danhedron:
Thanks for this @haphzd.

I had a feeling something wasn't quite right with the original code so I did some further looking. The data in the GXT isn't utf-16. It's just an (index into the font texture)[http://www.gtamodding.com/wiki/GXT#GTA_III]. This will need some deeper changes to replace the usage of utf-8 strings with the correct 16-bit strings, since converting those values to UTF-8 from UTF-16 is incorrect.

JayFoxRox:
Given the above, I'll close this.

@danhedron is already working on implementing GXT properly. Sorry for wasting your time @haphzd  :cry:  . Your contributions are still very much welcome + this at least brought our attention to various subjects. Thanks!

Issue #134, opened 2016-06-05
Two 8-Balls in the first mission 


gbudny:
Hi

I noticed this bug in the game:

http://picpaste.com/Untitled13-HAExhpm7.png

JayFoxRox:
This used to happen if 8-Ball is too far away from his respawn position. It usually happened on the second attempt of the mission but it could happen differently too (does it?)

---

OpenRW has had the [bad habit of replacing nearby characters which I have fixed since](https://github.com/rwengine/openrw/commit/340b2f715902d0a4ee5e1de9f32edba2fae573bc).
While this fixed this issue it was also completely wrong and caused (way more serious) bugs in other places (it would delete characters you are supposed to protect for example).

However, due to my change this should *always* happen now. It's a known issue.
- Just ignore useless characters for now.

---

The character *should* be removed either by https://github.com/Lighnat0r/GTA-III-SCM-Converted/blob/master/19_8ball.sc#L817 or the mission_cleanup opcode.
However, OpenRW currently doesn't implement this properly so unnecessary characters / vehicles / objects are kept around.

danhedron:
#207 should fix this by introducing more correct cleanup behaviour that cleans up all types of objects, note just vehicles.

PR #207, opened 2016-08-14
Generate script functions 

danhedron:
This is quite a big change but a lot of the code is generated.

- Add generated script stubs
- New ScriptFunctions file to collect the shared function implementations.
- Fix some bugs
  - #134

- [x] Document python script
- [x] Fix formatting problems.


scripts/generate_scripts_functions.py
```diff
@@ -0,0 +1,338 @@
+import re
```
JayFoxRox: Please add python version information and maybe what the deps are (which can easily be googled or installed through something like pip).
If you ever decide to leave others (we) will have to work with this tool and chances are they (we) are too dumb.

rwengine/src/script/ScriptModule.cpp
```diff
@@ -13,7 +13,7 @@ void ScriptModule::bind(ScriptFunctionID id, ScriptFunction func, bool condition
 				desc
 			}
 		}
-	);
+				);
```
JayFoxRox: .

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,215 @@
...
+	struct unpack<T&, arg>
+	{
+		static T& convert(const ScriptArguments& args)
+		{
+			return args.getParameterRef<T>(arg);
+		}
+	};
+
+	// unpack ScriptObjectType<T>& as non-ref so object is copied
+	// by the caller.
```
JayFoxRox: was this comment break added by clang-format? there is a longer line at line 110 which is not broken?!

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,215 @@
...
-	          ScriptFunction func,
-	          bool conditional,
-	          int args,
-	          const std::string& name,
-	          const std::string& desc
+			  ScriptFunction func,
+			  bool conditional,
+			  int args,
+			  const std::string& name,
+			  const std::string& desc
```
JayFoxRox: broke alignment

rwengine/src/script/ScriptModule.hpp
```diff
@@ -5,34 +5,215 @@
...
+			  Tfunc function)
+	{
+		functions.insert({id, {
+							  [=](const ScriptArguments& args) {
+								script_bind::do_unpacked_call(function, args);
+							  },
+							  argc,
+							  false,
+							  "opcode",
+							  ""
```
JayFoxRox: Very broken alignment

rwengine/src/script/ScriptTypes.hpp
```diff
@@ -26,6 +29,145 @@ typedef uint16_t SCMOpcode;
...
+using ScriptFloat = float;
+using ScriptFloatGlobal = float&;
+using ScriptFloatLocal = float&;
+using ScriptBoolean = int32_t;
+
+using ScriptVec2 = glm::vec2;
+using ScriptVec3 = glm::vec3;
+
+using ScriptRGB = glm::u8vec3;
+using ScriptRGBA = glm::u8vec4;
```
JayFoxRox: Shouldn't a script RGB still consist of 3 script integers?
The bound checking + conversion might be handled elsewhere, I was just wondering.

danhedron: Yes, they should match the range of the input data. Good catch.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2556,9 +2576,16 @@ void opcode_00c6(const ScriptArguments& args) {
...
-	RW_UNIMPLEMENTED_OPCODE(0x00d6);
-	RW_UNUSED(arg1);
-	RW_UNUSED(args);
+	if (arg1 <= 7) {
+		args.getThread()->conditionCount = arg1+1;
+		args.getThread()->conditionMask = 0xFF;
+		args.getThread()->conditionAND = true;
+	}
+	else {
+		args.getThread()->conditionCount = arg1-19;
```
JayFoxRox: beyond the scope:

This probably needs some documentation? Why -19?
This makes it possible to have a negative condition count if 8 <= arg1 <= 18
Are these bitmasks by any chance?

(The `arg1 <= 7` is not very clear to begin with either)

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -16,8 +19,8 @@ void opcode_0000(const ScriptArguments& args) {
 */
 void opcode_0001(const ScriptArguments& args, const ScriptInt time) {
 	RW_UNIMPLEMENTED_OPCODE(0x0001);
-	RW_UNUSED(time);
-	RW_UNUSED(args);
+	auto thread = args.getThread();
+	thread->wakeCounter = time > 0 ? time : -1;
```
JayFoxRox: Beyond the scope of PR: Did we verify the behaviour with negative wait? Shall we assert instead?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -28,8 +31,13 @@ void opcode_0001(const ScriptArguments& args, const ScriptInt time) {
...
 	RW_UNIMPLEMENTED_OPCODE(0x0002);
-	RW_UNUSED(arg1);
-	RW_UNUSED(args);
+	auto thread = args.getThread();
+	if (arg1 < 0) {
+		thread->programCounter = thread->baseAddress - arg1;
+	}
+	else {
+		thread->programCounter = arg1;
+	}
```
JayFoxRox: Maybe beyond the scope: This is formatted differently in all other cases (ternary bool operator). Lines 972, 985, 1030 etc.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2579,8 +2604,19 @@ void opcode_00d7(const ScriptArguments& args, const ScriptLabel arg1) {
...
+		auto obj = args.getWorld()->vehiclePool.find(oid);
+		if( obj )
+		{
+			args.getWorld()->destroyObjectQueued(obj);
+		}
+	}
+	
+	args.getState()->missionObjects.clear();
+	
+	*args.getState()->scriptOnMissionFlag = 0;
```
JayFoxRox: I'm not sure if you also intent to fix formatting with this, but this has a lot of useless line breaks which aren't done in the code above.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6623,16 +6657,13 @@ void opcode_0208(const ScriptArguments& args, const ScriptFloat arg1, const Scri
...
 	@arg arg3 
 */
-void opcode_0209(const ScriptArguments& args, const ScriptInt arg1, const ScriptInt arg2, ScriptInt& arg3) {
-	RW_UNIMPLEMENTED_OPCODE(0x0209);
-	RW_UNUSED(arg1);
-	RW_UNUSED(arg2);
-	RW_UNUSED(arg3);
+void opcode_0209(const ScriptArguments& args, const ScriptInt min, const ScriptInt max, ScriptInt& arg3) {
 	RW_UNUSED(args);
+	arg3 = std::rand() % (max - min) + min;
```
JayFoxRox: Beyond the scope: `max - min` may be negative and negative modulo is tricky in C++ iirc.
We should assert that `max > min` return constant if `max == min`

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -8324,10 +8355,11 @@ bool opcode_02cc(const ScriptArguments& args, const ScriptObject object) {
 	@arg arg2 
 */
 void opcode_02cd(const ScriptArguments& args, const ScriptLabel arg1, const ScriptLabel arg2) {
-	RW_UNIMPLEMENTED_OPCODE(0x02cd);
-	RW_UNUSED(arg1);
 	RW_UNUSED(arg2);
-	RW_UNUSED(args);
+	/// @todo determine what arg2 is used for
```
JayFoxRox: How about *not* making it `RW_UNUSED` then?
Warnings are good if they mean something

(Can we `-Werror` but keep these as warnings?)

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -12662,9 +12693,8 @@ void opcode_0415(const ScriptArguments& args, const ScriptVehicle vehicle, const
 	@arg arg1 
 */
 void opcode_0417(const ScriptArguments& args, const ScriptInt arg1) {
-	RW_UNIMPLEMENTED_OPCODE(0x0417);
-	RW_UNUSED(arg1);
-	RW_UNUSED(args);
+	auto offset = args.getVM()->getFile()->getMissionOffsets()[arg1];
```
JayFoxRox: Beyond the scope: We probably want some bound checking here or the script could do malicious things

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -60,10 +59,8 @@ void opcode_0003(const ScriptArguments& args, const ScriptInt time) {
 	@arg arg2 
 */
 void opcode_0004(const ScriptArguments& args, ScriptIntGlobal arg1G, const ScriptInt arg2) {
-	RW_UNIMPLEMENTED_OPCODE(0x0004);
-	RW_UNUSED(arg1G);
-	RW_UNUSED(arg2);
 	RW_UNUSED(args);
+	arg1G = arg2;
```
JayFoxRox: I'd have assumed that ScriptIntGlobals are like pointers? Are they a reference? If so, the name should suggest that.
It's hard to see that this changes something outside of this function.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -144,10 +131,8 @@ void opcode_0009(const ScriptArguments& args, ScriptFloatGlobal arg1G, const Scr
 	@arg arg2 
 */
 void opcode_000a(const ScriptArguments& args, ScriptIntLocal arg1L, const ScriptInt arg2) {
-	RW_UNIMPLEMENTED_OPCODE(0x000a);
-	RW_UNUSED(arg1L);
-	RW_UNUSED(arg2);
 	RW_UNUSED(args);
+	arg1L += arg2;
```
JayFoxRox: What does the `%2h%` stand for? Does that somehow imply this only works with halfwords?
Can we assert parameter types?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -1136,9 +969,7 @@ bool opcode_0057(const ScriptArguments& args, const ScriptPlayer player, ScriptV
 	@arg arg2G 
 */
 void opcode_0058(const ScriptArguments& args, ScriptIntGlobal arg1G, ScriptIntGlobal arg2G) {
-	RW_UNIMPLEMENTED_OPCODE(0x0058);
-	RW_UNUSED(arg1G);
-	RW_UNUSED(arg2G);
+	arg1G += arg2G;
```
JayFoxRox: Up until this point the RW_UNUSED(args) did appear before the operation. It seems to change at this point?!

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2143,14 +2143,14 @@ bool opcode_00b1(const ScriptArguments& args, const ScriptVehicle vehicle, Scrip
...
-	RW_UNUSED(gxtEntry);
-	RW_UNUSED(time);
-	RW_UNUSED(arg3);
-	RW_UNUSED(args);
+void opcode_00ba(const ScriptArguments& args, const ScriptString gxtEntry, const ScriptInt time, const ScriptInt style) {
+	const auto& text = script::gxt(args, gxtEntry);
+	args.getWorld()->state->text.addText<ScreenTextType::Big>(
+				ScreenTextEntry::makeBig(
+					gxtEntry, text, style, time
+					));
```
JayFoxRox: You might want to move the result of the `ScreenTextEntry::makeBig` to a temporary `auto&` variable to make this code more readable. I first thought the alignment was broken.

rwengine/src/script/ScriptFunctions.cpp
```diff
@@ -28,3 +30,13 @@ const char*script::getBlipSprite(ScriptRadarSprite sprite)
 {
 	return sprite_names[sprite];
 }
+
+ScriptModel script::getModel(const ScriptArguments& args, ScriptModel model)
+{
+	if (model < 0) {
+		/// @todo verify that this is how the game uses negative models
+		const auto& m = args.getVM()->getFile()->getModels()[-model];
```
JayFoxRox: Some range check here?

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -24,9 +27,34 @@ inline void getObjectPosition(GameObject* object,
...
+inline bool isInModel(const ScriptArguments& args,
+                      CharacterObject* character,
+                      int model)
+{
+	auto data = args.getWorld()->data->findObjectType<VehicleData>(model);
+	if (data) {
+		auto vehicle = getCharacterVehicle(character);
+		if (vehicle) {
+			return vehicle->model ? vehicle->model->name == data->modelName
+			                      : false;
```
JayFoxRox: Beyond scope?

I think this is easier to read:

```
if (vehicle->model == nullptr) {
  return false;
}
return vehicle->model->name == data->modelName;
```

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -43,13 +71,59 @@ inline bool objectInBounds(GameObject* object,
...
-bool objectInArea(const ScriptArguments& args, Tobj* object, const Tvec& min, const Tvec& max, bool marker)
+bool objectInArea(const ScriptArguments& args, const Tobj& object, const Tvec& min, const Tvec& max, bool marker, bool preconditions = true)
+{
+	if (marker) {
+		auto center = (min+max)/2.f;
+		auto radius = glm::abs(max-min);
+		drawAreaIndicator(args, center, radius);
+	}
+	return preconditions && objectInBounds(object, glm::min(min, max), glm::max(min, max));
+}
```
JayFoxRox: We could probably sort `min` and `max` first so we don't have to do the `abs` for the radius. We can also add a comment to the ordering line to mention that min and max may not be always ordered.

We should probably also rename these to pointA and pointB or something.

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -43,13 +71,59 @@ inline bool objectInBounds(GameObject* object,
...
+
+inline bool objectInSphere(GameObject* object,
+                           const ScriptVec2& center,
+                           const ScriptVec2& radius)
+{
+	auto p = glm::vec2(object->getPosition());
```
JayFoxRox: Do we have other ways to swizzle this? I think there are more obvious ways to extract `.xy()` ?

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -43,13 +71,59 @@ inline bool objectInBounds(GameObject* object,
...
+
+template<class Tobj>
+inline bool objectInZone(const ScriptArguments& args, Tobj object, const ScriptString zone) {
+	auto zfind = args.getWorld()->data->zones.find(zone);
+	if( zfind != args.getWorld()->data->zones.end() ) {
+		auto& min = zfind->second.min;
+		auto& max = zfind->second.max;
+		return objectInBounds(object, min, max);
+	}
+	return false;
```
JayFoxRox: Beyond the scope: This should probably be part of the object class in the future?

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -6,12 +6,15 @@
...
 #include <objects/GameObject.hpp>
 #include <script/ScriptTypes.hpp>
+#include <script/ScriptMachine.hpp>
 
 /**
  * Implementaitions for common functions likely to be shared
  * among many script modules and opcodes.
  */
 namespace script {
+const ScriptVec3 kSpawnOffset = ScriptVec3(0.f, 0.f, 1.f);
```
JayFoxRox: The typical spawn offset we used back in the day was 0.981 (which would be 1 frame worth of gravity probably)
I remember because we wrote a physics test at school and I used g = 0.981 m/(s*s) because I somehow mixed it up.

Maybe that was the proper offset? I must have heard of it from gtaforums or the MTA forums.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2536,45 +2513,29 @@ bool opcode_00e3(const ScriptArguments& args, const ScriptPlayer player, const S
...
-	RW_UNUSED(player);
-	RW_UNUSED(arg2);
-	RW_UNUSED(arg3);
-	RW_UNUSED(arg4);
-	RW_UNUSED(arg5);
-	RW_UNUSED(arg6);
-	RW_UNUSED(args);
-	return false;
+bool opcode_00e4(const ScriptArguments& args, const ScriptPlayer player, const ScriptVec2 coord0, const ScriptVec2 coord1, const ScriptInt arg6) {
+	auto ply = player->getCharacter();
```
JayFoxRox: `ply` is a shitty abbreviation for a character.
Other instances in this file too.

Beyond the scope?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2536,45 +2513,29 @@ bool opcode_00e3(const ScriptArguments& args, const ScriptPlayer player, const S
...
+	auto condition = ply->getCurrentVehicle() == nullptr;
+	return script::objectInRadius(args, ply, coord0, coord1, arg6, condition);
```
JayFoxRox: what's arg6?

JayFoxRox:
I still have to review quite a bit (that's a lot of code. Awesome :+1:  ). I mostly have nits which were probably moved over from the old code.
The only logical issue I came accross so far (iirc) is the one about the Globals / Locals not indicating that they are refs.

GitHub thinks all my comments are outdated, chances are this is fixed in a later commit I didn't see OR github just sucks.

I'd also add that it would be nice if changes on this PR were in seperate fixup commits (including the word "fixup" and the commit you intend to squash it with later) so we only have to review changes on the second pass.

danhedron:
Pushed new commits to deal with some of the issues. I'll squash these into the commits for the surrounding lines, or the previous commit that introduced that line.

rwengine/src/script/ScriptTypes.hpp
```diff
@@ -44,8 +44,8 @@ using ScriptBoolean = int32_t;
 using ScriptVec2 = glm::vec2;
 using ScriptVec3 = glm::vec3;
 
-using ScriptRGB = glm::u8vec3;
-using ScriptRGBA = glm::u8vec4;
+using ScriptRGB = glm::u32vec3;
+using ScriptRGBA = glm::u32vec4;
```
JayFoxRox: Shouldn't these be signed?

scripts/generate_scripts_functions.py
```diff
@@ -214,10 +214,10 @@ def arg_type(x, el):
     if 'AllowGlobalVar' in el.attrib:
         allowGlobal = True if el.attrib['AllowGlobalVar'] == 'true' else False
     if allowLocal == False and allowGlobal == True:
-        outRef = 'Global'
+        outRef = '&'
         outName = 'arg' + str(x) + 'G'
     elif allowLocal == True and allowGlobal == False:
-        outRef = 'Local'
+        outRef = '&'
```
JayFoxRox: Sorry to bother about this again, this is a visible ref now, but now we loose the info wether it's global or local?
Are we certain we don't need it?

(Am I overthinking this?)

danhedron: The variable name will contain the fact it's a local or a global for un-named variables. I'm not sure if we ever need to know if a value came from a local or a global though.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2512,9 +2512,9 @@ bool opcode_00e3(const ScriptArguments& args, const ScriptPlayer player, const S
 	@arg arg6 
 */
 bool opcode_00e4(const ScriptArguments& args, const ScriptPlayer player, const ScriptVec2 coord0, const ScriptVec2 coord1, const ScriptInt arg6) {
-	auto ply = player->getCharacter();
-	auto condition = ply->getCurrentVehicle() == nullptr;
-	return script::objectInRadius(args, ply, coord0, coord1, arg6, condition);
+	auto plyChar = player->getCharacter();
```
JayFoxRox: `playerChar` or even `playerCharacter` would have been better imo, but this'll do ;)
(I'm not a fan of abbreviations)

rwengine/src/script/ScriptTypes.hpp
danhedron: Yes

JayFoxRox:
Squash and rebase please for final review :)

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -0,0 +1,211 @@
...
+	z = p.z;
+}
+inline void setObjectPosition(GameObject* object, const ScriptVec3& coord)
+{
+	object->setPosition(coord + script::kSpawnOffset);
+}
+
+inline VehicleObject* getCharacterVehicle(CharacterObject* character)
+{
+	return character->getCurrentVehicle();
```
JayFoxRox: This crashed on me from opcode 00db (when exiting the taxi mission) because character was nullptr

danhedron: Had you done anything else?

I've tried cancelling the mission with a fare and before picking up a fare and not crashed.

JayFoxRox: I'm sure this happened before the game could pick a random ped. So it can probably be reproduced by teleporting to the airfield, then toggling the taxi mission.
This happened very often. Similarily the paramedic mission crashed too (when aborting it). I've documented the behaviour on the playable mission tracker as "Crashes in `opcode_0173` on exit because character is nullptr."

I'm not sure if we should fix this in the opcode handler or the common "getVehicle" function and each setter etc. (I'd assume opcode handler because functions should never expect a `nullptr` for `this` or objects they operate on).

It's probably beyond the scope of this PR anyway.

danhedron:
Squashed as many fixes as I could, git refuses to cleanly rebase the last one because so many adjacent lines are different.

Things to create issues about:
- [ ] Occasional crash with taxi sub-mission (repeatedly hitting space causes this for me)
- [ ] Hitting space starts the taxi sub-mission (it shouldn't)
- [ ] Check where null checks are needed to match original behaviour

JayFoxRox:
We should feature-freeze this.  The PR is big enough as is and it's easy to fix stuff in follow ups.

The crashes which are remaining are far less dramatic than those present in the current master script system. We are essentially blocking further work with this.

So unless there are structural problems with this we should merge it.

JayFoxRox:
I might have found an issue:

`opcode_0187` (and other instances too) will call `CreateObjectBlip` which returns a local variable of type `BlipData` by value (`data`). Then that gets passed to the `blip` parameter by address `blip = &data`. Wouldn't that mean that the memory for the `data` is released on scope exit? Then `blip` would point at invalid memory..

danhedron:
Yes, the pointer is dangling once the opcode handler returns, however the ScriptBlip only exists until the call to call_unpacked returns. The only time the pointer could be accessed is in the ScriptBlip destructor.

``createObjectBlip`` should be better. Then I think this is probably worth merging and handling any further issues as they come up.

darkf:
Skimmed the code. LGTM. Any minor issues persisting can be targeted in follow-up PRs. This patch is a long time coming, indeed.

JayFoxRox:
Typo: `Coordinages` -> `Coordinates`

danhedron:
There are probably going to be further issues with the code in here, but I'm too much of a bottleneck to fix them all individually. Please open a PR or issue for anything else.

JayFoxRox:
Please create seperate issues for the problems we've discussed in this issue. Thanks.

PR #209, opened 2016-08-16
Cleanup MinGW build

JayFoxRox:
This does some minor cleanup for MinGW to make the build process less error prone.

* I've replaced all uses of cmath as we can also use the glm constants.
* master wouldn't work when the system include path is read only as mman was installed into the system path which is a no-no anyway.
* rwengine itself depends on iconv (LoaderGXT) and mman (MADStream). While these are linked by rwgame they are also required for rwviewer (which uses rwengine, but doesn't link them = fail). This commit is a purely temporary measure (See notes below)

Ideally we'd also link to iconv from every platform, but @danhedron is working on removing iconv altogether, so it's really not that important.
I wasn't sure wether I can remove the link to mman from rwgame so I kept it for now (someone on IRC said they might try to get rid of mman / libmad anyway, too).

### DIFF
--- a/rwengine/CMakeLists.txt
+++ b/rwengine/CMakeLists.txt
@@ -140,6 +140,11 @@ add_library(rwengine
     ${RWENGINE_SOURCES}
 )
 
+if(MINGW)
+	target_link_libraries(rwengine
+	mman)
+endif()
+
 target_link_libraries(rwengine
 	rwlib
 	${MAD_LIBRARY}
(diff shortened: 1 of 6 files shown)


danhedron:
Looks good to me, still builds on Linux. +1

tsjost:
Still works fine on Mac as well

rwengine/CMakeLists.txt
```diff
@@ -140,7 +140,9 @@ add_library(rwengine
 )
 
 if(MINGW)
-	add_definitions(-D _USE_MATH_DEFINES)	
+	target_link_libraries(rwengine
+	iconv
```
JayFoxRox: This must be removed as #210 was merged.

PR #210, opened 2016-08-16
Overhaul GXT handling and menu text

danhedron:
This fixes:

- Incorrect text handling
  - Text is now stored as it is loaded from the GXT
  - Glyphs are looked up in the font map using the text properly
- Not localized menu text
  - Use the GXT strings for the main menu and the pause menu
  - Use the correct font for the menu
- Remove iconv 

- [ ] Remove iconv from #209 after rebase
- [x] Add named values for special chars used in code (Heart, Armour symbol, etc)

### DIFF
--- a/rwengine/src/data/GameTexts.hpp
+++ b/rwengine/src/data/GameTexts.hpp
@@ -1,6 +1,5 @@
-#pragma once
-#ifndef _GAMETEXTS_HPP_
-#define _GAMETEXTS_HPP_
+#ifndef RWENGINE_GAMETEXTS_HPP
+#define RWENGINE_GAMETEXTS_HPP
 #include <string>
 #include <unordered_map>
 
--- a/rwengine/src/data/GameTexts.hpp
+++ b/rwengine/src/data/GameTexts.hpp
@@ -28,6 +28,19 @@ namespace GameStringUtil
 GameString fromString(const std::string& str);
 }
 
+/**
+ * Since the encoding of symbols is arbitrary, these constants should be used in
+ * hard-coded strings containing symbols outside of the ASCII-subset supported by
+ * all fonts
+ */
+namespace GameSymbols
+{
+	static constexpr GameStringChar Money  = '$';
+	static constexpr GameStringChar Heart  = '{';
+	static constexpr GameStringChar Armour = '[';
+	static constexpr GameStringChar Star   = ']';
+}
+
 class GameTexts
 {
 	using StringTable = std::unordered_map<GameStringKey, GameString>;
--- a/rwengine/src/data/GameTexts.hpp
+++ b/rwengine/src/data/GameTexts.hpp
@@ -3,21 +3,47 @@
 #include <string>
 #include <unordered_map>
 
+/**
+ * Each GXT char is just a 16-bit index into the font map.
+ */
+using GameStringChar = uint16_t;
+/**
+ * The game stores strings as 16-bit indexes into the font
+ * texture, which is something simllar to ASCII.
+ */
+using GameString = std::basic_string<GameStringChar>;
+/**
+ * GXT keys are just 8 single byte chars.
+ * Keys are small so should be subject to SSO
+ */
+using GameStringKey = std::string;
+
+namespace GameStringUtil
+{
+/**
+ * @brief fromString Converts a string to a GameString
+ *
+ * Encoding of GameStrings depends on the font, only simple ASCII chars will map well
+ */
+GameString fromString(const std::string& str);
+}
+
 class GameTexts
 {
-	std::unordered_map<std::string, std::string> _textDB;
+	using StringTable = std::unordered_map<GameStringKey, GameString>;
+	StringTable m_strings;
 public:
 
-	void addText(const std::string& id, const std::string& text) {
-		_textDB.insert({ id, text });
+	void addText(const GameStringKey& id, GameString&& text) {
+		m_strings.emplace(id, text);
 	}
 
-	std::string text(const std::string& id) {
-		auto a = _textDB.find(id);
-		if( a != _textDB.end() ) {
+	GameString text(const GameStringKey& id) {
+		auto a = m_strings.find(id);
+		if( a != m_strings.end() ) {
 			return a->second;
 		}
-		return id;
+		return GameStringUtil::fromString("MISSING: " + id);
 	}
 
 };
--- a/rwgame/DrawUI.cpp
+++ b/rwgame/DrawUI.cpp
@@ -101,7 +101,7 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world, GameRenderer* re
 
 	{
 		std::stringstream ss;
-		ss << "@" << std::setw(3) << std::setfill('0')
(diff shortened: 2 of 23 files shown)


christoph-heiss:
Builds and runs fine on macOS, didn't notice any regressions (although I only did very limited testing).

So +1 from me.

rwengine/src/data/GameTexts.hpp
```diff
@@ -3,21 +3,42 @@
 #include <string>
 #include <unordered_map>
 
+/**
+ * The game stores strings as 16-bit indexes into the font
+ * texture, which is something simllar to ASCII.
+ */
+using GameString = std::basic_string<char16_t>;
+/**
+ * Keys are small so should be subject to SSO
```
JayFoxRox: Sorry, but what's "SSO"?

rwengine/src/data/GameTexts.hpp
```diff
@@ -3,21 +3,42 @@
 #include <string>
 #include <unordered_map>
 
+/**
+ * The game stores strings as 16-bit indexes into the font
+ * texture, which is something simllar to ASCII.
+ */
+using GameString = std::basic_string<char16_t>;
```
JayFoxRox: If this uses something similar to ascii, but not unicode, why exactly do we still want to use char16_t then?
Wouldn't it be wise to have something like a GameCharacter which is a plain uint16_t?

If we have `fromString` anyway to do the conversion anyway, why aren't we working in `std::string` OR GameCharacter then?
All our game strings are loaded from the GXT anyway and the display uses a fixed offset - we never have to assign a meaning (such as unicode) anywhere unless we want to inject our own strings (and for that we assume ascii and convert from out `std::string` for which we'll also assume ascii).
There is a different type of string in GTA which is in the script system - those 8 char strings which are also used for the GXT keys. Those are probably ASCII or, if they aren't case sensitive we can assume them to be 64 bit which have to match. We never have to assign a meaning to those.

rwengine/src/data/GameTexts.hpp
danhedron: Small string optimisation. Since keys are always <= 8 [most implementations](http://stackoverflow.com/questions/27631065/why-does-libcs-implementation-of-stdstring-take-up-3x-memory-as-libstdc/28003328#28003328) will always store the key string inside the string object. Not a big deal for performance but it means taking keys as rvalues references has no benefit.

rwgame/DrawUI.cpp
```diff
@@ -101,7 +101,7 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world, GameRenderer* re
 
 	{
 		std::stringstream ss;
-		ss << "@" << std::setw(3) << std::setfill('0')
+		ss << "{" << std::setw(3) << std::setfill('0')
```
JayFoxRox: (See my earlier commits on the commit after this one)

If we introduce a new `GameCharacter` we could have a lookup / enum for common symbols. That would make this code a lot more readable. We'd have to support some kind of `toString()` method OR we simply allow concatenation of `GameCharacter`s (a GameString could simply be a `std::vector<GameCharacter>` which behaves similar to the `std::string` class)

(Note that `GameCharacter` is actually really ambigious as a class name, I don't have any better ideas though)

rwengine/src/data/GameTexts.hpp
```diff
@@ -3,21 +3,42 @@
...
 class GameTexts
 {
-	std::unordered_map<std::string, std::string> _textDB;
+	using StringTable = std::unordered_map<GameStringKey, GameString>;
+	StringTable m_strings;
 public:
 
-	void addText(const std::string& id, const std::string& text) {
-		_textDB.insert({ id, text });
+	void addText(const GameStringKey& id, GameString&& text) {
```
danhedron: I'm not sure what you mean here. `id` is already just a set of single byte chars.

rwengine/src/data/GameTexts.hpp
danhedron: This should be `uint16_t` rather than `char16_t`.

We can't use std::string as they only store single byte chars. fromString should only be used for the odd place where we need to convert some ASCII alphanumerics to a `GameString`, such as in the formatting of numbers etc.

The 8 bit chars for keys are already handled by `GameStringKey`. I hadn't considered storing the keys as 64-bit integers, since all keys are 8 bytes we should be able to get away with that unless there is some junk after each key (I'd hope not). I can look into that to see if it makes things a bit simpler.

rwengine/src/data/GameTexts.hpp
JayFoxRox: oops.. I didn't realize there was "GameString**Key**". I was under assumption this used the same type as the actual GXT value.

rwengine/src/data/GameTexts.hpp
JayFoxRox: (See my other comment below, I didn't see we already used 2 different types)

What I meant with the second part was: For our custom strings, your intention was to use unicode strings (u"blah") which assigns a meaning to the symbols. We only display those strings with the game font map, so we can't assign meanings to game characters unless we make assumptions about the fontmap.
(You assigned that meaning in the u"" prefixed strings and moved that meaning to a `char16_t` (GameString = there shouldn't be a meaning here unless we know the fontmap))

Let's assume that the fontmap is ASCII for 'A'-'Z'. We could store our custom strings in `std::string` and do the conversion with `fromString` (only supporting 'A'-'Z'). Alternatively we could store a `GameString` in the source code directly, as it would appear in the GXT file (meaning as an array, maybe with a comment describing what it represents in a specific fontmap).

So, there is no point in prefixing all strings with u"". The fact that custom strings and game strings have the same width doesn't matter. Unless the host compiler couldn't express the subset we want to map (such as 'A'-'Z') in `char` / `std::string` OR we want to map unicode symbols for a japanese font map or something (from our custom strings).

rwengine/src/data/GameTexts.hpp
```diff
@@ -28,6 +28,19 @@ namespace GameStringUtil
...
+ * hard-coded strings containing symbols outside of the ASCII-subset supported by
+ * all fonts
+ */
+namespace GameSymbols
+{
+	static constexpr GameStringChar Money  = '[';
+	static constexpr GameStringChar Heart  = '{';
+	static constexpr GameStringChar Armour = '{';
+	static constexpr GameStringChar Star   = ']';
+}
```
JayFoxRox: nit: imo these could just be part of the GameString class in a Symbol namespace (Say GameString::Symbol::Heart) OR if we typedef to a GameStringChar they could be part of that class too (GameStringChar::Heart)?
Adding too many namespaces / classes for something this simple might be confusing.

danhedron: Is it possible to namespace identifiers into classes?

danhedron: No. I'm not sure how that could be achieved without inheriting from basic_string, which is bad, or proxying calls.

JayFoxRox:
Tested ingame, code would probably reduce to a couple of nits.
It's clearly an improvement over master.

LGTM.

PR #211, opened 2016-08-18
Clang format test

danhedron:
```
for DIR in $(find rw* -type d); do
	echo $DIR
	for FILE in $DIR/*.{cpp,hpp}; do
		clang-format-3.6 -i $FILE
		git add $FILE
	done
	git commit -m "clang-format files in $DIR"
done
```


danhedron:
Oops. Wrong button.

rwengine/src/engine/GameInputState.hpp
```diff
@@ -1,72 +1,65 @@
...
-	{
-		return currentLevels[c];
-	}
+  float operator[](Control c) const { return currentLevels[c]; }
 
-	bool pressed(Control c) const
-	{
-		return currentLevels[c] > kButtonOnThreshold;
-	}
+  bool pressed(Control c) const { return currentLevels[c] > kButtonOnThreshold; }
```
JayFoxRox: These should still be multi-line imo.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -10,504 +10,450 @@
...
+    if (targetDistance <= 0.4f) {
+      entering = true;
+      // Warp character to vehicle orientation
+      character->controller->setMoveDirection({0.f, 0.f, 0.f});
+      character->controller->setRunning(false);
+      character->setHeading(glm::degrees(glm::roll(vehicle->getRotation())));
+
+      // Determine if the door open animation should be skipped.
+      if (entryDoor == nullptr || (entryDoor->constraint != nullptr &&
+                                   glm::abs(entryDoor->constraint->getHingeAngle()) >= 0.6f)) {
```
JayFoxRox: This should have broken at the `||`. However, we can fix these issues manually.
Ideally we'd store more complicated logic in a boolean which also describes what's going on: `bool isDoorOpened = ...`

Hockenberry: Using clang-format means you never think again about manually fixing
formatting issues - and you never fix source code manually.

The source code format is described by the clang-format rules and
formatting is part of a save action. You need to find a configuration for
clang-format containing your "personal" style.

After using auto formatting (with clang-format for c++ and eclipse/java) in
professional environments I never discussed with my team over source code
formatting again.


2016-08-19 12:59 GMT+02:00 Jannik Vogel <notifications@github.com>:

danhedron:
- Reduced  column width to 90, some expressions are more readable when broken earlier
- Prevent single-line function definitions, vertical space isn't that scarce.

Issue #200, opened 2016-08-12
Add info about supported platforms in README.md


JayFoxRox:
Users who are unsure wether OpenRW works on their platform might instantly close the browser window again after seeing this repo. Especially with GTA being a mostly-windows game we should clarify that we are portable.

We also seem to support Windows, Linux, MacOS, FreeBSD and OpenBSD officially (through the CMake file). We should just name the big 3 + “various flavours of BSD” and claim that OpenRW is portable.
We should also add that we require C++11 and our deps (most importantly OpenGL 3.3 Core) or link the building page for that.

---

I've already extended the wiki FAQ page to list some supported platforms.

JayFoxRox:
Update: I've added C++11 and 3.3 as platform requirement in the FAQ for now

JayFoxRox:
This has been fixed by #215

PR #215, opened 2016-08-23
Update readme with the supported systems

danhedron:
Handles #200 

This just adds a list of operating systems that are supported so people know it's worth trying out. People who are interested will come across the minimum requirements while looking at the FAQ.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,9 @@
 
 [![Build Status](https://travis-ci.org/rwengine/openrw.svg?branch=master)](https://travis-ci.org/rwengine/openrw)
 
-OpenRW is an open source re-implementation of Rockstar Games' Grand Theft Auto III,
-a classic 3D action game first published in 2001.
+OpenRW is a cross-platform, open source re-implementation of Rockstar Games'
+Grand Theft Auto III, a classic 3D action game first published in 2001.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,7 +3,8 @@
 [![Build Status](https://travis-ci.org/rwengine/openrw.svg?branch=master)](https://travis-ci.org/rwengine/openrw)
 
 OpenRW is an open source re-implementation of Rockstar Games' Grand Theft Auto III,
-a classic 3D action game first published in 2001.
+a classic 3D action game first published in 2001. It is designed to run on Linux,
+macOS and Windows systems, along with several variants of BSD.
```
JayFoxRox: "OpenRW is a cross-platform, open-source re-implementation [..]" and then
"Our implementation has been ported to Linux, macOS, Windows and several variants of BSD." ?

- imo your proposal sounds like we don't want to support any other systems because it's only designed for these.
I also felt that yours sounded as if "Grand Theft Auto III [..] was designed to run on Linux, macOS, [..]"

danhedron: It reads much better that way. Thanks.

JayFoxRox:
LGTM.

PR #216, opened 2016-08-24
Improve SDL & GL initialisation failure handling

danhedron:
This adds some usability fixes:

- Check that we actually opened a window and get a gl context
- Don't show the window if we don't get an OpenGL context
- Show an error dialog for environments where stderr is not readily accessible

### DIFF
--- a/rwgame/GameWindow.cpp
+++ b/rwgame/GameWindow.cpp
@@ -10,7 +10,7 @@ GameWindow::GameWindow() :
 
 void GameWindow::create(const std::string& title, size_t w, size_t h, bool fullscreen)
 {
-	uint32_t style = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE;
+	Uint32 style = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_HIDDEN;
 	if (fullscreen)
 		style |= SDL_WINDOW_FULLSCREEN;
 
@@ -33,6 +33,8 @@ void GameWindow::create(const std::string& title, size_t w, size_t h, bool fulls
 		std::string sdlErrorStr = SDL_GetError();
 		throw std::runtime_error("SDL_GL_CreateContext failed: " + sdlErrorStr);
 	}
+
+	SDL_ShowWindow(window);
 }
 
 
--- a/rwgame/GameWindow.cpp
+++ b/rwgame/GameWindow.cpp
@@ -50,16 +50,12 @@ void GameWindow::close()
 void GameWindow::showCursor()
 {
 	SDL_SetRelativeMouseMode(SDL_FALSE);
-	SDL_ShowCursor(SDL_ENABLE);
-	SDL_SetWindowGrab(window, SDL_FALSE);
 }
 
 
 void GameWindow::hideCursor()
 {
 	SDL_SetRelativeMouseMode(SDL_TRUE);
-	SDL_ShowCursor(SDL_DISABLE);
-	SDL_SetWindowGrab(window, SDL_TRUE);
 }
 
 
--- a/rwgame/GameWindow.cpp
+++ b/rwgame/GameWindow.cpp
@@ -22,7 +22,17 @@ void GameWindow::create(const std::string& title, size_t w, size_t h, bool fulls
 	SDL_GL_SetAttribute(SDL_GL_STENCIL_SIZE, 8);
 
 	window = SDL_CreateWindow(title.c_str(), SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED, w, h, style);
+	if (window == nullptr) {
+		// Window creation failure is fatal
+		std::string sdlErrorStr = SDL_GetError();
+		throw std::runtime_error("SDL_CreateWindow failed: " + sdlErrorStr);
+	}
 	glcontext = SDL_GL_CreateContext(window);
+	if (glcontext == nullptr) {
+		// context creation failure is fatal
+		std::string sdlErrorStr = SDL_GetError();
+		throw std::runtime_error("SDL_GL_CreateContext failed: " + sdlErrorStr);
+	}
 }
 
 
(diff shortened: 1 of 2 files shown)


rwgame/main.cpp
```diff
@@ -1,9 +1,23 @@
...
+	try {
+		RWGame game(argc, argv);
 
-	RWGame game( argc, argv );
+		return game.run();
+	} catch (std::exception& ex) {
+		const char* kErrorTitle = "Fatal Error";
 
-	return game.run();
+		std::cerr << kErrorTitle << "\n" << ex.what() << std::endl;
```
JayFoxRox: I feel like this should go through our "logger" system.
It should probably have a "fatal" option which forwards it to a callback or something which would be the SDL error box and controlled exit?

(Might be beyond the scope of PR)

danhedron: I did think about that, and improving logging is something we should do for usability if we want to give more useful error messages in general. It's beyond this PR for sure.

rwgame/GameWindow.cpp
```diff
@@ -10,7 +10,7 @@ GameWindow::GameWindow() :
 
 void GameWindow::create(const std::string& title, size_t w, size_t h, bool fullscreen)
 {
-	uint32_t style = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE;
+	uint32_t style = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_HIDDEN;
```
JayFoxRox: nit: While you are at it, SDL2 calls this type `Uint32` which may or may not be the same as `uint32_t`.
(I prefer using intended data types exactly)

danhedron: Fair point. They're explicitly the same size but I'll change this to for consistency with what SDL2 uses.

PureTryOut:
Tested this out (as I was the cause of the original error anyway), works good now! It now gives me a nice message box with an error telling me exactly what is wrong, thanks.

danhedron:
@JayFoxRox noted on IRC that this would catch all types of exceptions which would make debugging useful exceptions like out_of_range difficult.

I changed the catch to only catch runtime_exception(s). These are used in OpenRW to throw "fatal" failures that mean continuing to execute would be pointless (if we don't have an OpenGL context, for example). Where they are throw is obvious so it's not a problem if they are all handled. In the future we can embed location into the thrown message if we think it would be useful for support.

danhedron:
Cursor is now freed if there is an error while the cursor is locked. This is quite unlikely to occur at the moment, but it's easy to avoid.

rwgame/GameWindow.cpp
```diff
@@ -34,12 +36,20 @@ void GameWindow::create(const std::string& title, size_t w, size_t h, bool fulls
 		throw std::runtime_error("SDL_GL_CreateContext failed: " + sdlErrorStr);
 	}
 
+	if (mainwindow_ == nullptr) {
+		mainwindow_ = window;
+	}
```
JayFoxRox: If you chose to do this, wouldn't it be easier to just keep track of the window which last grabbed the focus?

rwgame/main.cpp
```diff
@@ -9,6 +9,14 @@ int main(int argc, char* argv[])
 
 		return game.run();
 	} catch (std::runtime_error& ex) {
+		// Make sure the user can use the cursor to acknowlege the error if
+		// it occurs while the mouse is locked to the window.
+		SDL_SetRelativeMouseMode(SDL_FALSE);
+		SDL_Window* activeWindow = GameWindow::getMainWindow();
+		if (activeWindow != nullptr) {
+			SDL_SetWindowGrab(activeWindow, SDL_FALSE);
```
JayFoxRox: Can't we just somehow force the focus away from the main window?
If it's not the same focus group it might automaticly disable mousegrab when switching focus.

JayFoxRox:
This can be caused to crash when throwing a runtime error during a cutscene (/ audio playback)?

```
Fatal Error
thisIsAnErrorIForced

Thread 7 "rwgame" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffae288700 (LWP 16039)]
0x00007ffff63335ff in mad_header_decode () from /usr/lib/libmad.so.0
(gdb) bt
#0  0x00007ffff63335ff in mad_header_decode () from /usr/lib/libmad.so.0
#1  0x00007ffff6335a6e in ?? () from /usr/lib/libmad.so.0
#2  0x00007ffff6335fc3 in mad_decoder_run () from /usr/lib/libmad.so.0
#3  0x00007ffff5b36a9f in std::execute_native_thread_routine (__p=0x8260a90)
    at /build/gcc/src/gcc/libstdc++-v3/src/c++11/thread.cc:83
#4  0x00007ffff6974484 in start_thread () from /usr/lib/libpthread.so.0
#5  0x00007ffff52a86dd in clone () from /usr/lib/libc.so.6
```

---

I wouldn't worry about it too much.
This still catches the 2 most important cases:
* Config does not exist
* OpenGL 3.3 context init fail

LGTM
*(actually it doesn't - that message box is the ugliest thing I've seen in ages. Jokes aside: mergeable)*

rwgame/main.cpp
```diff
@@ -1,9 +1,28 @@
...
+	} catch (std::runtime_error& ex) {
+		// Catch runtime_error as these are fatal issues the user may want to
+		// know about like corrupted files or GL initialisation failure.
+		// Catching other types (out_of_range, bad_alloc) would just make
+		// debugging them more difficult.
 
-	return game.run();
+		const char* kErrorTitle = "Fatal Error";
+
+		std::cerr << kErrorTitle << "\n" << ex.what() << std::endl;
```
JayFoxRox: nit: this should use logging code like I said, it's totally non-standard compared to all other text we print. It works for now though

danhedron: I agree. This will get fixed in a follow up.

PR #220, opened 2016-08-28
Fixes #219: Take TOBJ data into account when building render lists

### LINKED ISSUE #219: Add support for TOBJ: Objects that exist depending on time of day
Currently, OpenRW makes no effort to hide objects that should only be visible at night, or during the day. At any hour in the game, every such object is visible.

So far, the data required to do this is already parsed by `LoaderIDE.cpp` into `ObjectData::timeOn` and `ObjectData::timeOff`, [here](https://github.com/rwengine/openrw/blob/bda0107bb1d720a0a3b91a8861f5830a3e9f9057/rwengine/src/loaders/LoaderIDE.cpp#L76). These values correspond to the hour of day at which the object should be visible and at which it should no longer be visible, respectively.

More research needs to be done as to whether these objects are purely graphical (e.g if a collideable object disappears at night, it must be replaced by the another model with the same collision mesh, or if one of the two has a collision mesh and the other doesn't), or if objects that are not visible should also have their collisions disabled.

Currently some (but not all) lit window models seem to be offset from their respective window by ~40cm. This is visible near the debug start area. I'm not sure whether this is a bug in openrw or if this is part of the actual game, so more research is required on this point.

There is also the question of how breakable TOBJ objects should be handled. I haven't found any in my GTA3 files, but http://www.gtamodding.com/wiki/TOBJ#Type_2 claims they exist, so further research is needed on this too.

Mischa-Alff:
This is an implementation of TOBJ support for openrw, fixing issue #219. 

So far, I don't handle breakable TOBJ objects, as I can't find any in my GTA3 files and as such I don't think support for that would be required to play GTA3. I can look into it further if needed, however.

It also seems that for static meshes that have day/night variants, only one of the two has collision information, so it seems to be purely graphical for those. If we ever have breakable TOBJ, we need to look into how the physics for those should be handled.

As I pointed out in #219, the windows near the debug spawn area all seem to be offset from their respective buildings. I have no idea what the cause of this is.

### DIFF
--- a/rwengine/src/render/ObjectRenderer.cpp
+++ b/rwengine/src/render/ObjectRenderer.cpp
@@ -198,17 +198,14 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
 
 
 	// Handles times provided by TOBJ data
-	if (instance->object->timeOn != 0 || instance->object->timeOff != 0)
+	const auto currentHour = m_world->getHour();
+	if (instance->object->timeOff < instance->object->timeOn)
 	{
-		const auto currentHour = m_world->getHour();
-		if (instance->object->timeOff < instance->object->timeOn)
-		{
-			if ( currentHour >= instance->object->timeOff && currentHour < instance->object->timeOn )
-				return;
-		} else {
-			if ( currentHour >= instance->object->timeOff || currentHour < instance->object->timeOn )
-				return;
-		}
+		if ( currentHour >= instance->object->timeOff && currentHour < instance->object->timeOn )
+			return;
+	} else {
+		if ( currentHour >= instance->object->timeOff || currentHour < instance->object->timeOn )
+			return;
 	}
 
 	auto matrixModel = instance->getTimeAdjustedTransform(m_renderAlpha);
--- a/rwengine/src/render/ObjectRenderer.cpp
+++ b/rwengine/src/render/ObjectRenderer.cpp
@@ -196,6 +196,21 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
 		return;
 	}
 
+
+	// Handles times provided by TOBJ data
+	if (instance->object->timeOn != 0 || instance->object->timeOff != 0)
+	{
+		const auto currentHour = m_world->getHour();
+		if (instance->object->timeOff < instance->object->timeOn)
+		{
+			if ( currentHour >= instance->object->timeOff && currentHour < instance->object->timeOn )
+				return;
+		} else {
+			if ( currentHour >= instance->object->timeOff || currentHour < instance->object->timeOn )
+				return;
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -196,6 +196,20 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
 		return;
 	}
 
+
```
JayFoxRox: nit: A comment stating that this handles times supplied by TOBJ would be nice imo

rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -196,6 +196,20 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
...
+	if (instance->object->timeOn != 0 || instance->object->timeOff != 0)
```
JayFoxRox: This shouldn't have to be checked (we set timeOn and timeOff to 0 and 24 (check again?) for normal OBJs so they are always shown).

Mischa-Alff: We set both timeOn and timeOff to zero for normal OBJs, so without this line, none of the normal OBJs are rendered.

rwengine/src/render/ObjectRenderer.cpp
Mischa-Alff: Added

rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -196,6 +196,18 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
...
+	// Handles times provided by TOBJ data
+	const auto currentHour = m_world->getHour();
+	if (instance->object->timeOff < instance->object->timeOn)
+	{
+		if ( currentHour >= instance->object->timeOff && currentHour < instance->object->timeOn )
+			return;
+	} else {
+		if ( currentHour >= instance->object->timeOff || currentHour < instance->object->timeOn )
+			return;
+	}
```
JayFoxRox: nit:  I guess I'd have save some bool result if the time is in the timeOn / timeOff range (`isInRange`) and then checked if `drawIfInRange` is also true. I prefer if variables are used because their name can explain what the intention is.

rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -196,6 +196,18 @@ void ObjectRenderer::renderInstance(InstanceObject *instance,
 		return;
 	}
```
JayFoxRox: nit: useless double-newline

JayFoxRox:
Code LGTM, but **currently untested**.

(I'll try this later. Remind me if I forget about this)

danhedron:
LGTM. tested and it seemed to work.

PR #221, opened 2016-08-30
Script: Implement opcode_0363 (Object visibility)

### LINKED ISSUE #51: Doors at Luigis place don't open
I'm scene where Luigi gives you Luigis girls mission characters clip through the door when they come to talk with you.

JayFoxRox:
Fixes #51. Only implements 0363. No known issues.

---

Style wise it's debatable wether `setVisible` should be part  of GameObject instead. It might deserve a comment about being entirely script controlled (and not something like TOBJ) but I wasn't sure what kind of doc format I should use.

I'd also prefer `args.getModel()` or an overload for `ScriptModel`, but it's beyond the scope of this PR.

---

Probably conflicts with #220 (I'll happily rebase after #220 is merged).

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -11,6 +11,8 @@
 #include <ai/PlayerController.hpp>
 #include <data/CutsceneData.hpp>
 
+#include <boost/algorithm/string/predicate.hpp>
+
 /**
 	@brief NOP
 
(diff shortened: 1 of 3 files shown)


rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -10045,15 +10047,36 @@ void opcode_0362(const ScriptArguments& args, const ScriptCharacter character, S
...
-void opcode_0363(const ScriptArguments& args, ScriptVec3 coord, const ScriptFloat radius, const ScriptModel model, const ScriptBoolean arg6) {
-	RW_UNIMPLEMENTED_OPCODE(0x0363);
-	RW_UNUSED(coord);
-	RW_UNUSED(radius);
-	RW_UNUSED(model);
-	RW_UNUSED(arg6);
-	RW_UNUSED(args);
+void opcode_0363(const ScriptArguments& args, ScriptVec3 coord, const ScriptFloat radius, const ScriptModel model, const ScriptBoolean visible) {
+	auto& models = args.getVM()->getFile()->getModels();
+	auto& modelName = models[-model];
```
danhedron: nit: no check for model > 0; not sure where this is allowed to happen so any annotation on ScriptModel usage is useful.

danhedron:
LGTM. Tested + Reviewed

rwengine/src/script/modules/GTA3ModuleImpl.inl
JayFoxRox: I agree.. however, instead of fixing it here we should just cross fingers imo and create a prioritized issue about proper `ScriptModel` / `args.getModel()` instead.

This also happens elsewhere in the code and another piece of code uses `modes[abs(index)]` (which felt even worse)

PR #222, opened 2016-08-30
Coloured radar-blips and radardisc z-order

### LINKED ISSUE #61: Mission icon on map is white square
In original it's purple dot or small square. At least for the first mission. 

http://i.imgur.com/PaGXcEw.png

JayFoxRox:
This fixes #61 by implementing the colored blips. Their size is also initialized correctly now.
Also blips which aren't supposed to be shown (controlled via script) are hidden too now (This does not hide the savegame icons etc. when on a mission because it's a seperate mechanism).

Not all script functions for the blips have been implemented yet. However, it's certainly better than nothing at all / white blips.

I've also changed the order the radar north and player marker are drawn so they don't draw over the blips.

---

I accidently went a little overboard and implemented triangular up/down arrows but later found out they were only introduced in vice city. So I'll keep that code to myself for now - it's not part of the PR.
I'll probably refactor the code a bit and submit a version where it is `#ifdef`'d out until we have VC support.

---

I used to have a hack which scaled the blip size to 150% of what it currently is. This was introduced as GTA uses a 16px radar_north at 640x480 (according to some screenshots I found). OpenRW uses 24 instead, which means we have a 50% increase in scale somehow..?
I then checked out [another video on youtube](https://www.youtube.com/watch?v=RoQStKto_n4) and the blip fit nicely between the vertical parts of the **N**orth symbol. So I removed the hack again.

More research has to be done on our hud scaling - but we should worry about more important things first.

### DIFF
--- a/rwengine/src/engine/GameState.hpp
+++ b/rwengine/src/engine/GameState.hpp
@@ -181,6 +181,7 @@ struct BlipData
 		Vehicle = 1,
 		Pickup = 2,
 		Character = 3,
+		Instance = 4,
 	};
 	BlipType type;
 	GameObjectID target;
--- a/rwengine/src/engine/GameState.hpp
+++ b/rwengine/src/engine/GameState.hpp
@@ -187,9 +187,12 @@ struct BlipData
 	// If target is null then use coord
 	glm::vec3 coord;
 	
-	/* Texture for use in the radar */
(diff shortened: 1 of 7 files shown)


danhedron:
LGTM. Tested & Reviewed

PR #227, opened 2016-09-01
Fix debug-camera

JayFoxRox:
This fixes 3 seperate issues:

- Undefined behaviour as `_invertedY` was not initialized and caused random camera inversion. This new implementation uses the user-configuration like `IngameState` does.
- Speed-modifier (`sonicSpeed`) was broken because the modifier flag was checked in key-up / down instead of checking for shift directly (which didn't work for unknown reasons?). Ideally we should check the key state instead. However, this is an okay workaround.
- The camera speed was adjusted as it was often too fast for accurate placement (it's still fast).

(I don't feel like looking into why the speed modifier didn't work. It might be a result from more than one modifier being active on XFCE. The original code should have simply AND'ed for the shift press instead or sticked to the working solution which is already used for all other keys in `DebugState` - as implemented here)


danhedron:
LGTM. reviewed + tested

Issue #177, opened 2016-07-30
Improve command line argument parsing and add usage


darkf:
`rwgame --help` should display all available parameters, and it should support both short and long options.

There are some nice libraries for this, such as:

- boost::program_options (http://www.boost.org/doc/libs/1_61_0/doc/html/program_options.html)
- http://tclap.sourceforge.net/
- http://optionparser.sourceforge.net/
- https://github.com/tanakh/cmdline
- http://ezoptionparser.sourceforge.net/

Some of these are a single header file. Can we include one of them in `external/` or as a submodule and use them?

JayFoxRox:
We depend on boost anyway and it's widely available.
I didn't review the other libraries though.

Boost gets my vote.

---

However, this is clearly a low-priority issue as the current system works.

danhedron:
This is done in #229. we can open another issue if we end up removing boost in the future.

PR #229, opened 2016-09-01
Parse command line arguments with Boost

svenstucki:
I've had a go at issue #177. It's a simple change, but there are still some points that are up for discussion.

For example, I don't fully understand each available option yet - maybe you could provide some accurate descriptions for the options with a TODO.

The most important point is how to exit when `--help` is used. In the short term an exception can be used, however it's not very clean.

In the longer term command line parsing should probably be decoupled from the SDL/engine initialization and maybe even merged with config file parsing (btw. Boost has support for that, but might not work together with the current config setup). This would require more rewriting though.

### DIFF
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -62,7 +62,7 @@ RWGame::RWGame(int argc, char* argv[])
 		("height,h",     po::value<size_t>(),      "Game resolution height in pixel")
 		("fullscreen,f",                           "Enable fullscreen mode")
 		("newgame,n",                              "Directly start a new game")
-		("test,t",                                 "To be used with -n, starts in a test location instead")
+		("test,t",                                 "Starts a new game in a test location")
 		("load,l",       po::value<std::string>(), "Load save file")
 		("benchmark,b",  po::value<std::string>(), "Run benchmark and store results in file")
 	;
@@ -162,20 +162,17 @@ RWGame::RWGame(int argc, char* argv[])
 	{
 		loading->setNextState(new BenchmarkState(this, benchFile));
 	}
+	else if( test)
+	{
+		loading->setNextState(new IngameState(this, true, "test"));
+	}
 	else if( newgame )
 	{
-		if( test )
-		{
-			loading->setNextState(new IngameState(this,true, "test"));
-		}
-		else
-		{
-			loading->setNextState(new IngameState(this,true));
-		}
+		loading->setNextState(new IngameState(this, true));
 	}
 	else if( ! startSave.empty() )
 	{
-		loading->setNextState(new IngameState(this,true, startSave));
+		loading->setNextState(new IngameState(this, true, startSave));
 	}
 	else
 	{
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -50,6 +50,7 @@ RWGame::RWGame(int argc, char* argv[])
 	bool fullscreen = false;
 	bool newgame = false;
 	bool test = false;
+	bool help = false;
 	std::string startSave;
 	std::string benchFile;
 
@@ -64,18 +65,25 @@ RWGame::RWGame(int argc, char* argv[])
 		("newgame,n",                              "Directly start a new game")
 		("test,t",                                 "Starts a new game in a test location")
 		("load,l",       po::value<std::string>(), "Load save file")
-		("benchmark,b",  po::value<std::string>(), "Run benchmark and store results in file")
+		("benchmark,b",  po::value<std::string>(), "Run benchmark from file")
 	;
 
 	po::variables_map vm;
-	po::store(po::parse_command_line(argc, argv, desc), vm);
-	po::notify(vm);
-
-	if( vm.count("help") )
+	try
 	{
-		// TODO: This is a hack
-		std::cout << desc << std::endl;
-		throw std::runtime_error("Terminate");
+		po::store(po::parse_command_line(argc, argv, desc), vm);
+		po::notify(vm);
+	}
+	catch (po::error& ex)
+	{
+		help = true;
+		std::cout << "Error parsing arguments: " << ex.what() << std::endl;
+	}
+
+	if( help || vm.count("help") )
+	{
+		std::cout << desc;
+		throw std::invalid_argument("");
 	}
 	if( vm.count("width") )
 	{
@@ -178,7 +186,7 @@ RWGame::RWGame(int argc, char* argv[])
 	{
 		loading->setNextState(new MenuState(this));
 	}
-	
+
 	StateManager::get().enter(loading);
 
 	log.info("Game", "Started");
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -24,6 +24,7 @@
 #include <objects/VehicleObject.hpp>
 
 #include <boost/algorithm/string/predicate.hpp>
+#include <boost/program_options.hpp>
 #include <functional>
 
 #include "GitSHA1.h"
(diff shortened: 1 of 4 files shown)


danhedron:
I already want to move some initialisation out from the engine init so that we can access the logger in the outermost exception handler. Moving the command line parsing can be done in another PR later that improves initialisation.

rwgame/RWGame.cpp
```diff
@@ -48,39 +49,60 @@ RWGame::RWGame(int argc, char* argv[])
...
+	// Define and parse command line options
+	namespace po = boost::program_options;
+	po::options_description desc("Available options");
+	desc.add_options()
+		("help",                                   "Show this help message")
+		("width,w",      po::value<size_t>(),      "Game resolution width in pixel")
+		("height,h",     po::value<size_t>(),      "Game resolution height in pixel")
+		("fullscreen,f",                           "Enable fullscreen mode")
+		("newgame",                                "Start a new game") // TODO
+		("test",                                   "Testmode") // TODO
```
danhedron: This interacts with `newgame`, it creates some vehicles & weapons instead of starting the game script.

rwgame/RWGame.cpp
```diff
@@ -48,39 +49,60 @@ RWGame::RWGame(int argc, char* argv[])
...
+	// Define and parse command line options
+	namespace po = boost::program_options;
+	po::options_description desc("Available options");
+	desc.add_options()
+		("help",                                   "Show this help message")
+		("width,w",      po::value<size_t>(),      "Game resolution width in pixel")
+		("height,h",     po::value<size_t>(),      "Game resolution height in pixel")
+		("fullscreen,f",                           "Enable fullscreen mode")
+		("newgame,n",                              "Directly start a new game")
+		("test,t",                                 "To be used with -n, starts in a test location instead")
```
JayFoxRox: This works without `-n` iirc.

You'd just use `-t`, `-l` or `-n` to skip the menu (each represents a menu option).
(I've never used `-b` but it's probably also in that group)

svenstucki: No, I've checked that. With `-t` only it shows the main menu. Otherwise I would've figured out that this corresponds to the Test menu entry :).

svenstucki: Actually, I meant "debug" menu entry.

Maybe this can be considered a bug/enhancement though.

JayFoxRox:
You should add the new boost dep here so we can get proper travis output https://github.com/rwengine/openrw/blob/master/.travis.yml#L13 :smile: 

Note that I'm not even sure about using boost anymore, I was under the impression this part was a header-only lib. \- I'm currently working on the MSVC port of OpenRW and boost is a giant pita (250MB) dependency. Even stripped to only headers it's like 30MB. We are using it only for filesystem and the unit tests right now.
As the unit tests are optional and filesystem is not actually necessary for MSVC (as it also implements C++17 filesystem already) we could ditch boost as a dependency on Windows asap, and in a couple of months when C++17 has gained some ground. We'd only require a replacement for this and `boost::iequals`.

It's just food for thought though. You are doing a good job here and we can probably use this for a while. *IF* we decide to ditch boost at a later point we can probably still switch to another header-only lib way easier than with our current code :)

(I also recommend joining us on irc #openrw / freenode)

svenstucki:
As discussed on IRC - the latest commit should be a good compromise between not feeling too dirty and not overdoing it in a single PR.

The constructor no (should) no longer throw Boost exceptions, but uses a `std::invalid_argument` exception to exit whenever an invalid argument or `--help` is present.

rwgame/RWGame.cpp
```diff
@@ -67,14 +68,20 @@ RWGame::RWGame(int argc, char* argv[])
...
+		help = true;
+	}
 
-	if( vm.count("help") )
+	if( help || vm.count("help") )
 	{
-		// TODO: This is a hack
 		std::cout << desc << std::endl;
-		throw std::runtime_error("Terminate");
+		throw std::invalid_argument("");
```
JayFoxRox: nit: shouldn't this only happen in the `help` case?
I didn't test yet, but it looks like `./rwgame --help` would cause invalid argument? Is that standard? Here is a little `cat` (meow) experiment:

```
$ cat  --error; echo $?
cat: unrecognized option '--error'
Try 'cat --help' for more information.
```
Returns 1
```
$ cat  --help; echo $?
Usage: cat [OPTION]... [FILE]...
Concatenate FILE(s) to standard output.
...
Full documentation at: <http://www.gnu.org/software/coreutils/cat>
or available locally via: info '(coreutils) cat invocation'
```
Returns 0

rwgame/RWGame.cpp
```diff
@@ -48,39 +49,67 @@ RWGame::RWGame(int argc, char* argv[])
...
+	po::options_description desc("Available options");
+	desc.add_options()
+		("help",                                   "Show this help message")
+		("width,w",      po::value<size_t>(),      "Game resolution width in pixel")
+		("height,h",     po::value<size_t>(),      "Game resolution height in pixel")
+		("fullscreen,f",                           "Enable fullscreen mode")
+		("newgame,n",                              "Directly start a new game")
+		("test,t",                                 "Starts a new game in a test location")
+		("load,l",       po::value<std::string>(), "Load save file")
+		("benchmark,b",  po::value<std::string>(), "Run benchmark and store results in file")
```
danhedron: The benchmark only prints the result at the moment

danhedron:
All options seem to work. Looks close to ready imo.

rwgame/RWGame.cpp
svenstucki: Yeah, as I tried to explain, the exception is just a crutch to exit the game from the RWGame constructor.

I've added the unrecognized option text. It will only show the first "wrong" option, but so does e.g. cat. For example:
```
$ cat -a -b
cat: invalid option -- 'a'
```

The only thing that's left is the return code. It will indeed be the same for both a wrong command line argument and `--help`. This is a bit unfortunate, but would be easier to fix after reorganizing the code a bit.

rwgame/RWGame.cpp
svenstucki: Ah, no idea why I thought it's storing results in a file. Fixed that.

svenstucki:
I've rebased to current master and added some small improvements regarding your comments above.

Btw. I don't think Boost has a way to make `--newgame`, `--test`, etc. exclusive (i.e. making having more than one of those arguments an error). The way around this would be to change the options a bit, e.g. use `--run <newgame|test|load|benchmark> [<file>]` (`--run` would default to newgame) instead. It's a bit harder to parse, but certainly doable.

danhedron:
I'm keen to get this merged so that #235 can be re-based. There's an issue when passing invalid or ambiguous arguments:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::ambiguous_option> >'
  what():  option '--h' is ambiguous and matches '--height', and '--help'
fish: “./rwgame/rwgame --h” terminated by signal SIGABRT (Abort)
```
or
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::invalid_option_value> >'
  what():  the argument ('text') for option '--width' is invalid
fish: “./rwgame/rwgame --width text” terminated by signal SIGABRT (Abort)
```
I think just catching ``boost::program_options::error`` should handle all these other errors descriptively enough. You may want to squash or fix-up some of the smaller commits, but I'll leave that up to you.

Any further tidying up can be done when the initialisation is reorganised.

svenstucki:
You're right, I completely forgot to handle other exceptions. Fixed with your recommendation.

I've reduced the number of commits a bit.

danhedron:
Great, thanks for doing this. It looks ready for merging to me, do you mind updating the title?

Tested & reviewed.

svenstucki:
Nice! Thanks for accepting it.

PR #230, opened 2016-09-01
Make sure SDL_Quit() is called on exit

svenstucki:
On my Fedora using the fullscreen mode permamently changes the display resolution. Obviously this is extremely inconvenient. The fix is making sure `SDL_Quit` is called before exit. This function is safe to call - even if SDL init fails.

### DIFF
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -189,6 +189,8 @@ RWGame::~RWGame()
 	log.info("Game", "Cleaning up work queue");
 	delete work;
 
+	SDL_Quit();
+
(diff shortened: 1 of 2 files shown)


danhedron:
Thanks for this cleanup. We should probably be using SDL_WINDOW_FULLSCREEN_DESKTOP too.

danhedron:
LGTM. Tested and Reviewed.

We should switch to using `SDL_WINDOW_FULLSCREEN_DESKTOP` instead of just `SDL_WINDOW_FULLSCREEN` as it will never leave the user with the wrong video mode.

PR #232, opened 2016-09-02
Cheat handler + Money display

JayFoxRox:
This implements a cheat handler.
Only a couple of cheats are actually implemented.
There will have to be some decision how we will differentiate different official R* versions of GTA3 as cheats were changed in different regions and updates. I've added some defines near the cheat handler as a temporary workaround.

This also fixes some issue with the game state which broke the "IFIWEREARICHMAN" cheat (money used to be `uint16_t` in OpenRW, and so many other values which are listed as dword on gtamodding).

As I was messing with the money values anyway I decided to improve the money display.
It's now zero-padded to 8 digits. The value also increases slowly now.
The behaviour for the [money counting was reverse engineered (from a video) and documented here](https://docs.google.com/spreadsheets/d/1VwUk-HeDuJ_yC7p9vIhhoCzII7TPcZOvNGGOj1TjfIs/edit#gid=0). As you can see, OpenRW counts slightly faster than the original. I guess this is related to frameskips and being bound to framerate on the actual game. As OpenRW is running with dynamic timestep I introduced a timer instead and retrieve the 30Hz from that.

---

This causes crashes if cheats are typed before going ingame. Note that the actual game accepts cheats during the pause menu so we can't easily check in IngameState. I'm not sure but the original game might even crash when typing cheats on the main menu? There is probably an easy fix for this, but I don't know how to check wether we are ingame. The obvious solution for now is to not type cheats in the menus. I consider solving this crash beyond the scope.
(I'm pretty sure the original game also crashed if player related cheats were entered but there wasn't a player character; so I didn't add a check for that either)

### DIFF
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -24,6 +24,7 @@
 #include <objects/VehicleObject.hpp>
 
 #include <boost/algorithm/string/predicate.hpp>
+#include <functional>
 
 #include "GitSHA1.h"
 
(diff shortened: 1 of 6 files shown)


rwgame/RWGame.cpp
```diff
@@ -797,4 +968,10 @@ void RWGame::globalKeyEvent(const SDL_Event& event)
 		break;
 	default: break;
 	}
+
+	std::string keyName = SDL_GetKeyName(event.key.keysym.sym);
```
danhedron: Did you see https://wiki.libsdl.org/SDL_TextInputEvent? The cheats all use latin chars so it's fine as is.

danhedron:
Looks good to me, just 1 minor issue.

The debug menu is currently <kbd>M</kbd>, which conflicts with some cheats, should we move it to some symbol/function key?

We should also create an issue about opcode 0109 not being implemented, since missions currently don't give you any money (this is probably outside of the scope of this PR, but a small fix).

JayFoxRox:
We'll just let 0109 sit until people notice and fix it for us - it's beyond the scope.

Moving the <kbd>M</kbd> key is not necessary. I remember that using the "asperine" cheat in Vice City usually caused my player to leave the vehicle so you had to hold the arrow keys while typing that cheat.
So conflicting with input (even debug) is absolutely normal and fine for cheats.

---

I forgot to mention it earlier, but note that the cheat input window is intentionally not cleared: typing "gunsgunsgunsguns" activates "gunsgunsguns" twice.

rwgame/RWGame.cpp
JayFoxRox: I'm aware of it and I initially planned to use it. However, I read many people were having problems with it.
https://wiki.libsdl.org/SDL_StartTextInput < Note that this also activates an onscreen keyboard.

I'm aware that my way of collecting input is pretty hacky though, but I feel it's an okay solution which is hopefully portable.
Cheats are not too important anyway, so even if this breaks it's not serious. We'll worry about it once people report issues.

danhedron:
Alright. In that case, looks good to me.

JayFoxRox:
FTR: [This was recently updated in GTA:LC](https://github.com/spaceeinstein/gtalc-dinput8/commit/3ff02b83f25a98cc73e1db02576ef78fc018b579#diff-a6b6df20cbdd3661acbf0676fb2f6381R233). We might be able to learn something from their code.

Issue #154, opened 2016-06-23
Files can be loaded from any location


danhedron:
The current implementation of FileIndex allows any file to be loaded from any location.
This happens to work for GTA III but it is not correct.

Verify where different files are loaded from:
TXD, DFF, ...
- [ ] Loaded from archives

IFP
- [ ] Hardcoded / cuts.img

main.scm
- [ ] Hardcoded

sfx.*
- [ ] Hardcoded

wav/mp3
- [ ] Hardcoded directory path

ifp/ide
- [ ] Listed in gta3.dat

JayFoxRox:
Heads up: https://github.com/danhedron/openrw/pull/11 - @danhedron is working on it

danhedron:
Some of these have been fixed or improved in #234, more work is needed to fix the rest.

Issue #176, opened 2016-07-28
Files are opened in text-mode


JayFoxRox:
We should open all files in binary mode, not text-mode (Exceptions to this could be config files, but even for those binary is probably more logical on most platforms).

Apparently we didn't catch all fstream / fopen in https://github.com/rwengine/openrw/commit/e82be612a1a96818f6ed46398a89082f301add45 .

There is at least one remaining issue in LoaderCOL.

This mostly affects platforms where a newline is '\r\n' such as Windows.
Errors like this will be less likely if at least one Windows user reviews new PRs.

darkf:
There are quite a few `std::fstream` constructors without specifying the `std::ios_base::binary` flag (nevermind that some of them should probably be `std::istream`s). I tried to fix some of them, but things broke in very odd ways. Might try again sometime.

darkf:
This should be fixed as of 6d6d7726538f6a80640da3b901c7aca06411b073.

The few non-binary ones left need to be opened in text mode for `std::getline` to function properly.

JayFoxRox:
Shall we open a new issue for the std::getline thing?

I believe, ideally we should get independent of the text mode. Then replicate behavior of text mode on all platforms using binary mode so we are guaranteed to be portable?

PR #234, opened 2016-09-02
Improved file handling for un-archived files

danhedron:
This fixes the following:
- style fixes & better indentation
- Improve some naming
- Implement case-insensitive lookup of files in the game directory
- Load files from explicit paths instead of finding them using just their filename
- Remove some dead code
- Load IDE files at the correct time

### DIFF
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -93,14 +93,16 @@ void GameData::parseDAT(const std::string& path)
 			#ifndef RW_WINDOWS
 				line.erase(line.size()-1);
 			#endif
-			
+
 			size_t space = line.find_first_of(' ');
 			if(space != line.npos)
 			{
 				cmd = line.substr(0, space);
 				if(cmd == "IDE")
 				{
-					addIDE(line.substr(space+1));
+					auto path = line.substr(space+1);
+					auto systempath = index.findFilePath(path).native();
+					loadIDE(systempath);
 				}
 				else if(cmd == "SPLASH")
 				{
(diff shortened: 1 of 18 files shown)


JayFoxRox:
This would fix #154 .

I think #176 also fits well with this one.

danhedron:
- remove the FileFuture from this, since it's not needed at all
- rebase to master

rwengine/src/engine/GameData.cpp
```diff
@@ -49,10 +38,15 @@ GameData::~GameData()
...
 {
+	index.indexGameDirectory(datpath);
 	index.indexTree(datpath);
-	
-	parseDAT(datpath+"/data/default.dat");
-	parseDAT(datpath+"/data/gta3.dat");
+
+	// Initalize all the archives.
+	loadIMG("models/gta3.img");
+	loadIMG("anim/cuts.img");
```
danhedron: Needs a note that this shouldn't be considered for loading files in the same way as gta3.img

rwengine/src/engine/GameData.cpp
```diff
@@ -62,102 +56,81 @@ void GameData::load()
...
-						if(texpath[t] == '\\') {
-							texpath[t] = '/';
-						}
-					}
-					std::string texname = texpath.substr(texpath.find_last_of("/")+1);
-					loadTXD(texname);
-				}
+				auto path = line.substr(space+1);
+				auto systempath = index.findFilePath(path).native();
+				loadIDE(systempath);
```
danhedron: This resolve-before-call style is at different from the way the other loadX functions are implemented. Should be consistent.

danhedron:
I've reviewed this a few times myself. There may be one or two things worth improving but there are more changes coming soon and this is holding up another PR. Merging it now and anything else can be fixed in another PR.

PR #236, opened 2016-09-08
Fixed .native() and some cast errors

PerikiyoXD:
Signed-off-by: PerikiyoXD <perikiyoxd@gmail.com>

### DIFF
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -68,7 +68,7 @@ void GameData::load()
 void GameData::loadLevelFile(const std::string& path)
 {
 	auto datpath = index.findFilePath(path);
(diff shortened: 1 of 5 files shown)


rwengine/src/engine/GameData.cpp
```diff
@@ -290,7 +290,7 @@ void GameData::loadGXT(const std::string &name)
 void GameData::loadWaterpro(const std::string& path)
 {
 	auto syspath = index.findFilePath(path);
-	std::ifstream ifstr(syspath.c_str(), std::ios_base::binary);
+	std::ifstream ifstr((const char*)syspath.c_str(), std::ios_base::binary);
```
danhedron: This should just use .string() instead of casting, c_str() will be returning wide characters here. The same applies elsewhere the code calls .c_str() on paths.

PR #241, opened 2016-09-22
Correct object data storage

danhedron:
This goes part-way to fixing how openrw stores model data.

Objects now keep track of the model they are assigned

Models and their frames are loaded into the ModelInfo for each object. This should ease the next step of removing Model altogether.


rwengine/src/loaders/LoaderIDE.cpp
```diff
@@ -48,34 +48,35 @@ bool LoaderIDE::load(const std::string &filename) {
...
-                    getline(strstream, textureName, ',');
-                    getline(strstream, numClumps, ',');
-
-                    objs->numClumps = atoi(numClumps.c_str());
-                    for (size_t i = 0; i < objs->numClumps; i++) {
-                        std::string drawDistance;
-                        getline(strstream, drawDistance, ',');
-                        objs->drawDistance[i] = atoi(drawDistance.c_str());
+                    getline(strstream, objs->name, ',');
+                    getline(strstream, objs->textureslot, ',');
```
danhedron: This could all be simpler still by just using sscanf.

rwengine/src/objects/GameObject.hpp
```diff
@@ -53,10 +55,11 @@ class GameObject {
     bool visible;
 
     GameObject(GameWorld* engine, const glm::vec3& pos, const glm::quat& rot,
-               ModelRef model)
+               BaseModelInfo* modelinfo, ModelRef model)
         : _lastPosition(pos)
         , _lastRotation(rot)
         , objectID(0)
+        , modelinfo_(modelinfo)
```
danhedron: This should increase the refcount for modelinfo.

rwengine/src/data/ModelData.hpp
```diff
@@ -134,32 +238,124 @@ struct VehicleData : public ObjectInformation {
...
+    float wheelscale_;
+    int numdoors_;
+    std::string handling_;
+    VehicleClass vehicleclass_;
+    int frequency_;
+    int level_;
+    int componentrules_;
+    std::string vehiclename_;
+    size_t primaryColourIds_[8];
+    size_t secondayColourIds_[8];
```
danhedron: The colour sets should be stored here instead of in GameData

rwengine/src/objects/InstanceObject.hpp
```diff
@@ -22,15 +22,11 @@ class InstanceObject : public GameObject {
...
-                   InstanceObject* lod, std::shared_ptr<DynamicObjectData> dyn);
+                   const glm::quat& rot, const glm::vec3& scale,
+                   BaseModelInfo* modelinfo, InstanceObject* lod,
+                   std::shared_ptr<DynamicObjectData> dyn);
     ~InstanceObject();
 
-    SimpleModelInfo* getModel() const {
-        return getModelInfo<SimpleModelInfo>();
-    }
-
```
danhedron: Don't need to add this just to remove it later.

rwengine/src/data/ModelData.hpp
danhedron: Actually this can be done later, no need to store this here for now.

PR #244, opened 2016-10-20
General Game cleanup

danhedron:
You know what sucks? raw pointers. So lets' get rid of them.

- Split RWGame code that has nothing to do with the game into GameBase
- Use logger to print runtime exceptions so we can remove std::cerr usage
- Avoid dynamically allocating some of RWGame's members that don't need to be
- Switch to C++14 so we can use make_unique
- Use unique_ptr where appropriate to avoid manual memory management
- Kill SCMOpcodes since it wasn't needed anymore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 sudo: required
 dist: trusty
-language: cpp
+language: generic
+env:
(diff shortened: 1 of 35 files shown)


darkf:
Generally LGTM

PR #247, opened 2016-11-07
Fixup BT_BULLET_VERSION check

JayFoxRox:
Bullet 2.83.5 was 283, 2.83.6 was 284, 2.84 (API changes affecting OpenRW) was 285, 2.85.1 is still 285.
We want to check for 2.84 and upwards, so our check must be for 285.

Also see https://github.com/bulletphysics/bullet3/issues/854#issuecomment-258944820 (apparently 2.84 was a mistake, when really it looks like 2.83.6 was the mistake?!)

I'd like to have this tested by IRC user vincentflyson (who reported issues with the original check).


vflyson:
I can confirm this has solved the issue on Debian 9 x86_64 (bullet 2.83.7, gcc 6.2.0).
The game engine loads the debug level and it is playable with no crashes.

Thank you for fixing it.

danhedron:
Looks good to me

PR #249, opened 2016-11-17
update OpenAL headers (fixes compilation on OSX)

JohnPaulHarold:
~~This PR wraps some OpenAL headers in an if else block for the OSX system. On OSX, the headers are included in the system, but require a slightly different path.~~

Instead, adjust the path to the header to take advantage of a cmake script that does this if else routine for you. Thanks to @pcercuei via IRC for this insight.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ documentation/
 *~
 *swp
 *.user*
-
+.DS_Store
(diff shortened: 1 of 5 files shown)


JayFoxRox:
also, did you check what FindOpenAL suggests? the cmake file suggests using "" instead of <>

source: https://github.com/Kitware/CMake/blob/master/Modules/FindOpenAL.cmake#L21

JohnPaulHarold:
will try this again

Issue #128, opened 2016-06-04
Debug camera not used if cutscene / scripted camera active


JayFoxRox:
During cutscenes the debug camera is not activated. It also won't work if the camera is controlled by the script.
That is a problem because it would be useful to debug those scenes too.
We currently can't look at places where characters and vehicles will be spawned for example.

(It would also be useful if we could unpause the game while the debug menu is opened / debug camera is active. But that's another issue)

danhedron:
Fixed by #252

PR #252, opened 2016-12-02
Improve camera and look control

danhedron:
This lays the ground work for more complicated player controls, strafing etc. and moves responsibility for interpolating camera transformation into State.


danhedron:
Now includes a change to fix #128

PR #260, opened 2016-12-22
Updated .gitignore

PerikiyoXD:
Now ignoring:
· Cmake Files
· GitSHA1.cpp (this file is autogenerated by CMake)
· Build files (rwgame.exe)
· Visual Studio project files
· Makefile && install_manifest.txt


.gitignore
```diff
@@ -1,7 +1,43 @@
+# Ignore certain files #
+
+# CMake - General Files #
+
+cmake/
+cmakefiles/
+*.cmake
+CMakeCache.txt
```
JayFoxRox: These should only be part of the build folder. Naming a folder for cmake modules or scripts is a very typical setup so I don't recommend ignoring that folder either

JayFoxRox:
Aside from the auto generated files this is not necessary (should not be done).
CMake supports seperate folders for source and building. Using the same folder should not be done.

If we went that road, we might also create custom .gitignore lines for ignoring CodeWarrior projects, CodeBlocks project files and whatever generators people might use with CMake which could possibly clutter up the source directory.

The whole purpose of ignoring the `build/` folder in the existing .gitignore is to make building within the git repo possible within a build folder. That way editor / temporary files etc will already be ignored by git.
So instead of adding these explicit gitignore lines it would be much more sane and maintainable to extend the build/ ignore to contain other combinations of build folders such as "[B,b]uild/".
However, even that seems kind of arbitrary as you can have as many subfolders within "build" as you want (meaning you don't have to ignore "build-linux" if you can just use the already ignored "build/linux").

PerikiyoXD:
+1, just noticed we had this conversation before. Gonna revert some of the stuff. On some cases, its needed (ie: Visual Studio Code uses .vscode on the root folder, it's inevitable to have that folder away, kinda like .git folder) but I do agree about the `build/` folder

.gitignore
```diff
@@ -1,7 +1,33 @@
...
+
+*.DS_Store
+
+# Visual Studio Code files #
+
+.vscode/
+
+# Build #
+
+*.exe
```
JayFoxRox: This should end up in the build folder probably? = no need to ignore it here

paulsapps:
Probably doesn't matter but you can always use name.g.cpp style names for generated files, that way you can just ignore *.g.cpp

PerikiyoXD:
Dont tell me, atm i'm just the guy who modifies the gitignore :)

danhedron:
The ``*swp`` entry should be kept as those files are created by vim while editing files.

danhedron:
@PerikiyoXD add an entry for ``*.swp`` and this PR can be merged.

danhedron:
Closing due to inactivity.

PR #261, opened 2016-12-28
cmake libMAD fix + use boost for INI configuration read & generation

### LINKED ISSUE #172: Provide openrw.ini
[Generating/Bundling the openrw.ini should be automated](https://github.com/rwengine/openrw/wiki/Configuration).

Auto-generating it (if it's not found) has been suggested.

Another alternative to that would be to assume a default config and assume the game is in the current directory (for a gta3.exe drop-in replacement style binary).
Packagers could then attach a premade openrw.ini (included in this repo / extra repo) to their packages on Linux and other platforms if there is need for it.

We could also have a CLI to generate the config (probably in a platform dependent location) like X11 does.

I believe the original idea was to create an installer for platforms like Windows, though, I kind of dislike this idea.

madebr:
This PR fixes and adds the following

- [x] CMake now fails when libMAD is not found
- [x] use boost::property_tree for _openrw.ini_ read and default generation
       * Remove inih
       * When _openrw.ini_ fails to parse, a default one is printed to stdout.
       * This partially fixes #172. No default configuration file is written to disk (yet).
- [ ] bogus data in _openrw.ini_ are ignored.

This is my first PR for openrw. Please tell me if something is off.


danhedron:
Since we already depend on boost, I'm OK with removing inih.

My issues:

- The default configuration should be printed regardless of it being a debug build.
- There should be some instruction on what to do with the default configuration, perhaps a note on where the file needs to be, and what should be changed.
- If the configuration contains end of line comments, the parser includes the comment as part of the value. So the behaviour is different.
- Does this break the build for Windows, or make it more difficult to build OpenRW on Windows?

Other than that, the changes look reasonable.

haphzd:
Does not break anything on Windows.

rwgame/GameConfig.cpp
```diff
@@ -62,25 +60,80 @@ std::string GameConfig::getDefaultConfigPath() {
...
+
+    read_config("input.invert_y", this->m_inputInvertY, false, boolt);
+
+    if (!success) {
+        std::stringstream strstream;
+        pt::write_ini(strstream, defaultConfig);
+
+        RW_MESSAGE("Failed to parse configuration file (" + path + ").");
+        RW_MESSAGE("Create one looking like this at \"" + path + "\":");
+        RW_MESSAGE(strstream.str());
```
danhedron: ``RW_MESSAGE`` is used for debug messages, compiled in Release these messages won't go anywhere (I've updated the wiki to mention this).

I think handling failure would make more sense in the GameBase constructor where it calls ``config.isValid()`` as there you can use ``log.error()`` to output a user-facing message before exiting.

rwgame/GameConfig.cpp
```diff
@@ -88,50 +88,139 @@ struct BoolTranslator {
...
+    ParseType destType, std::string &destination)
+{
     bool success = true;
+    pt::ptree srcTree;
+
+    if ((srcType == ParseType::STRING) || (srcType == ParseType::FILE)) {
+        std::istream *istream = nullptr;
+        switch (srcType) {
+            case ParseType::STRING:
+                istream = new std::istringstream(source);
```
danhedron: For future OpenRW code it's preferable to avoid dynamic allocation, especially without using unique or shared ptr to avoid leaking memory if an exception gets thrown. This isn't a likely occurrence here but it's an approach openrw should adopt throughout.

Is it possible to do this without relying on a raw pointer? Perhaps refactoring the `if (istream != nullptr)` block into a function-local  lambda or a static member and passing a reference to an istream.

tests/test_Logger.cpp
```diff
@@ -2,30 +2,30 @@
 #include <core/Logger.hpp>
 #include <test_globals.hpp>
 
-class CallbackReceiver : public Logger::MessageReciever {
+class CallbackReceiver : public Logger::MessageReceiver {
```
danhedron: What has changed in this file?

tests/test_config.cpp
```diff
@@ -4,17 +4,25 @@
...
     GameConfig config("openrw_test.ini", "/tmp");
 
     BOOST_CHECK(config.isValid());
 
     BOOST_CHECK_EQUAL(config.getGameDataPath(), "/dev/test");
+    BOOST_CHECK_EQUAL(config.getGameLanguage(), "american");
+    BOOST_CHECK_EQUAL(config.getInputInvertY(), false);
 }
 
+
```
danhedron: Since this PR adds support for writing configuration to a file/a string, could you add a test to make sure that keeps working until it's used in-game?

tests/test_Logger.cpp
madebr: Spelling check

tests/test_config.cpp
madebr: Will do in this PR.

tests/test_Logger.cpp
danhedron: I see it now. It's more obvious in the inline diff.

danhedron:
Is there any progress on these changes?

madebr:
Sorry. I dozed off.

Looking at the output of Travis, I don't see the maybe-unitialized warning on my machine.
(I'm running g++ (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1))

I also added boost-filesystem to the requirements of the test suite (in tests/CMakeLists.txt).
That should give no problems since rwlib already has a hard dependency on that library.

madebr:
My temp directory got littered with openrw temporary files.
After this commit, they are deleted after each test case.

tests/test_config.cpp
```diff
@@ -1,20 +1,193 @@
...
+    for (auto &section : config) {
+        os << "[" << section.first << "]" << "\n";
+        for (auto &keyValue : section.second) {
+            os << keyValue.first << "=" << keyValue.second << "\n";
+        }
+    }
+    return os;
+}
+
+class RandomFile {
```
paulsapps: hmm seems like "TempFile" is more accurate name?

madebr: True, will change that.

rwgame/GameConfig.cpp
```diff
@@ -155,6 +155,7 @@ bool GameConfig::parseConfig(
                     } else {
                       success = false;
                       RW_MESSAGE(e.what());
+                      return;
                     }
                 }
```
JayFoxRox: I think this entire code path could be nicer with an early out:
```
try {
   sourceValue = srcTree.get<config_t>(key, translator);
} catch (pt::ptree_bad_path &e) {
   if (!optional) {
      success = false;
      RW_MESSAGE(e.what());
      return;
   }
   sourceValue = defaultValue;
}
```
(Makes it more obvious that we abort when the argument is mandatory but failed to be parsed)

JayFoxRox: Also: Why don't we at least warn when the value could not be parsed? It should at least spit a warning if the value could not be parsed properly.

madebr: I included your suggestions in the next patch.
It also fixes a crash when parsing non-sensical data.

rwgame/GameConfig.cpp
```diff
@@ -62,25 +61,172 @@ std::string GameConfig::getDefaultConfigPath() {
...
+        }
+        srcTree.put(key, sourceValue, translator);
+
+        switch (destType) {
+            case ParseType::DEFAULT:
+                // Target cannot be DEFAULT (case already handled)
+                success = false;
+                break;
+            case ParseType::CONFIG:
+                //Don't care if success == false
```
JayFoxRox: nit: other comments start with a space

tests/test_config.cpp
```diff
@@ -1,20 +1,206 @@
...
+
+namespace fs = boost::filesystem;
+
+typedef std::map<std::string, std::map<std::string, std::string>> simpleConfig_t;
+
+simpleConfig_t getValidConfig() {
+    simpleConfig_t result;
+    // Don't change game.path and input.invert_y keys. Tests depend on them.
+    result["game"]["path"] = "\t/dev/test  \t \r\n";
+    result["game"]["\tlanguage\t "] = "      american ;american english french german italian spanish.";
```
JayFoxRox: is this intended? (weird spacing and escape sequences in what seems to be a key)

madebr: This is intentional.
I did that to test the robustness of the parser.

JayFoxRox: nit: maybe add a comment to make clear this is intended and maybe add some reasoning.

madebr: Yeah, I know.
I'm including your suggestions atm.
Patches will be coming.

rwgame/main.cpp
```diff
@@ -1,11 +1,10 @@
 #include <iostream>
 #include "RWGame.hpp"
 #include "SDL.h"
-
 #include <core/Logger.hpp>
 
 int main(int argc, char* argv[]) {
-    // Initialise Logging before anything else happens
+    // Initialize Logging before anything else happens
```
JayFoxRox: inb4 danhedron :gb:

madebr: Oops. Sorry!

I didn't notice the difference between British and American vocabulary at the moment of committing.
My IDE just noticed a spelling error.
Does openrw has a policy on this?

I will happily withdraw this commit.

danhedron: We don't have a proscribed spelling, so I don't think it's worth correcting regional spelling.

rwgame/GameConfig.cpp
```diff
@@ -62,25 +65,297 @@ std::string GameConfig::getDefaultConfigPath() {
...
+GameConfig::ParseResult GameConfig::parseConfig(
+    GameConfig::ParseType srcType, const std::string &source,
+    ParseType destType, std::string &destination)
+{
+    pt::ptree srcTree;
+    ParseResult parseResult(srcType, source, destType, destination);
+
+    try {
+        if (srcType == ParseType::STRING) {
+            pt::read_ini(source, srcTree);
```
danhedron: Unless I'm reading the documentation wrong, read_ini only accepts a stream or a file-path as an argument. It doesn't look like anything in this PR hits this path.

madebr: Yes. Nothing hits this path. Suggestion?
And while incorporating your suggestions, I introduced a bug here.
Fix in next commit.

rwgame/main.cpp
JayFoxRox: Last I checked all our colors were colours. I'm pretty sure we use BE across the codebase

madebr: The commit is withdrawn :)

danhedron:
LGTM. Any further configuration loading changes can be made in another PR once this is merged.

madebr:
What I will do in another PR:

- Keep bogus data in the log when saving. Right now these are deleted upon save.
  This feature will be interesting when people/developers switch between openrw versions.
  Configuration options will appear and disappear.
- Something else that I'm missing?

PR #262, opened 2017-01-11
Overhaul Clump, object and rendering Implementation

danhedron:
Overview of changes:
- Re-write the Clump / Atomic / Frame / Geometry code
- Each object has an Atomic/Clump as appropriate
- Animator animates Frame hierarchy directly
- Delete Skeleton
- Vastly simplify rendering logic
- Add LOD switching for vehicles
- Improve building LOD
- Remove inter-tick interpolation, this will need more work

Things to make a note of:
- Rendering is now 30Hz regardless of the framerate, this will be fixed later
- Performance is marginally improved on my machine
- No gameplay should be broken


haphzd:
LOD switching is nuts, middle section of Callahan bridge is even permanently missing. Some videos here - https://jii.moe/HyHTLjLIx.zip

danhedron:
The missing bridge was a bug in model changing. It was just sitting at the origin, setting the new atomic's frame resolves that.

danhedron:
This has been open a while. I've not seen major issues so we can fix anything else in follow ups.

PR #267, opened 2017-02-11
make the health indicator blink if low

vflyson:
There's no game->getSecond(), so currently made it disappear once every two game-minutes, even though the vanilla engine makes it blink exactly twice in a single game-minute.
The time runs slightly faster in the original game than so I thought it's okay for now.

### DIFF
--- a/rwgame/DrawUI.cpp
+++ b/rwgame/DrawUI.cpp
@@ -104,7 +104,7 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world,
     infoTextY += ui_textHeight;
 
     if (player->getCharacter()->getCurrentState().health > ui_lowHealth
-        || (int)(world->getGameTime() * 4.f) % 2) {
+        || fmod(world->getGameTime(), 0.5) >= .25) { // UI: Blinking health indicator if health is low
         std::stringstream ss;
         ss << std::setw(3) << std::setfill('0')
            << (int)player->getCharacter()->getCurrentState().health;
--- a/rwgame/DrawUI.cpp
+++ b/rwgame/DrawUI.cpp
@@ -17,6 +17,7 @@ constexpr size_t ui_ammoSize = 14;
 constexpr size_t ui_ammoHeight = 16;
 constexpr size_t ui_armourOffset = ui_textSize * 3;
 constexpr size_t ui_maxWantedLevel = 6;
+constexpr size_t ui_lowHealth = 9;
 const glm::u8vec3 ui_timeColour(196, 165, 119);
 const glm::u8vec3 ui_moneyColour(89, 113, 147);
 const glm::u8vec3 ui_healthColour(187, 102, 47);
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I'd suggest introducing a new variable in `void IngameState::tick(float dt)` to at least get the right behaviour.

Ideally we'll replace all of that with better code in the future? (I don't really like the autolook timer or my money hack either)

danhedron:
There's a floating point game timer ``GameState::gameTime``, that will give sub-second accuracy.

vflyson:
Taken care of accidentally including the submodules in the commit.

The `::tick(float dt)` way of doing it is probably more correct, but might be a bit too complex for my level of understanding of the current codebase.

Seems like using `gameTime` here worked quite well. It blinks with a pretty similar frequency to the original engine.

I agree with JayFoxRox and think that the code I'm committing here will eventually be replaced with the introduction of librw or just during one of code refactors in the future. I'm not too proud of this commit, but still feel like it's worth including just to make the UI behavior more complete at this stage.

danhedron:
If blinking twice is the correct behaviour (it looks like it is) could you make this blink twice per in-game minute. That way we can have the correct behaviour in place when comparing behaviour before & after if this code gets refactored.

Besides that, looks good to me.

vflyson:
Sure. Could you please hint me at what the multiplier for gameMinute is, relative to gameTime?
I'm currently using 6.0 (https://github.com/rwengine/openrw/pull/267/commits/42dbdb72f4cce6a6222d41037d8f2431848b9e83#diff-dc88e3c6f116fe7ce03ca1b2e3bf3668R107) but it's likely not half of an in-game minute.

JayFoxRox:
@vflyson I'm assuming the health blink is not tied to the game clock (`gameMinute`) but instead to a global timer (`gameTime`) (which may or may not pause in certain situations [pause menu etc.]). Did you try running the original game with cheats (do those exist) to advance the clock faster or something?

I'm not sure how fast any of these timers advance (but assume frames or seconds), so we still need @danhedron or you'll have to figure this out yourself :)

vflyson:
@JayFoxRox I've used adrenaline today (which on my 20th anniversary version freezes the in-game time completely) and the health indicator continued to blink with exactly the same frequency. It is most likely tied to the computer's clock in vanilla, not to in-game seconds of sorts.

Found a video on youtube where the health is low. The health indicator goes off 2 times and on 2 times as well in a single in-game minute, which makes it 1/4 of a game minute. I've updated the multiplier here, should be good to go.

rwgame/DrawUI.cpp
```diff
@@ -102,19 +103,20 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world,
 
     infoTextY += ui_textHeight;
 
-    {
+    if (player->getCharacter()->getCurrentState().health > ui_lowHealth
+        || (int)(world->getGameTime() * 4.f) % 2) {
```
JayFoxRox: how about `fmod`?
I'm totally against float pointers in general because they easily become inprecise due to float precision.
However, while we have them it looks like an even worse idea to convert them to integer.

vflyson: Thanks for the tip!

danhedron:
LGTM

PR #268, opened 2017-02-12
Logger colourify

PerikiyoXD:
Let's add colours to our Console!

Just added some colours to the Logger output. Hope it's now a little bit better for your eyes!

### DIFF
--- a/rwengine/src/core/Logger.cpp
+++ b/rwengine/src/core/Logger.cpp
@@ -38,12 +38,13 @@ void Logger::verbose(const std::string& component, const std::string& message) {
     log(component, Logger::Verbose, message);
 }
 
-std::map<Logger::MessageSeverity, char> severityStr{{Logger::Error, 'E'},
-                                                    {Logger::Warning, 'W'},
-                                                    {Logger::Info, 'I'},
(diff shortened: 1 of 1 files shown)


danhedron:
I'm not opposed to colour output, but getting colour output right in all terminals is quite difficult. I'd rather not start adding code to openrw for the esoteric art of figuring out if we should output colour or not.

This needs at least being able to turn off colour output with a command line switch, and making only output to the terminal coloured rather than all log listeners. Since some users might not want or be able to have colour output, and this is intended for output to a log file, where writing colour escape codes doesn't usually work.

PerikiyoXD:
*sigh*

The nearest thing to a standard are ANSI escape codes. Blame MS because their system symbols are just dumb, not me because i want to add color...

paulsapps:
https://github.com/paulsapps/alive/blob/dev/include/logger.hpp

This gets working colour output on osx/linux/windows

JayFoxRox:
I'd rather like to see a small script to colorfy the output if necessary.
We have further debugging capabilities using the debug menu ingame + annotated GL traces.

Ideally we'd drop / get rid of most console output as we don't want / need it in the long term (except for unexpected situations, which should be very very rare).

PR #269, opened 2017-02-13
apply various colors to pickups' coronas, use the correct texture

vflyson:
Comparing to what I see in the original game, the texture for big pickups is `coronaringa` while the texture for money and perhaps some other small pickups must be `coronacircle`.

The value of the 4th parameter of glm::vec4() is being omitted by the engine at the moment and does not affect the transparency of the sprite, it seems.

Neither I could adjust the size of the corona using `m_corona->particle.size = glm::vec2(0.5f, 0.5f);`.
Possibly the scaling has not been implemented yet; investigating that was outside the scope of this PR so I've decided not to.

Here is the color table that I made based on the colors which @aap extracted from the binary:
```
ff0000 bat

00ff00 pistol
8080ff uzi
ffff00 shotgun
ff00ff ak47
00ffff m16
ff8000 sniper
00ff80 rocket
8000ff flame
80ff00 molotov
ffffff grenade

ff0000 detonator
80ff80 armor, bribe
0000ff rampage
ffff00 health

000000 ???
```

### DIFF
--- a/rwengine/src/objects/PickupObject.cpp
+++ b/rwengine/src/objects/PickupObject.cpp
@@ -3,6 +3,23 @@
 #include <objects/CharacterObject.hpp>
 #include <objects/PickupObject.hpp>
 
+uint32_t colours[14] = {
+    0xff0000, // bat, detonator, adrenaline
+    0x00ff00, // pistol
+    0x8080ff, // uzi
+    0xffff00, // shotgun
+    0xff00ff, // ak47
+    0x00ffff, // m16
+    0xff8000, // sniper
+    0x00ff80, // rocket
+    0x8000ff, // flame
+    0x80ff00, // molotov
+    0xffffff, // grenade
+    0x80ff80, // bodyarmour, bribe
+    0x0000ff, // info, killfrenzy
+    0xffff00  // health, bonus
+};
+
 bool PickupObject::doesRespawn(PickupType type) {
     switch (type) {
         case Once:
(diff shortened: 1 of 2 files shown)


JayFoxRox:
My recommendation is to keep the color table seperated so it's easier to switch between GTA3, VC or even total conversions later. Also aap said on IRC that the alpha is 100% (probably getting all the transparency from the texture) whereas this seems to set it to 50%?
However, this should work for now.

aap:
Just for the record, this is how they compute the color for weapon pickups:
``out = color[weapontype] * 0.5 * (sin(timeInMs*0.0030664064)*0.3 + 0.3)``

vflyson:
@JayFoxRox I agree with using a table, however the current function does not supply any sort of ID aside from texture/model name, therefore I decided to postpone any further improvements related to colors. I assume the original engine had an integer associated with every weapon, which is `weapontype` within @aap's snippet? However info/health/armor/rampage/package/bonus are all pickups as well, and seem to have an identical corona sprite.

aap:
Yeah, it's not only weapons. Adrenaline and camera pickups are index 11, body armor and bribe 12, info and killfrenzy 13, health and bonus 14

vflyson:
It seems to me as if weapons have `coronaringa` as a texture while other pickups use `coronacircle`, possibly both scaled to the same size (except for money, but money perhaps belong to another group and just uses similar code).

vflyson:
So I'm not very proud of the last commit here, looking for suggestions on how to improve it.

Most likely the color table has to get moved into the world object instance. I need to figure out how to retrieve the model name within ::tick() to assign various colors depending on the pickup type.

A tip on how to do something like what R* did with function pointer address value as a seed to make every pulse start at a random time could also come in handy.

rwengine/src/objects/PickupObject.cpp
```diff
@@ -143,6 +131,43 @@ void PickupObject::tick(float dt) {
...
+    else if (modelinfo->name == "health" || modelinfo->name == "bonus")
+        coronaColourId = 14;
+#endif
+    float timeInMs = engine->getGameTime();
+    float m = 0.5f * (sin(timeInMs * 3.0664064f) * 0.3f + 0.3f);
+    float *colour = glm::value_ptr(colours[coronaColourId]);
+    m_corona->particle.colour = glm::vec4(colour[0] * m,
+                                          colour[1] * m,
+                                          colour[2] * m,
+                                          colour[3] * m);
```
danhedron: vec4 has an overload on the * operator for scalar multiplication, you can just write `m_corona->particle.colour = colours[coronaColourId] * m` instead of doing it component-wise.

rwengine/src/objects/PickupObject.cpp
```diff
@@ -143,6 +131,43 @@ void PickupObject::tick(float dt) {
...
+    else if (modelinfo->name == "grenade")
+        coronaColourId = 10;
+    else if (modelinfo->name == "bodyarmour" || modelinfo->name == "bribe")
+        coronaColourId = 12;
+    else if (modelinfo->name == "info" || modelinfo->name == "killfrenzy")
+        coronaColourId = 13;
+    else if (modelinfo->name == "health" || modelinfo->name == "bonus")
+        coronaColourId = 14;
+#endif
+    float timeInMs = engine->getGameTime();
```
danhedron: Since gameTime is currently in seconds, `timeInMs` should probably be named appropriately so if we start counting in milliseconds in the future we'll know what changes need to be made to the units in this function.

rwengine/src/objects/PickupObject.cpp
```diff
@@ -3,6 +3,23 @@
...
+    , glm::vec4(0.f, 1.f, 1.f, 1.f) // m16
+    , glm::vec4(1.f, 0.5f, 0.f, 1.f) // sniper
+    , glm::vec4(0.f, 1.f, 0.5f, 1.f) // rocket
+    , glm::vec4(0.5f, 0.f, 1.f, 1.f) // flame
+    , glm::vec4(0.5f, 1.f, 0.f, 1.f) // molotov
+    , glm::vec4(1.f, 1.f, 1.f, 1.f) // grenade
+    , glm::vec4(0.5f, 1.f, 0.5f, 1.f) // bodyarmour, bribe
+    , glm::vec4(0.f, 0.f, 1.f, 1.f) // info, killfrenzy
+    , glm::vec4(1.f, 1.f, 0.f, 1.f) // health, bonus
+};
```
JayFoxRox: Suggestion: Use original format (so a byte array or uint32 array) as it appeared in GTA memory for these structures and construct the glm vector where necessary.
That will make it easier to reuse data in other projects (which may not use glm) + it will make it easier to add new games like VC where we'd otherwise add a lot of manual work to do the conversion before adding the data here.

I'm not sure if this is how arrays are currently formated.

rwengine/src/objects/PickupObject.cpp
```diff
@@ -115,6 +131,39 @@ void PickupObject::tick(float dt) {
         }
     }
 
+    int coronaColourId = 0;
+#if 0
```
JayFoxRox: Add a note here why this is disabled and this `#if` exists in the first place

PR #270, opened 2017-02-18
Fix weapon scrollback

vflyson:
The order of the weapons in the inventory seems to be a bit off comparing to vanilla.
Couldn't find where to alter that quickly and it's probably enough commits for a single PR so next time Ü

### DIFF
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -586,7 +586,8 @@ void CharacterObject::cycleInventory(bool up) {
         currentState.currentWeapon = 0;
     } else {
         for (int j = currentState.currentWeapon - 1; j >= 0; --j) {
-            if (currentState.weapons[j].weaponId != 0) {
+            bool isFist = j == 0;
+            if (currentState.weapons[j].weaponId != 0 || isFist) {
                 currentState.currentWeapon = j;
                 return;
             }
@@ -594,7 +595,8 @@ void CharacterObject::cycleInventory(bool up) {
 
         // Nothing? set the highest
         for (int j = kMaxInventorySlots - 1; j >= 0; --j) {
-            if (currentState.weapons[j].weaponId != 0 || j == 0) {
+            bool isFist = j == 0;
+            if (currentState.weapons[j].weaponId != 0 || isFist) {
                 currentState.currentWeapon = j;
                 return;
             }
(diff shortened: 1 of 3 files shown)


rwengine/src/objects/CharacterObject.cpp
```diff
@@ -586,7 +586,7 @@ void CharacterObject::cycleInventory(bool up) {
         currentState.currentWeapon = 0;
     } else {
         for (int j = currentState.currentWeapon - 1; j >= 0; --j) {
-            if (currentState.weapons[j].weaponId != 0) {
+            if (currentState.weapons[j].weaponId != 0 || j == 0) {
```
JayFoxRox: A comment wouldn't hurt

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -573,6 +573,8 @@ void CharacterObject::removeFromInventory(int slot) {
 }
 
 void CharacterObject::cycleInventory(bool up) {
+    if (currentVehicle != nullptr) return;
```
JayFoxRox: Maybe this should happen when calling this function instead of in this function?
I'd assume character functions to work at all times, however, input in cars should be treated differently and that processing is probably done where this is called from.

JayFoxRox:
I only have some tiny nits left which I'll keep to myself.

Code LGTM but I never tried running it / untested.

vflyson:
Do tell, I want it to be perfect!
Oh, you're missing out.

danhedron:
LGTM, changes work as expected

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -586,15 +586,19 @@ void CharacterObject::cycleInventory(bool up) {
         currentState.currentWeapon = 0;
     } else {
         for (int j = currentState.currentWeapon - 1; j >= 0; --j) {
-            if (currentState.weapons[j].weaponId != 0) {
+            bool isFist = j == 0;
+            if (currentState.weapons[j].weaponId != 0
+                || isFist // this bit here is to let us scroll down to the fist
+                ) {
```
JayFoxRox: nit: I believe you can remove the comment now and move it all in one line.
If you keep the comment it should probably explain that the standard check wouldn't work (presumably) because fist is weapon 0 (so the check will always fail), also you always have a fist pickup

PR #271, opened 2017-02-19
Implement basic Anim groups

danhedron:
- Refactor Animation Groups
- Add hard-coded data for each animation group
- Refactor CharacterController to remove usage of Animation* (mostly, still talks directly to Animator in a few places)
- Apply the correct animation group to every character

I don't like the hard-coded data in AnimGroup.cpp, any ideas on how that could be better written is appreciated.

### DIFF
--- a/rwengine/src/data/AnimGroup.cpp
+++ b/rwengine/src/data/AnimGroup.cpp
@@ -1,39 +1,393 @@
 #include "AnimGroup.hpp"
+#include <algorithm>
+#include <vector>
 
-AnimGroup::AnimGroup(AnimationSet &animations, const std::string &name)
-    : idle(find(animations, name, "idle_stance"))
-    , walk(find(animations, name, "walk_player"))
-    , walk_start(find(animations, name, "walk_start"))
-    , run(find(animations, name, "run_player"))
-    , sprint(find(animations, name, "sprint_civi"))
-    , walk_right(find(animations, name, "walk_player_right"))
-    , walk_right_start(find(animations, name, "walk_start_right"))
-    , walk_left(find(animations, name, "walk_player_left"))
-    , walk_left_start(find(animations, name, "walk_start_left"))
-    , walk_back(find(animations, name, "walk_player_back"))
-    , walk_back_start(find(animations, name, "walk_start_back"))
-    , jump_start(find(animations, name, "jump_launch"))
-    , jump_glide(find(animations, name, "jump_glide"))
-    , jump_land(find(animations, name, "jump_land"))
-    , car_sit(find(animations, name, "car_sit"))
-    , car_sit_low(find(animations, name, "car_lsit"))
-    , car_open_lhs(find(animations, name, "car_open_lhs"))
-    , car_getin_lhs(find(animations, name, "car_getin_lhs"))
-    , car_getout_lhs(find(animations, name, "car_getout_lhs"))
-    , car_pullout_lhs(find(animations, name, "car_pullout_lhs"))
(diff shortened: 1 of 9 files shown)


rwengine/src/data/AnimGroup.cpp
```diff
@@ -179,6 +179,162 @@ AnimGroup AnimGroup::getBuiltInAnimGroup(AnimationSet &animations,
         {"phone_talk"},
     }};
 
+    if (name == "player") {
+        group.animations_[0].name = "walk_player";
+        group.animations_[1].name = "run_player";
+        group.animations_[2].name = "sprint_civi";
+        group.animations_[3].name = "idle_stance";
+        group.animations_[4].name = "walk_start";
+    }
```
JayFoxRox: nit: I feel these could be done differently. Otherwise one might accidentally use a wrong index somewhere and it will screw everything up.

danhedron: I agree. I will replace this with something that's harder to get wrong.

danhedron:
I've cleaned up the hard-coded animation group data.

rwengine/src/data/AnimGroup.cpp
```diff
@@ -1,39 +1,390 @@
...
+             {"duck_low", 0x14},
+             {"rblock_cshoot", 0x14},
+             {"weapon_throwu", 0x18},
+             {"handsup", 0x50},
+             {"handscower", 0x58},
+             {"fucku", 0x21c},
+             {"phone_in", 0x10},
+             {"phone_out", 0x18},
+             {"phone_talk", 0x16}}},
+    {
```
JayFoxRox: What happened here? The formatting / indenting looks a bit off somehow

rwengine/src/data/AnimGroup.cpp
```diff
@@ -1,39 +1,390 @@
...
+AnimGroup AnimGroup::getBuiltInAnimGroup(AnimationSet &animations,
+                                         const std::string &name) {
+    auto findgroup = [&]() {
+        auto it =
+            std::find_if(kBuiltInAnimGroups.begin(), kBuiltInAnimGroups.end(),
+                         [&](const AnimGroup &g) { return g.name_ == name; });
+        if (it != kBuiltInAnimGroups.end()) {
+            return *it;
+        } else {
+            return kBuiltInAnimGroups[0];
```
JayFoxRox: Are we sure we silently want to return the first group?
If this is a commonly used group (and this is meant as a shortcut) we should probably make name an optional (default nullptr) argument and return the group based on that.

To me this looks like a situation which should abort.

rwengine/src/data/AnimGroup.cpp
danhedron: The first line of this set of braces is short enough that clang-format doesn't break the inner array onto a new line so it tries to compact the end as well. It's ugly but I don't know if there's anything I can do to stop it short of making the first array entry longer somehow.

rwengine/src/data/AnimGroup.cpp
danhedron: Looking at the input data it shouldn't ever ask for anything that doesn't exist. This is a failure case for sure.

rwengine/src/data/AnimGroup.cpp
danhedron: After a bit of twiddling I noticed this is a feature of clang-format. Adding a final comma to a list initialisation causes the list to be expanded over multiple lines instead of compacted. I've made all the entries consistent now.

PR #272, opened 2017-02-19
More weapon scrollback fixes

vflyson:
Made a few more small fixes to the weapon inventory system.
The main issue that I currently see is that weapons which have clips don't get reloaded by themselves when picked up or scrolled to like in the vanilla engine. Most likely there needs to be a separate method to accomplish that which we'd call from both UseItem, onCharacterTouch and from cycleInventory.

### DIFF
--- a/rwgame/DrawUI.cpp
+++ b/rwgame/DrawUI.cpp
@@ -185,18 +185,27 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world,
         const CharacterState& cs = player->getCharacter()->getCurrentState();
         const CharacterWeaponSlot& slotInfo = cs.weapons[cs.currentWeapon];
 
-        bool isProjectile = weapon->fireType == WeaponData::PROJECTILE;
-        bool isShotgun = weapon->modelID == 176;
-        bool isSniper = weapon->modelID == 177;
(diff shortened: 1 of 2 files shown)


vflyson:
Okay, after discovering something called Wine and playing GTA-III 1.0 while adjusting values of column F in `data/weapon.dat`, found out that:

 - setting F to 1 will make guns reload after every shot, the clip + total will be merged into one value in the UI
 - setting F to 2 or more for projectiles will show a clip size in the UI, but they will reload after every shot/throw just like always (purely a matter of displaying the ammo count in that case)
 - setting F to >= 1000 will merge clip + total into one number
 - setting F to 0 for firearms will make it never reload that weapon, merges total + clip in the UI as well; works surprisingly similar to setting it to 1000, but I guess R* decided that it's rather inaccurate to use clip size of 0 and went for > 999 as a condition for having no clip size displayed

So a number in column F of 2 through 999 (inclusive) will make the clip size value visible.
Melee does not have ammo count, but projectiles hide the amount of ammo in the clip not through some hardcoded value, but through that "1" in `weapon.dat`'s F column.

danhedron:
LGTM

PR #275, opened 2017-02-21
Add an SDL window icon to the executable

vflyson:
Beware: `rwgame/WindowIcon.h` contains over 10K LOC, do not be tempted to click on "Load diff".

Since including SDL2_image just for a single PNG file might be an overkill, I've decided to add the icon this way.
Went for the max. resolution of the icon that I managed to find, since it's 2017 and all that...


JayFoxRox:
Which logo is it?

Also most platforms expect their logos in various resolutions. Creating the mac one is the most trouble as they use some proprietary format where good tools only exist for apples own OS.

So I guess this would work as a temporary solution. In the long run we should follow platform guidelines and embed the icon how the platforms want it (resource files for Windows, linked in for Mac, freedesktop for others)

vflyson:
OpenRW Logo
It's this one.

I wasn't aware about different formats being needed for various platforms, I thought SDL takes care of it.

danhedron:
I guess @JayFoxRox is talking about the application icon itself, which will require a .desktop file for Linux, and PE resources for Windows.

I think we can probably use a small icon for the window, perhaps 1/2 as big at least. There are some ugly resizing artefacts visible:

icon_dock
activity_icon

vflyson:
Gradually downscaled the 256x256 icon by 10% to 128x128, seems to look prettier.
Please let me know if needs any more tweaks.

danhedron:
openrw_resized
Looks much better.

vflyson:
I couldn't find a way for SDL to store multiple resolutions for the window titlebar icon, and it scales things up/down rather crudely, so I guess this is as good as it gets until I figure out how to force SDL to retain the initial icon (inherited from a .desktop file or some other launcher).

JayFoxRox:
I might have been wrong @vflyson : http://stackoverflow.com/questions/19311609/how-do-i-set-window-titlebar-icon-and-alt-tab-icon-in-sdl2-c-visual-studio

I'll probably check other projects again soon as I thought I saw this happening. We can also talk to the SDL2 people.

vflyson:
I've tried to search for other people using SDL to store icons in multiple resolutions, but I rather failed to find anything of such nature.

danhedron:
I don't think that's a major concern, someone can always look into it later. LGTM.

rwgame/WindowIcon.h
```diff
@@ -0,0 +1,2659 @@
+unsigned int windowIconWidth = 128;
+unsigned int windowIconHeight = 128;
+unsigned int windowIconBitsPerPixel = 32;
+
+unsigned char windowIconData[] =
```
JayFoxRox: yuk.. Please use a hex array with proper line length, not a string.

rwgame/GameWindow.cpp
```diff
@@ -31,11 +31,39 @@ void GameWindow::create(const std::string& title, size_t w, size_t h,
         throw std::runtime_error("SDL_GL_CreateContext failed: " + sdlErrorStr);
     }
 
+#include "WindowIcon.h"
+
+    Uint32 rmask, gmask, bmask, amask;
+
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
+    // Big Endian
+    int offset = (windowIconBitsPerPixel == 32) ? 0 : 8;
```
JayFoxRox: You should probably assert that `(windowIconBitsPerPixel == 24) || (windowIconBitsPerPixel == 32)` ?
(This is what the code suggests)

darkf:
Why is this being stored in the code (a la XPM) instead of a file?

vflyson:
@darkf: Since including SDL2_image just for a single PNG file might be an overkill, I've decided to add the icon this way.
.

darkf:
@vflyson Then give it its own binary file or BMP or something that's not confusing and people can edit independently, not a gigantic C file.

rwgame/WindowIcon.h
vflyson: You never like what I make! Breaks my heart

rwgame/GameWindow.cpp
vflyson: In this case the image buffer is always 32-bit, since nobody really uses square icons nowadays. I'll just remove the bpp lines altogether, the image's hardcoded anyway.

JayFoxRox:
@darkf is probably right btw.

We should keep the image in the repo and if we want to avoid SDL_image we should just see if there is a cmake build step we could do to generate the logo. At the very least it should be documented how to turn the graphic into a logo and both files should be in the repo.

vflyson:
I've formatted the byte array to fit under 81 characters, please let me know if it needs more work.
Made bitsPerPixel thing go, since 32 bits is what icons use.

Regarding having a blob file instead of a giganitc C file: wish it was that simple.
SDL_CreateRGBSurfaceFrom needs to have what this branch currently has on the input, and that seems to be the only way to set a window icon in SDL2 without using SDL_image.

SDL_image is another dependency which would require an additional CMAKE file. Aside from that, it'd need an icon file within the filesystem

rwgame/WindowIcon.h
```diff
@@ -0,0 +1,4103 @@
...
+0xea,0x93,0x55,0xff,0xea,0x93,0x55,0xff,0xea,0x93,0x55,0xff,0xea,0x93,0x55,0xff,
+0xea,0x93,0x55,0xff,0xea,0x93,0x55,0xff,0xea,0x93,0x55,0xff,0xe9,0x92,0x55,0xff,
+0xf4,0x98,0x58,0xff,0xd0,0x82,0x4c,0xff,0x2d,0x1c,0x10,0xff,0x00,0x00,0x00,0xff,
+0x00,0x00,0x00,0xff,0x00,0x00,0x00,0xff,0x04,0x02,0x03,0xff,0x57,0x21,0x33,0xff,
+0xc2,0x4b,0x73,0xff,0xc1,0x4b,0x72,0xff,0xbb,0x49,0x6f,0xff,0xbb,0x49,0x6f,0xff,
+0xbb,0x48,0x6f,0xff,0xba,0x48,0x6f,0xff,0xba,0x48,0x6f,0xff,0xba,0x48,0x6f,0xff,
+0xba,0x48,0x6f,0xff,0xb9,0x48,0x70,0xff,0xb8,0x48,0x70,0xff,0xb9,0x48,0x70,0xff,
+0xb9,0x47,0x70,0xff,0xb8,0x47,0x70,0xff,0xb8,0x47,0x70,0xff,0xb7,0x47,0x70,0xff,
+0xb7,0x47,0x70,0xff,0xb7,0x47,0x70,0xff,0xb7,0x47,0x70,0xff,0xb8,0x48,0x71,0xff,
+0xc0,0x4a,0x76,0xff,0x5e,0x24,0x39,0xff,0x02,0x
```
JayFoxRox: How can this have an odd size? O.o

vflyson: It's been generated from what GIMP gave me, "GIMP RGBA C-Source image dump" is the format.
I assume it has that last byte in it to signal the end of the buffer.

```
#include <stdio.h>
#include <stdlib.h>

#define COL 16

#include "../openrw/rwgame/WindowIcon.h"

void main()
{
    int i = 0;
    int size = sizeof(windowIconData);

    while (i < size) {
        if ((i > 0) && (i % COL == 0)) {
            printf("\n");
        }

        printf("0x%02x", windowIconData[i]);
        printf(",");

        i++;
    }

    puts("");
}
```
is what I used to convert it to an array of bytes.

darkf:
I mean... Parsing BMP is trivial, [stb_image](https://github.com/nothings/stb/blob/master/stb_image.h) isn't big and is header-only, etc. Hell if you want to get creative, package it as a TXD. ;) (Please don't actually.)

I just don't see the point of XPMs in 2017 and it is inconvenient.

rwgame/WindowIcon.h
vflyson: The "bonus" byte was from the \0 of the string that GIMP gave me initially, since the genius programmer here used sizeof() where he should've used strlen(). Good catch as always, J!

vflyson:
Parsing BMP might be trivial, but parsing any external file would mean that that thing would have to "hang in" somewhere in the filesystem, most likely an additional dir created in the repository, etc. So what's more inconvenient is debatable. This code is very simple to read and modify right now, and it's just an icon anyway, not a big deal. As JFR would say, it's "bikeshedding". I think he likes that term.

Added a bit of instructions within the code, right where the snippet's located. It's not exactly a tutorial, but gives a good hint at what the buffer's made of.
And btw, that `*cough* convert d3e29328-76de-11e6-8936-d549e5d71b6a.png -define h:format=rgba -depth 8 -size 128x128 foo.h *cough*` didn't produce the right array from the image for SDL_CreateRGBSurfaceFrom, it must be a different kind of RGBA buffer. And yes, I've tried it both with `*cough*`s and without.

Oh, and could you please put "please don't actually" in front or at least in the middle of the sentence next time? I've had it implemented almost all the way before I finished reading your comment.

madebr:
What @darkf said + [stb_image_resize.h](https://github.com/nothings/stb/blob/master/stb_image_resize.h) + header creation tool + generates the header at build time ([add_custom_command](https://cmake.org/cmake/help/v3.8/command/add_custom_command.html).

vflyson:
Thank you for the comments and reviews. I'll close this PR and start over with a new branch using the method suggested in this thread.

ShFil119:
Any progress on it?

JayFoxRox:
@vflyson 

> As JFR would say, it's "bikeshedding". I think he likes that term.

Yes indeed. And looking back at this PR I hate how it was handled, and get the same bad feeling for the new iteration of it.

> That "*cough* `convert d3e29328-76de-11e6-8936-d549e5d71b6a.png -define h:format=rgba -depth 8 -size 128x128 foo.h` *cough*" didn't produce the right array [editors note: This command line was suggested by me via IRC]

I've only seen this comment now, not sure if I ever responded anywhere. It's possible that it's the wrong format, but you can construct a custom https://wiki.libsdl.org/SDL_PixelFormat for it, swap bytes around to a temporary buffer, or use another command line.

> Oh, and could you please put "please don't actually" in front or at least in the middle of the sentence next time? I've had it implemented almost all the way before I finished reading your comment.

I'm not even sure why darkf said that to be honest. I think it's a fine idea, as TXD is just a container format for pixels, and we support loading it anyway. It even supports compressed textures.

darkf:
I thought people actually read comments before enacting them, deary me... It's not the worst idea in the world, but as I mentioned in the other PR a BMP would probably make more sense. TXDs are not editable without specialized tools AFAIK.

PR #276, opened 2017-02-21
Add save directory for Windows

SonofUgly:
Tested and working on Windows 8.
Part of: [Tasks#Savegames](https://github.com/rwengine/openrw/wiki/Tasks#savegames)

### DIFF
--- a/rwengine/src/engine/SaveGame.cpp
+++ b/rwengine/src/engine/SaveGame.cpp
@@ -1313,13 +1313,18 @@ bool SaveGame::getSaveInfo(const std::string& file, BasicState* basicState) {
 std::vector<SaveGameInfo> SaveGame::getAllSaveGameInfo() {
     using namespace boost::filesystem;
 
-    // TODO consider windows
+#if defined(RW_WINDOWS)
+    auto homedir = getenv("USERPROFILE");
+    const char gameDir[] = "Documents\\GTA3 User Files";
+#else
     auto homedir = getenv("HOME");
+    const char gameDir[] = "GTA3 User Files";
+#endif
+
     if (homedir == nullptr) {
         std::cerr << "Unable to determine home directory" << std::endl;
         return {};
     }
-    const char gameDir[] = "GTA3 User Files";
 
     path gamePath(homedir);
     gamePath /= gameDir;


rwengine/src/engine/SaveGame.cpp
```diff
@@ -1313,13 +1313,18 @@ bool SaveGame::getSaveInfo(const std::string& file, BasicState* basicState) {
 std::vector<SaveGameInfo> SaveGame::getAllSaveGameInfo() {
     using namespace boost::filesystem;
 
-    // TODO consider windows
+#if defined(RW_WINDOWS)
+    auto homedir = getenv("USERPROFILE");
+    const char gameDir[] = "\\My Documents\\GTA3 User Files";
+#else
     auto homedir = getenv("HOME");
+    const char gameDir[] = "GTA3 User Files";
```
JayFoxRox: That's very un-posix-ish / un-freedesktop-ish.
We should probably use a proper dot directory for the settings and local program data.

rwengine/src/engine/SaveGame.cpp
```diff
@@ -1313,13 +1313,18 @@ bool SaveGame::getSaveInfo(const std::string& file, BasicState* basicState) {
 std::vector<SaveGameInfo> SaveGame::getAllSaveGameInfo() {
     using namespace boost::filesystem;
 
-    // TODO consider windows
+#if defined(RW_WINDOWS)
+    auto homedir = getenv("USERPROFILE");
+    const char gameDir[] = "\\My Documents\\GTA3 User Files";
```
JayFoxRox: If the non-windows one doesn't start with a backslash, why does this one?

madebr: For Windows, I would suggest using [SHGetKnownFolderPath](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762188(v=vs.85).aspx) and [FOLDERID_SavedGames](https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457(v=vs.85).aspx).

haphzd: "My Documents" is a localized name, use "Documents" instead, or what madebr said (but with FOLDERID_Documents).

haphzd: "My Documents" is a localized name, use "Documents" instead, or what madebr said (but with FOLDERID_Documents).

paulsapps: Example of the SH* API: https://github.com/paulsapps/alive/blob/a87957409b574e5a12ee85b9a9a1c0de91255f77/include/resourcemapper.hpp#L466

SonofUgly: Just changed it to `"Documents\\GTA3 User Files"` for now.
Seams using `SHGetKnownFolderPath` means adding more includes and duplicating a bit more code? Having some issues (don't really know what I'm doing), so perhaps it'd be best left for someone else if you want it like that.

danhedron: I may be wrong but using ``SHGetKnownFolderPath`` may be required for unicode usernames? Does the original game handle such paths?

Unless there is an issue including the Windows API it should be possible to call ``SHGetKnownFolderPath``. The main issue would be the wchar_t since OpenRW assumes UTF-8.

JayFoxRox:
How do we go forward with this?

Does the current implementation work:

1. on all localizations?
2. on what version of windows?
3. with unicode usernames?
4. is it compatible to the original game save folders?

If 4 is not true, how do we approach this?

Aside from that it's probably better than master, so if we don't get a response soon we should create a new issue for this and merge this one.

madebr:
We will know (with a higher confidence) when Windows Continuous Integration is configured. Appvoyer?

SonofUgly:
>is it compatible to the original game save folders?

Yes.
For localizations and Windows versions, I only have one Windows install so I can only say English and Windows 8+ for certain, but from what I've read I think it should work on other localizations and XP+. Unicode usernames, IDK.

darkf:
IMO you should be using SHGetKnownFolderPath, which gives you the canonical location for game storage locations (e.g. My Games or My Documents depending on the version.), as mentioned.

This doesn't really add much if anything if it doesn't do that.

madebr:
I've got utf16 to utf8 conversion (and vice versa) working on Windows.
I will open a PR in the coming days.

paulsapps:
I've got code for the utf8 conversion etc here: https://github.com/paulsapps/alive/blob/ef648dde38d189cb763efb8f365304088cf51bc4/src/gamefilesystem.cpp#L58

This create a dir in Libraries\Documents which is the same as C:\users\username\documents

SonofUgly:
#376

PR #277, opened 2017-02-23
Implement in-car perks

vflyson:
This is quite primitive yet works fine.
Somehow I've kept running into one trouble after another, spending way too much time on trying to get $25 out of cabs. There seems to be no easy way to increase the amount of money from the inside of that class.

My wild guess is that this all should be happening via OpCodes, and perhaps within some other class. Every vehicle instance should also have a flag which shows that the perk has already been collected (which gets set to true if the car got spawned out of a garage).

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -237,6 +237,9 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
             }
         } else if (character->getCurrentCycle() == cycle_enter) {
             if (character->animator->isCompleted(AnimIndexAction)) {
+                /// @todo move to a more suitable place
+                vehicle->grantOccupantRewards(character);
+
                 // VehicleGetIn is over, finish activity
                 return true;
             }
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -384,6 +384,10 @@ void CharacterObject::setPosition(const glm::vec3& pos) {
     getClump()->getFrame()->setTranslation(pos);
 }
 
+bool CharacterObject::isPlayer() const {
+    return true; /// @todo implement detection via playerObject
+}
+
 bool CharacterObject::isAlive() const {
     return currentState.health > 0.f;
 }
--- a/rwengine/src/objects/CharacterObject.hpp
+++ b/rwengine/src/objects/CharacterObject.hpp
@@ -113,6 +113,8 @@ public:
 
     virtual void setPosition(const glm::vec3& pos);
 
+    bool isPlayer() const;
+
     bool isAlive() const;
     bool takeDamage(const DamageInfo& damage) override;
 
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -86,6 +86,7 @@ VehicleObject::VehicleObject(GameWorld* engine, const glm::vec3& pos,
     , throttle(0.f)
     , brake(0.f)
     , handbrake(true)
+    , mHasSpecial(true)
     , info(info)
     , colourPrimary(prim)
     , colourSecondary(sec)
@@ -773,4 +774,37 @@ void VehicleObject::setSecondaryColour(uint8_t color) {
 
 bool VehicleObject::isStopped() const {
     return fabsf(physVehicle->getCurrentSpeedKmHour()) < 0.75f;
+}
+
+bool VehicleObject::collectSpecial() {
+    bool hadSpecial = mHasSpecial;
+
+    mHasSpecial = false;
+
+    return hadSpecial;
+}
+
+void VehicleObject::grantOccupantRewards(CharacterObject* character) {
+    if (character->isPlayer() && collectSpecial()) {
+        if (getVehicle()->vehiclename_ == "TAXI"
+            || getVehicle()->vehiclename_ == "CABBIE"
+            || getVehicle()->vehiclename_ == "BORGNINE") {
+            // Earn $25 from taxi cabs
+            /// @todo implement this
+        } else if (getVehicle()->vehiclename_ == "POLICAR") {
+            // Police cruisers give 5 shotgun cartridges
+            character->addToInventory(4, 5);
+        } else if (getVehicle()->vehiclename_ == "ENFORCR") {
+            // Obtain 100 armour if it's an Enforcer
+            character->getCurrentState().armour = 100.f;
+        } else if (getVehicle()->vehiclename_ == "AMBULAN") {
+            // Receive up to 20 HPs from ambulances
+            float health = character->getCurrentState().health;
+            float fullHealth = 100.f;
+            if (health < fullHealth) {
+                character->getCurrentState().health =
+                    (health >= 80.f) ? fullHealth : health + 20.f;
+            }
+        }
+    }
 }
--- a/rwengine/src/objects/VehicleObject.hpp
+++ b/rwengine/src/objects/VehicleObject.hpp
@@ -2,6 +2,7 @@
 #ifndef _VEHICLEOBJECT_HPP_
 #define _VEHICLEOBJECT_HPP_
 #include <map>
+#include <objects/CharacterObject.hpp>
 #include <objects/GameObject.hpp>
 #include <objects/VehicleInfo.hpp>
 
@@ -29,6 +30,7 @@ public:
     VehicleInfoHandle info;
     glm::u8vec3 colourPrimary;
     glm::u8vec3 colourSecondary;
+    bool mHasSpecial;
 
     std::map<size_t, GameObject*> seatOccupants;
 
@@ -150,6 +152,14 @@ public:
     * @return True if the vehicle isn't moving
     */
     bool isStopped() const;
+
+    /**
+    * @brief collectSpecial
+    * @return True if mHasSpecial was true opon calling
+    */
+    bool collectSpecial();
+
+    void grantOccupantRewards(CharacterObject* character);
 
 private:
     void registerPart(ModelFrame* mf);


rwengine/src/ai/CharacterController.cpp
```diff
@@ -239,6 +239,28 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
...
+                    character->getCurrentState().armour = 100.f;
+                } else if (vehicle->getVehicle()->vehiclename_ == "AMBULAN") {
+                    // Receive up to 25 HPs from Ambulance
+                    if (character->getCurrentState().health < 100.f) {
+                        if (character->getCurrentState().health > 75.f) {
+                            character->getCurrentState().health = 100.f;
+                        } else {
+                            character->getCurrentState().health += 25.f;
+                        }
+                    }
```
JayFoxRox: What happens in VC? [does GTA3 have raised health too?]
Are we certain that this is capped to 100 and not the maximum heath? (150 etc.?)

You can probably solve this easiest by creating a temporary variable and clamping the health before assigning it.
Also the `100.f` should then probably be a variable called `maxHealth` or something (in the player class or something or optionally in this code branch with a `// todo: Replace by correct maxHealth once it is implemented`)

I believe I made a similar mistake in the cheat-handler.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -239,6 +239,28 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
         } else if (character->animator->getAnimation(AnimIndexAction) ==
                    anm_enter) {
             if (character->animator->isCompleted(AnimIndexAction)) {
```
JayFoxRox: Note that animations do not necessarily describe what goes on in the game world.
The player might be warped into the car if the doors are not accessible. Similarily the animation might not play if the script puts the player in a vehicle (will rewards still work?!) etc.

I feel like we should instead have some kind of callback which signals that a player has entered a vehicle (possibly informing about seatIndex, vehicle, character (not necessarily as we can probably retrieve it from the seat)).

However, the current implementation will probably work too.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -239,6 +239,28 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
...
+                if (vehicle->getVehicle()->vehiclename_ == "TAXI"
+                    || vehicle->getVehicle()->vehiclename_ == "CABBIE"
+                    || vehicle->getVehicle()->vehiclename_ == "BORGNINE") {
+                    // Earn $25 from taxi cab
+                    // @todo
+                } else if (vehicle->getVehicle()->vehiclename_ == "POLICAR") {
+                    // Police cruisers give 5 shotgun cartridges
+                    character->addToInventory(4, 5);
```
JayFoxRox: Do these `character` events also happen for non-player characters?

JayFoxRox:
This should not happen through OpCodes.
These bonuses are hardcoded and independent of the script processor (afair).

You can probably check my cheat-handler to see how I incremented money, maybe you can figure something out.
If this is merged in current state we should have 3 new issues about these perks:
- Does not collect vehicle bonus if not entering car through animation
- Non-players collect vehicle bonus
- Vehicles will repeatedly give out a vehicle bonus

Some of these behaviours would have a rather strong gameplay impact (cheating and possibly messing with AI or even core game missions, think of unarmed AI using a cop car and suddenly having a weapon). So make sure we get them right.

rwengine/src/ai/CharacterController.cpp
vflyson: It's both 20 HPs in VC and in GTA-III for Ambulances, my bad!

rwengine/src/ai/CharacterController.cpp
vflyson: That hasn't crossed my mind, but now I'm thinking that it probably does give bonuses to non-player character objects, since it's within the `src/ai/` directory it's likely covering both player and NPCs.
What'd be the best way to make sure the character object is the player?

JayFoxRox: No idea - I'm sure there'll be some hints somewhere in the codebase.

Please check with the original game wether this might actually be the correct behaviour or prepare to create an issue on merge.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -252,12 +252,12 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
...
-                            character->getCurrentState().health = 100.f;
-                        } else {
-                            character->getCurrentState().health += 25.f;
-                        }
+                    float health = character->getCurrentState().health;
+                    // todo: Replace by correct fullHealth once implemented
+                    float fullHealth = 100.f;
+                    if (health < fullHealth) {
+                        character->getCurrentState().health =
+                            (health >= 80.f) ? fullHealth : health + 20.f;
```
JayFoxRox: * The comment does not match this anymore
* I'd suggest something like `std::min(health + 20.f, fullHealth)` but I see that it could give problems with prostitute health boosts which temporarily raise the health higher than fullHealth. So it looks like we'd still require a check `if (health < fullHealth)` to avoid lowering the health.
* You still have to check with the original game wether this also happens `if (health > 100.f)` when fullHealth is greater than 100.

Optionally create issues after merge so we can check this later

rwengine/src/ai/CharacterController.cpp
vflyson: That's a tricky part. I've searched quite a bit, couldn't find any snippets in the code yet. I thought of making sure the character's model is Claude, that perhaps would work better than matching the XYZ coords of the player model against the character object.

As for the original game -- I've been giving rides to NPCs in police cruisers during missions, they never get the shotgun. I don't recall witnessing an  NPC stealing a special vehicle ever. Thieves only steal fancy cars, so neither taxis nor cruisers/ambulances can be initially boarded by non-player characters inside the game, at least to my knowledge.
I have not seen a case where a player would enter a special vehicle as a non-driver as well. It is likely that the "bonus" gets depleted once the vehicle gets boarded by the player character for the first time.

If you could give me a creative hint at how to test it, aside from REing the thing (which I'm honestly, below horrible with), I'd really appreciate it.

JayFoxRox: Create an issue about it, we need a mission script to test it then (or RE).

Checking for player has to be done using the playerchar pointer or something - @danhedron can probably tell you how, but we can also delay the check to later once we know if this affects actors (players, referenced [mission chars] and derefenced [peds])

rwengine/src/ai/CharacterController.cpp
vflyson: I've tested a few times and the game mechanics of GTA-III never raises the health above 100 from ambulances, it does waste the bonus inside that car though, so the second time the player would enter the vehicle the bonus would not get triggered even if the player's health will be below 100.

JayFoxRox: even with a 100% game [at which point you can have more than 100 HP from health pickups]?
You can always have more than 100% health from prostitutes.
Also good to know about the flag still being cleared :)

vflyson: That I don't know, don't have a 100% finished savegame. Perhaps should get one and test it.

Prostitutes give up to extra 25 health above maximum of 100 in GTA-III. If the 100% game completion raises the fullHealth value, perhaps it affects the math and lets you have more than 200 HPs. Although I it's related to Vice City and partially to SA and LCS/VCS, not GTA-III.

vflyson: The re-entry does not give armour if none taken previously for Enforcers as well.
It's a different story for taxi cabs and police cruisers, just because there's no limit set on shotgun ammo or money.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -239,6 +239,28 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
         } else if (character->animator->getAnimation(AnimIndexAction) ==
                    anm_enter) {
             if (character->animator->isCompleted(AnimIndexAction)) {
+                if (vehicle->getVehicle()->vehiclename_ == "TAXI"
+                    || vehicle->getVehicle()->vehiclename_ == "CABBIE"
+                    || vehicle->getVehicle()->vehiclename_ == "BORGNINE") {
```
danhedron: Could this be done in ``VehicleObject::setOccupant`` or ``CharacterObject::enterVehicle``? Those seem like more a better fit than here.

rwengine/src/ai/CharacterController.cpp
danhedron: The ``GameState::playerObject`` value should match ``getGameObjectID()`` called on the player's character, we can just wrap that in a ``CharacterObject::isPlayer()`` function if the bonuses should only apply to the player.

rwengine/src/ai/CharacterController.cpp
vflyson: Correct, this could be done in either of those as well, however comparing to the original behavior, both of them would result in character receiving the perk too early.
It goes like this:
```
I [World] Creating Vehicle ID 119 (BANSHEE)
CharacterObject.cpp:420
CharacterObject.cpp:420
VehicleObject.cpp:519
CharacterController.cpp:242
```
The :242 line (where the code currently resides) is being called once the player's inside a vehicle, not once the process of entering a vehicle has been initiated. I guess it should be fairly easy to move it, and I could move it right now wherever you feel like it fits but it won't be as close to what happens in vanilla as the current placement.

rwengine/src/ai/CharacterController.cpp
vflyson: I might be doing something wrong but I keep getting `error: invalid use of incomplete type ‘class GameState` both in isPlayer() (As well as when I tried implement the taxi cab money perk).
Currently just have `return true` within isPlayer() with a `/// @todo` next to it.

danhedron:
LGTM. We'll need an issue for non-players receiving these items.

vflyson:
Opened #281

Issue #214, opened 2016-08-22
Cars in tunnels float up


JayFoxRox:
I've been able to reproduce an issue which has been mentioned before (but not documented here?!):

![Up, up and away.. kind of](http://i.imgur.com/A51hQAp.png)

If you spawn vehicles in a tunnel (but not all of the tunnels?) they float up (because they are underwater probably). Interestingly, this does not happen to cars which you drive into the tunnel (Banshee on the right).

*When I tried to reproduce this last time, I only got some vehicles to float up, but I couldn't repeat it elsewhere and was unable to drive to those points as I forgot where I spawned the floating cars.
I drove through the tunnels searching for them but had no luck floating up or finding my cars (and I've tried various spots underwater and under land!)*

Lihis:
The tunnel which entrance is near Stadium in Staunton Island:

AI driven cars, which spawn from this point, until the intersection where you can go either Portland Island or Shoreside Vale will float up:
![floatup](https://user-images.githubusercontent.com/3246197/45925420-64bf1000-bf1d-11e8-80e5-b9a755f5a5f4.jpeg)

Also happening at parts of this tunnel which lead to airport. AI driven cars seems to only float up only if they spawned at affected area, if they drove to affected area they don't float up.

PR #283, opened 2017-04-16
refactoring and clean up

aldoxtor:


rwengine/src/objects/CharacterObject.cpp
```diff
@@ -480,10 +480,7 @@ float CharacterObject::getJumpSpeed() const {
 }
 
 bool CharacterObject::isOnGround() const {
-    if (physCharacter) {
-        return physCharacter->onGround();
-    }
-    return true;
+    return physCharacter == nullptr || physCharacter->onGround();
```
JayFoxRox: This is less readable

rwengine/src/engine/Animator.cpp
```diff
@@ -51,9 +51,6 @@ void Animator::tick(float dt) {
             BoneTransform xform;
             if (b.first->type == AnimationBone::R00) {
                 xform.rotation = kf.rotation;
-            } else if (b.first->type == AnimationBone::RT0) {
-                xform.rotation = kf.rotation;
-                xform.translation = kf.position;
```
JayFoxRox: Why was this removed? This is probably a supported version the fallback below might just be a fallback / non-defined behaviour which is still handled

rwengine/src/engine/Animator.cpp
```diff
@@ -89,10 +86,8 @@ void Animator::tick(float dt) {
 
 bool Animator::isCompleted(unsigned int slot) const {
     if (slot < animations.size()) {
-        return animations[slot].animation
-                   ? animations[slot].time >=
-                         animations[slot].animation->duration
-                   : true;
+        return animations[slot].animation == nullptr || animations[slot].time >=
+                                                        animations[slot].animation->duration;
```
JayFoxRox: This is less readable imo. Should at least contain braces or made into variables.
I think an early out would also be nice:

```
if (animations[slot].animation == nullptr) {
  return true;
}
return animations[slot].time >= animations[slot].animation->duration;
```

rwengine/src/engine/GameWorld.cpp
```diff
@@ -737,10 +737,7 @@ void GameWorld::clearCutscene() {
 bool GameWorld::isCutsceneDone() {
     if (state->currentCutscene) {
         float time = getGameTime() - state->cutsceneStartTime;
-        if (state->skipCutscene) {
-            return true;
-        }
-        return time > state->currentCutscene->tracks.duration;
+        return state->skipCutscene ? true : time > state->currentCutscene->tracks.duration;
```
JayFoxRox: Old code was more readable.
We are not trying to win a "shortest code" competition. More explicit is probably better

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -497,7 +494,7 @@ void CharacterObject::setJumpSpeed(float speed) {
 
 void CharacterObject::resetToAINode() {
     auto nodes = engine->aigraph.nodes;
-    bool vehicleNode = !!getCurrentVehicle();
+    bool vehicleNode = getCurrentVehicle() != nullptr;
```
JayFoxRox: imo such comparisons in assignments should always be in braces. What does our style guide say @danhedron ?

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -351,7 +351,7 @@ void VehicleObject::tickPhysics(float dt) {
                     if (_lastHeight >= wH) {
                         // we are for real, underwater
                         inWater = true;
-                    } else if (inWater == false) {
+                    } else if (inWater == false) { // This condition is unnecessary
```
JayFoxRox: True. Remove it altogether or refactor this code branch?

rwgame/states/BenchmarkState.cpp
```diff
@@ -4,7 +4,7 @@
 #include "RWGame.hpp"
 
 BenchmarkState::BenchmarkState(RWGame* game, const std::string& benchfile)
-    : State(game), benchfile(benchfile), benchmarkTime(0.f), frameCounter(0) {
+    : State(game), benchfile(benchfile), benchmarkTime(0.f), duration(0.f), frameCounter(0) {
```
JayFoxRox: imo we should move more of these constants into the member declaration.
But this is probably fine too.

rwviewer/ViewerWidget.cpp
```diff
@@ -163,7 +165,11 @@ void ViewerWidget::drawFrameWidget(ModelFrame* f, const glm::mat4& m) {
         glLineWidth(1.f);
     }
     dp.textures = {whiteTex};
-    renderer->getRenderer()->drawArrays(thisM, _frameWidgetDraw, dp);
+    
+    RW_CHECK(renderer != nullptr, "GameRenderer is null");
+    if(renderer != nullptr) {
+        renderer->getRenderer()->drawArrays(thisM, _frameWidgetDraw, dp);
+    }
```
JayFoxRox: This should probably just error if the GameRenderer is null.

(I think the viewer tools is dead anyway :P )

JayFoxRox:
Please add a PR description. I also think this should not be 3 seperate commits unless there is a pattern / structure in the commit order I'm not seeing.

rwengine/src/engine/Animator.cpp
aldoxtor: I removed 'else if' because it is doing the same thing like 'else'.

rwengine/src/objects/VehicleObject.cpp
aldoxtor: this condition should doing something or it is useless?

rwviewer/ViewerWidget.cpp
aldoxtor: if something is dead it shouldn't be in repo.

rwengine/src/objects/VehicleObject.cpp
JayFoxRox: The code itself is pretty broken and does not what the comment suggests. The check for line 350 is seemingly never done. We also had issue #214 which is probably affected by this.
So it probably needs a full refactor.
However, seeing how this is mostly a small cleanup, I'd suggest linking this "weird" code from #214 and not worrying about it now (e.g. remove this part from the PR).

aldoxtor:
I correct this PR.

rwviewer/ViewerWidget.cpp
JayFoxRox: agreed, but that's a seperate issue.
(We are currently transitioning renderers, first internally and then to librw by aap, that broke a ton of stuff)

JayFoxRox:
Sorry for taking so long with the review.

LGTM

rwviewer/ViewerWidget.cpp
danhedron: The viewer should still be working.

danhedron:
Thanks for tidying up, LGTM.

PR #286, opened 2017-04-24
Visual Studio 2017

madebr:
Hey,

These patches allow me to build OpenRW using Visual Studio 2017 on Windows 10.

Description:
- ~~cmake: Add LINK_STATIC switch: openal-soft and boost need special defines to know that library is linked as a static library. This eases distribution and testing as no extra dll is needed.~~
- Fix a bunch of compile errors


CMakeLists.txt
```diff
@@ -112,24 +112,25 @@ add_subdirectory(rwlib)
...
+            DESTINATION "${DOC_DIR}"
+    )
 endif()
+
 # And copy to build directory for CI
 file(COPY COPYING
-	 DESTINATION .
-	 )
+     DESTINATION .
+     )
```
JayFoxRox: this still has some extra spaces

.gitignore
```diff
@@ -4,4 +4,6 @@ documentation/
 *~
 *swp
 *.user*
+.vs
+*.bak
```
JayFoxRox: Fair enough for VS, but in general users should configure this in their global local git config. We can't add settings for every IDE in existence to a project

rwengine/src/engine/GameData.cpp
```diff
@@ -446,6 +446,7 @@ void GameData::loadModel(ModelID model) {
             slotname = name;
             break;
         case ModelDataType::PedInfo:
+        {
```
JayFoxRox: "How did this ever work?" - normally my clang complains about such things.
Also I'd suggest moving the opening brace on the previous line as we do for `if` and functions. But it's ultimately @tsjost and @danhedron call

cmake/modules/FindGLM.cmake
```diff
@@ -56,10 +56,6 @@ INCLUDE(FindPackageHandleStandardArgs)
...
 
-IF(GLM_FOUND)
-    SET(GLM_INCLUDE_DIRS "${GLM_INCLUDE_DIR}")
+SET(GLM_INCLUDE_DIRS "${GLM_INCLUDE_DIR}")
 
-    IF(NOT GLM_FIND_QUIETLY)
-        MESSAGE(STATUS "GLM_INCLUDE_DIR = ${GLM_INCLUDE_DIR}")
-    ENDIF(NOT GLM_FIND_QUIETLY)
-ENDIF(GLM_FOUND)
+mark_as_advanced(GLM_INCLUDE_DIR)
```
JayFoxRox: Did you write this patch yourself or did you look for an updated FindGLM?

cmake/modules/FindLibSndFile.cmake
```diff
@@ -19,16 +19,14 @@ find_path(LIBSNDFILE_INCLUDE_DIR
...
-find_package_handle_standard_args(LibSndFile  DEFAULT_MSG  LIBSNDFILE_LIBRARY LIBSNDFILE_INCLUDE_DIR)
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(LibSndFile DEFAULT_MSG LIBSNDFILE_LIBRARY LIBSNDFILE_INCLUDE_DIR)
 
-if(LIBSNDFILE_FOUND)
-  set(LIBSNDFILE_LIBRARIES ${LIBSNDFILE_LIBRARY})
-  set(LIBSNDFILE_INCLUDE_DIRS ${LIBSNDFILE_INCLUDE_DIR})
-endif(LIBSNDFILE_FOUND)
+set(LIBSNDFILE_LIBRARIES ${LIBSNDFILE_LIBRARY})
+set(LIBSNDFILE_INCLUDE_DIRS ${LIBSNDFILE_INCLUDE_DIR})
```
JayFoxRox: Why unconditionally? imo this should still depend on the library being found - that's kind of cmake standard / expected behaviour?

tests/test_globals.hpp
```diff
@@ -31,10 +31,11 @@ struct print_log_value<glm::vec3> {
         s << glm::to_string(v);
     }
 };
+BOOST_NS_MAGIC_CLOSING
 }
 }
-BOOST_NS_MAGIC_CLOSING
 
+#if BOOST_VERSION < 106400
```
JayFoxRox: We should add an issue about this after merge so we can remove such checks in a couple of years when all distros upgraded boost.

rwengine/src/audio/MADStream.hpp
```diff
@@ -10,7 +12,13 @@
 #include <stdio.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#if defined(HAVE_UNISTD_H)
 #include <unistd.h>
+#elif defined(HAVE_IO_H)
+#include <io.h>
+#else
+#error "Need an alternative for unistd.h and io.h"
+#endif
```
JayFoxRox: imo these 2 should be implied for unix like platforms and seperate checks / implementations should be added for windows imo

tests/CMakeLists.txt
```diff
@@ -53,10 +53,14 @@ set(TEST_SOURCES
...
-ADD_DEFINITIONS(-DBOOST_TEST_DYN_LINK)
-
 add_executable(run_tests ${TEST_SOURCES})
 
+if(NOT LINK_STATIC)
+	target_compile_definitions(run_tests
+		PUBLIC
+			BOOST_TEST_DYN_LINK)
+endif()
+
```
JayFoxRox: I feel like this should be a seperate PR.

I feel we should discuss what the standard cmake way of enforcing this is, I'd assume there are ways of doing this?

CMakeLists.txt
```diff
@@ -7,10 +7,12 @@ project(OpenRW)
 #
 
 # Global Build Configuration
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DRW_DEBUG=1")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall -pthread -Wextra -Wpedantic")
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DRW_DEBUG=1")
+if(NOT MSVC)
```
JayFoxRox: this check should be the other way around imo. Add compiler specific flags, not avoiding incompatible flags

CMakeLists.txt
```diff
@@ -108,14 +108,6 @@ find_package(GLM REQUIRED)
 find_package(LibSndFile REQUIRED)
 find_package(SDL2 REQUIRED)
 
-include_directories(
-    ${SDL2_INCLUDE_DIR}
-    ${GLM_INCLUDE_DIRS}
-    ${OPENGL_INCLUDE_DIR}
-    ${BULLET_INCLUDE_DIRS}
-    ${OPENAL_INCLUDE_DIR}
-)
```
JayFoxRox: This commit also feels like more build changes which are not related to VS2017, they should probably be a seperate PR

cmake/modules/FindGLM.cmake
madebr: I wrote it myself.

JayFoxRox: I suggest using a file you find elsewhere for this. GLM is a pretty common lib. CMake depends on standardization. Often cmake files are promoted to being included in newer CMake versions.

Also note that GLM used to come with it's own cmake file apparently (which we should probably use as it's the most likely to be included in CMake upstream): https://chromium.googlesource.com/external/github.com/g-truc/glm/+/0.9.0/FindGLM.cmake (not sure if there are more recent versions or if/why it was removed)

tests/CMakeLists.txt
madebr: I will change it to this: https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html

.gitignore
madebr: Changed it to only ignore bak files. Those are artifacts of WinMerge.

tests/CMakeLists.txt
madebr: On second thought, I believe the current behavior is fine. The naming might be off. BUILD_SHARED_LIBS changes the behavior of add_library, which is not wanted.

rwengine/src/engine/GameData.cpp
madebr: Moved the brace.
I haven't checked but maybe it 'worked' because the variable is static?

cmake/modules/FindGLM.cmake
madebr: Retested and the patches to findLibSndFile and findGLM are not needed.
So left them out.

tests/test_globals.hpp
madebr: I'm not surewhich version added the nullptr_t specialization. It might be 1.63.0 (or lower).

cmake/modules/FindLibSndFile.cmake
madebr: I kicked the FindLibSndFile.cmake patch out of my patch set.

CMakeLists.txt
madebr: Ok, will do another PR for that.

tests/test_globals.hpp
JayFoxRox: Please check and fix this check or mark this comment as a todo. I'd go as far and call this a bug otherwise ;)

CMakeLists.txt
madebr: Changed it

rwengine/src/script/ScriptModule.hpp
```diff
@@ -175,9 +175,11 @@ class ScriptModule {
     std::map<ScriptFunctionID, ScriptFunctionMeta> functions;
 };
 
+#if !defined(_MSC_VER)
 template <>
 void ScriptModule::bind<void(const ScriptArguments&)>(
     ScriptFunctionID id, int args, void (*function)(const ScriptArguments&));
+#endif
```
JayFoxRox: If this is not necessary for MSVC, why is this necessary in the first place?
Unless there's a compiler bug / extension I'd assume such things to be the same across compilers with the same C++ std.

madebr: It can also be commented out on gcc 6.3.1 and clang 3.8.1.
Since we are using -std=c++14, I can assume all supported compilers do not need this specialization.

madebr: The specialization has now been removed.

tests/CMakeLists.txt
madebr: Removed this patch. Will open a new PR/issue about this when this gets merged.

tests/test_globals.hpp
madebr: Added a comment: specialization needed in 1.60.0, not needed in 1.64.0.

madebr: using [wandbox.org](https://wandbox.org), I've verified that the `nullptr_t` specialization is included in boost starting from version 1.64.0.
1.63.0 needs the specialization: https://wandbox.org/permlink/XL4uPZWetPGcMbT9
1.64.0 fails to compile: https://wandbox.org/permlink/F8qO7hUFbX3C4ITE

madebr:
Added AppVeyor Windows CI:
- I wrote the build logic in a python script `build_windows.py`. That way it is possible to run and develop it on my own system.
- The Windows dependencies are available on https://github.com/madebr/openrw-windows-dependencies. You might want to clone that repository to the openrw team github account and modify `WIN_DEP_URL` in `build_windows.py`.
- example output: https://ci.appveyor.com/project/madebr/openrw/build/1.0.72
- Artifacts are available: e.g. https://ci.appveyor.com/project/madebr/openrw/build/job/6uu613fx93nlju91/artifacts

@danhedron @JayFoxRox : _final_ review requested

madebr:
- Rebased on master

New features:
- Updated `FindFFmpeg.cmake` to work on Windows
- Including `Windows.h` in `defines.h` in PR rwengine/openrw#313 included lots of things. Avoid that.
- Do not push artifacts by default: this behavior can be overridden by including `[artifact:1]` in your commit message.

darkf:
- What's wrong with using SDLmain?
- It's not clear how you changed FindFFmpeg because you changed all of the formatting in the same commit...
- You seem to be using "Windows" to mean "Visual Studio"

madebr:
Thanks for the review:
- I added the call to `SDL_SetMainReady` to the tests first because the Boost framework does not need a main function as entry point. At the same time, I added the same logic to rwgame. I wanted the differences between the game and tests to be small.
- The FindFFmpeg commit _can_ be separated in multiple patches. I don't see the added value in dissecting this commit. The commit message contains the list of changes:
  * use FindPackageHandleStandardArgs package
  * add swresample on Windows (libavcodec has a dependency on swresample)
  * additionally link to secur32.lib and Ws2_32.lib
  * formatting changes..
- Sorry about the slip of the tongue about Visual Studio/Windows. This PR is about adding Visual Studio support. Do you refer to the naming of the build script `build_windows.py`?

madebr:
- first patches fix compilation errors on windows
- Simplified the CMake module overrides and fix a link error on windows.
- rwgame needs a main function, so define `SDL_MAIN_HANDLED` in `main.cpp`.
- I've unified the testing for windows and linux:
  * `cmake/ctest/build.ctest` configures, builds, tests, does coverage, ... and submits
  * `cmake/ctest/calculate_names.ctest` creates names for the build (for CDash)
  * The default configuration for a system can be found in `cmake/ctest/configure_{LINUX/WINDOWS].ctest`
  * For CI, some extra environmental variables must be set. see `.travis.yml` and `.appveyor.yml`
  * For the experimental script, run it as `ctest -S cmake/ctest/script_experimental.ctest -DRUN_COVERAGE=TRUE`

madebr:
Actually all patches (except the spelling fix :D ) are needed for building on Visual Studio.
- The simplification of the packages is needed as well, since CMake on Windows thought the glu32 dependency of OpenGL was an object instead of a library.
- The last patch is strictly not needed but is for AppVeyor.

What about [the Windows dependencies](https://github.com/madebr/openrw-windows-dependencies). Should it stay there?

cornytrace:
What is the progress on merging this?

madebr:
Updated the windows dependencies and this patch series:
- Result at https://ci.appveyor.com/project/madebr/openrw/build/1.0.376

Things left to do or to discuss:
- where to put [openrw-windows-dependencies](https://github.com/madebr/openrw-windows-dependencies)?
- create a nice how-to + script for windows users to fetch the dependencies and start developing
- create an account op appveyor.com, set-up CI and add a badge to the readme

PR #287, opened 2017-04-29
fixing magic numbers in playAnimation()

aldoxtor:
repleace magic numbers constant variables.


JayFoxRox:
This is good.

@danhedron any reason why we couldn't use an enum class?

PR #288, opened 2017-04-29
properly skip activity

aldoxtor:
when activity is skipping we must clear activity for character too.

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -40,8 +40,10 @@ void CharacterController::skipActivity() {
     // vehicle
     // or jumping.
     if (getCurrentActivity() != nullptr &&
-        getCurrentActivity()->canSkip(character, this))
+        getCurrentActivity()->canSkip(character, this)) {
+        character->activityFinished();
         setActivity(nullptr);
+        }
 }
 
 void CharacterController::setNextActivity(
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -43,7 +43,7 @@ void CharacterController::skipActivity() {
         getCurrentActivity()->canSkip(character, this)) {
         character->activityFinished();
         setActivity(nullptr);
-        }
+    }
 }
 
 void CharacterController::setNextActivity(


aldoxtor:
I dont write this before, but this PR fix bug with animations. When you are entering vehicle and in certaint moment interupt this activity animation player is frozen.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -40,8 +40,10 @@ void CharacterController::skipActivity() {
     // vehicle
     // or jumping.
     if (getCurrentActivity() != nullptr &&
-        getCurrentActivity()->canSkip(character, this))
+        getCurrentActivity()->canSkip(character, this)) {
+        character->activityFinished();
```
JayFoxRox: I'm not sure about the purpose of `activityFinished`.
@danhedron Does it mean: "the activity has finished successfully" or is it more of a [cleanup measure as the comment in the header suggests](https://github.com/rwengine/openrw/blob/79bdd78952d4fdd0071599225ee1807cf9cf4059/rwengine/src/objects/CharacterObject.hpp#L190)

In case it's the later, maybe it should automaticly happen in the destructor or `setActivity` ?

rwengine/src/ai/CharacterController.cpp
```diff
@@ -40,8 +40,10 @@ void CharacterController::skipActivity() {
     // vehicle
     // or jumping.
     if (getCurrentActivity() != nullptr &&
-        getCurrentActivity()->canSkip(character, this))
+        getCurrentActivity()->canSkip(character, this)) {
```
JayFoxRox: I know this is not your code addition, but I think while you are at it, you should introduce a temporary variable to avoid checking a different currentActivity than what's actually used in the canSkip().
We never know what kind of code we'll add, but I think with network play or something in the future (callbacks etc.), we probably want this safer and it's an easy addition :)

darkf:
Any progress on this?

rwengine/src/ai/CharacterController.cpp
ShFil119: It looks like activityFinished flushes actual animation. I think it should be at least renamed.

darkf:
Seems like @aldoxtor is AWOL, so if someone like @ShFil119 or @husho (or anyone, really) does not want to pick this up or approve it then I'm afraid it would be best to close it after a while.

husho:
https://github.com/rwengine/openrw/issues/341#issuecomment-363880727

PR #289, opened 2017-05-02
implement isPlayer()

aldoxtor:
Add implementation isPlayer function in CharacterObject.

### DIFF
--- a/rwengine/src/engine/GameWorld.hpp
+++ b/rwengine/src/engine/GameWorld.hpp
@@ -6,11 +6,11 @@ class PlayerController;
 class Logger;
 
 class GameData;
-class GameState;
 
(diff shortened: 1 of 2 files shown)


JayFoxRox:
I'm a little worried we'll forget about this soon after. Can you add at least one commit where you actually use it? I remember having a handful of discussions where it would be useful, and I'd imagine we currently have weird workarounds for such checks.

aldoxtor:
I used `isPlayer()` for debugging. The only occurance this function is in this method:
`void VehicleObject::grantOccupantRewards(CharacterObject* character)`. I think that implementation is better than `return true`.

JayFoxRox:
Oh I didn't even realize we already had a signature / dummy implementation for it

PR #291, opened 2017-05-03
add InVehicle activity

aldoxtor:
Add InVehicle activity and move some code there.

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -15,16 +15,15 @@ CharacterController::CharacterController(CharacterObject *character)
     : character(character)
     , _currentActivity(nullptr)
     , _nextActivity(nullptr)
-    , m_closeDoorTimer(0.f)
     , currentGoal(None)
     , leader(nullptr)
     , targetNode(nullptr) {
     character->controller = this;
 }
 
-bool CharacterController::updateActivity() {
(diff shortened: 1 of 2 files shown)


rwengine/src/ai/CharacterController.cpp
```diff
@@ -292,8 +266,44 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
...
+
+        if (std::abs(d.x) > 0.01f) {
+            character->getCurrentVehicle()->setHandbraking(false);
+        }
+        character->getCurrentVehicle()->setThrottle(d.x);
+    }
+
+
+    // If character is idle in vehicle, try to close the door.
+    auto v = character->getCurrentVehicle();
```
paulsapps: dunno what these return but shouldn't they be auto&'s to prevent creating copies?

darkf: It returns a pointer

darkf:
Needs a rebase

darkf:
As I said in #288, this applies here too:

Seems like @aldoxtor is AWOL, so if someone like @ShFil119 or @husho (or anyone, really) does not want to pick this up or approve it then I'm afraid it would be best to close it after a while.

husho:
https://github.com/rwengine/openrw/issues/341#issuecomment-363880727

PR #293, opened 2017-06-17
Micro-optimizations

ShFil119:
Reducing four instructions to one.

### DIFF
--- a/rwengine/src/engine/GameWorld.cpp
+++ b/rwengine/src/engine/GameWorld.cpp
@@ -49,7 +49,7 @@ public:
         bool valA = a && a->type() == GameObject::Instance;
         bool valB = b && b->type() == GameObject::Instance;
 
-        if (!(valA && valB) && (valB || valA)) {
+        if (valA ^ valB) {
             // Figure out which is the dynamic instance.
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I fail to see how this is an "optimization".
You do realize that an optimizing compiler will compile this into a single CPU instruction anyway probably?

Code should be readable by humans - not computers. That's what the compiler does for you.

The old code was already pretty bad, but the new one is even worse in my opinion.
It's less descriptive and it makes it easy to cause bugs in the future.
You've replaced a logical operation with a bitwise operation.

If you want to check for inequality you should use `if (valA != valB) {`.

Overall this change feels like bikeshedding. There are bigger issues to tackle (script functions, CI, review of existing PRs, better documentation, support for more gameplay features etc.).

ShFil119:
I wanted to make it more clear to read, but overall "valA != valB" is better.

tsjost:
To make this more clear to read, val{A,B} should be renamed to {a,b}IsInstance, the condition put into a variable like `bool exactly_one_gameobject_is_instance = aIsInstance != bIsInstance;` and then this used in the if statement. It's still not clear why only one gameobject must be an instance though.

PR #294, opened 2017-06-18
Higher precison of convertion

ShFil119:
Changed 3,14/180 to constant value with higher precision.


JayFoxRox:
Please do not close PRs without reason and use more descriptive titles in the future (higher precision conversion of.. what? everything? physics? all over the codebase? in one location? ..).

ShFil119:
Sorry.

PR #295, opened 2017-06-18
Increase precision

ShFil119:
Update 3,141/180 to constant var with higher precision.


JayFoxRox:
Sounds like bikeshedding. The old code was also more readable as it at least hinted the relation of PI and 180. So if you introduce need magic constants, I'd suggest adding a comment where it comes from.

Also, if-need-be: https://glm.g-truc.net/0.9.4/api/a00136.html#ga4fb76e28851c9ff6653532566084e091
(which is already documented elsewhere + we use glm anyway)

ShFil119:
Yes, glm sounds more reasonable.

JayFoxRox:
Please close unused PRs yourself next time.

PR #296, opened 2017-06-18
Replace manual conversion with glm's function

ShFil119:
Glm allows to use angle manipulation functions. One of them is conversion degrees to radians.
It gives optimal precision and isn't strictly connected with float.
https://glm.g-truc.net/0.9.4/api/a00136.html#ga4fb76e28851c9ff6653532566084e091
// For mainteiners:
Please close https://github.com/rwengine/openrw/pull/295
New pull request prevents trash in history, and it's easier to read. Thx.


tsjost:
Thanks!

PR #298, opened 2017-07-28
Two more glm replaces and small clean up.

ShFil119:
I've found two more. ;)

### DIFF
--- a/rwengine/src/objects/GameObject.cpp
+++ b/rwengine/src/objects/GameObject.cpp
@@ -23,7 +23,8 @@ void GameObject::setRotation(const glm::quat& orientation) {
 }
 
 float GameObject::getHeading() const {
-    return glm::degrees(glm::roll(getRotation()));
+    auto headingInRadians = glm::roll(getRotation());
+    return glm::degrees(headingInRadians);
 }
 
 void GameObject::setHeading(float heading) {
--- a/rwengine/src/objects/GameObject.cpp
+++ b/rwengine/src/objects/GameObject.cpp
@@ -23,12 +23,10 @@ void GameObject::setRotation(const glm::quat& orientation) {
 }
 
 float GameObject::getHeading() const {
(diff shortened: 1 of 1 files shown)


rwengine/src/objects/GameObject.cpp
```diff
@@ -23,12 +23,10 @@ void GameObject::setRotation(const glm::quat& orientation) {
 }
 
 float GameObject::getHeading() const {
-    auto hdg = glm::roll(getRotation());
-    return hdg / glm::pi<float>() * 180.f;
+    return glm::degrees(glm::roll(getRotation()));
```
JayFoxRox: I think having a "hdg" (maybe rename to: `heading`) variable makes sense as it shows the intention / documents what goes on (extract heading -> convert to degrees). Without it, I'm somewhat lost reading that line.

Other instance in file too.

The use of `glm::degrees` looks better though :)

rwengine/src/objects/GameObject.cpp
```diff
@@ -23,12 +23,11 @@ void GameObject::setRotation(const glm::quat& orientation) {
 }
 
 float GameObject::getHeading() const {
-    auto hdg = glm::roll(getRotation());
-    return hdg / glm::pi<float>() * 180.f;
+    auto headingInRadians = glm::roll(getRotation());
+    return glm::degrees(headingInRadians);
```
JayFoxRox: nit: the idea was also to keep `setHeading` and `getHeading` in sync. If both use the same variable name for the same content, it's easier to guess what goes on. anyhow, I'm fine with the current solution, maybe just something to keep in mind for next time :)

JayFoxRox:
Please always provide a reason *why* you close your PRs.
This makes it easier for maintainers to improve the review process / create an overall better atmosphere.

ShFil119:
Sorry, I forgot. :(
Current solution is good enough, so I didn't see the sense of holding "trash pull".

PR #301, opened 2017-08-07
renderer as shared ptr

ShFil119:
Actually render is causing memory leak.
Another option is to delete in destructor.

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -68,7 +68,7 @@ DrawBuffer ssRectDraw;
 GameRenderer::GameRenderer(Logger* log, GameData* _data)
     : data(_data)
     , logger(log)
-    , renderer(new OpenGLRenderer)
+    , renderer(std::make_shared<OpenGLRenderer>())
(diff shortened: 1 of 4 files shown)


danhedron:
The convention is to avoid manual allocation and deletion where possible, so this is the right track.

Does it need to use shared ptr? GameRenderer should be able to own the renderer outright with unique ptr.

ShFil119:
Yes, pointer is also being kept in MapRenderer.

paulsapps:
can't see why the ownership of the renderer is shared..

ShFil119:
Ok, you probably thinking about unique passed with reference?
I'll prepare new commit and pull request. ;)

JayFoxRox:
You can change commits and re-use this branch.. (instead of closing it)
Please avoid creating a new PR for changes as this discussion provides important background knowledge: Why did we choose unique_ptr over shared_ptr?

To make this easier, try to use more generic PR names such as "Renderer as smart-pointer" next time.

*(Opening a closed PR is tricky, so next time, just avoid closing it at all, unless you don't want to work on a new revision at all anymore)*

Issue #239, opened 2016-09-11
Make more liberal use of smart pointers


darkf:
The codebase is quite spaghetti at some points and there are countless memory leaks just from allocating pointers and never freeing them (most don't even have destructors :|)

`std::unique_ptr` and `std::shared_ptr` should be sprinkled pretty heavily around.

danhedron:
In rwgame there are a few things in particular that need to have their lifetimes managed:

- State instances
- some members of RWGame should either be made uniqure_ptrs or just member variables

If you see anything else that's dubious (i.e. a raw new or delete) please leave a note so it's possible to track what needs fixing.

danhedron:
Some more thoughts:
- Activities
   - These should just be stored as a union/variant anyway, doesn't need dynamic allocation
- Bullet shapes / Collision Model
   - I will be changing collision stuff soon so I'll take care of these pointers
- SCMFile
    - Probably doesn't need to be dynamically allocated, RVO and move semantics should be enough. Should use unique_ptr internally
- IDE objects
    - Affected by changes in [upcoming PR](https://github.com/danhedron/openrw/pull/13).
- Animation / Skeletons
   - Highly likely to be affected or deleted with librw

JayFoxRox:
Was this handled in #244 ?

danhedron:
Partially. There's lots more to do (some incoming).

danhedron:
This issue is rather vague and more work has been done in this area.

PR #302, opened 2017-08-07
Improving handling game shaders

ShFil119:
Ok, I've edited pull request, because now happens much more. ;)

I've found some problems. First OpenGL programs have never been removed/closed. So I've added code doing it (in commit 75b188f), but now destructor have never been called, because memory allocated by OpenGLShaderPrograms are handled by raw pointers and never is freed. So I converted them unique pointers(commit 30a12ea).

I also improved the management of shaders in function compileProgram(commits 86d9122, 49f38e3, c858454). It should reduce memory usage.

Template I used: 
glCreateProgram
glCreateShader
glShaderSource
glCompileShader
glAttachShader
glLinkProgram <- after this line opengl got everything it needs, you can free shader resources
glDetachShader
glDeleteShader

for{//render loop
   glUseProgram
   //...drawing operations
   glUseProgram(0);
}

glDeleteProgram 

Removes this type of memory leak:
==21845== 1,075 bytes in 1 blocks are possibly lost in loss record 2,921 of 3,413
==21845==    at 0x4C2BE7F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21845==    by 0xDE31CA2: ??? (in /usr/lib/xorg/modules/dri/radeonsi_dri.so)
==21845==    by 0x577114: compileShader(unsigned int, char const*) (OpenGLRenderer.cpp:10)
==21845==    by 0x5772F6: compileProgram(char const*, char const*) (OpenGLRenderer.cpp:47)
==21845==    by 0x577BE0: OpenGLRenderer::createShader(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (OpenGLRenderer.cpp:220)
==21845==    by 0x57F7B0: WaterRenderer::WaterRenderer(GameRenderer*) (WaterRenderer.cpp:12)
==21845==    by 0x5611D2: GameRenderer::GameRenderer(Logger*, GameData*) (GameRenderer.cpp:77)
==21845==    by 0x46B9A4: RWGame::RWGame(Logger&, int, char**) (RWGame.cpp:35)
==21845==    by 0x45B0F9: main (main.cpp:13)
References:
#239

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -81,22 +81,22 @@ GameRenderer::GameRenderer(Logger* log, GameData* _data)
         renderer->createShader(GameShaders::WorldObject::VertexShader,
                                GameShaders::WorldObject::FragmentShader);
 
-    renderer->setUniformTexture(worldProg, "texture", 0);
-    renderer->setProgramBlockBinding(worldProg, "SceneData", 1);
-    renderer->setProgramBlockBinding(worldProg, "ObjectData", 2);
+    renderer->setUniformTexture(worldProg.get(), "texture", 0);
+    renderer->setProgramBlockBinding(worldProg.get(), "SceneData", 1);
+    renderer->setProgramBlockBinding(worldProg.get(), "ObjectData", 2);
 
     particleProg =
         renderer->createShader(GameShaders::WorldObject::VertexShader,
                                GameShaders::Particle::FragmentShader);
 
-    renderer->setUniformTexture(particleProg, "texture", 0);
-    renderer->setProgramBlockBinding(particleProg, "SceneData", 1);
-    renderer->setProgramBlockBinding(particleProg, "ObjectData", 2);
+    renderer->setUniformTexture(particleProg.get(), "texture", 0);
+    renderer->setProgramBlockBinding(particleProg.get(), "SceneData", 1);
+    renderer->setProgramBlockBinding(particleProg.get(), "ObjectData", 2);
 
     skyProg = renderer->createShader(GameShaders::Sky::VertexShader,
                                      GameShaders::Sky::FragmentShader);
 
-    renderer->setProgramBlockBinding(skyProg, "SceneData", 1);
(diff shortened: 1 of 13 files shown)


paulsapps:
Its a bigger change but I find its nice to create RAII wrappers for the oGL stuff.

rwengine/src/render/GameRenderer.cpp
```diff
@@ -212,6 +212,7 @@ GameRenderer::GameRenderer(Logger* log, GameData* _data)
 
 GameRenderer::~GameRenderer() {
     glDeleteFramebuffers(1, &framebufferName);
+    glDeleteProgram(ssRectProgram);
```
ShFil119: Or maybe should we convert ssRectProgram to  ShaderProgram?

rwengine/src/render/OpenGLRenderer.hpp
```diff
@@ -252,21 +252,21 @@ class OpenGLRenderer : public Renderer {
 
     std::string getIDString() const override;
 
-    ShaderProgram* createShader(const std::string& vert,
+    std::unique_ptr<ShaderProgram> createShader(const std::string& vert,
                                 const std::string& frag) override;
-    void setProgramBlockBinding(ShaderProgram* p, const std::string& name,
+    void setProgramBlockBinding(std::unique_ptr<ShaderProgram>& p, const std::string& name,
```
danhedron: I think it would make more sense for the ``setProgramBlockBinding``, ``setUniform`` and ``useProgram`` to take either a ``ShaderProgram*`` or a ``ShaderProgram&``, the interface shouldn't care if they're in a unique_ptr or not, and they shouldn't be null anyway.

rwengine/src/render/GameRenderer.cpp
danhedron: It should be doable. But isn't required for this PR.

PR #303, opened 2017-08-08
Renderer as smart-pointer

ShFil119:
As suggested in #301, I converted pointer to unique type.

### DIFF
--- a/rwengine/src/render/GameRenderer.hpp
+++ b/rwengine/src/render/GameRenderer.hpp
@@ -49,7 +49,7 @@ class GameRenderer {
     Logger* logger;
 
     /** The low-level drawing interface to use */
-    Renderer* renderer;
+    std::unique_ptr<Renderer> renderer;
 
     // Temporary variables used during rendering
     float _renderAlpha;
@@ -140,8 +140,9 @@ public:
     void setupRender();
     void renderPostProcess();
 
+    /** The allocated memory belongs to a unique pointer renderer in GameRenderer */
     Renderer* getRenderer() {
-        return renderer;
+        return renderer.get();
     }
 
     void setViewport(int w, int h);
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -68,11 +68,11 @@ DrawBuffer ssRectDraw;
 GameRenderer::GameRenderer(Logger* log, GameData* _data)
     : data(_data)
     , logger(log)
-    , renderer(new OpenGLRenderer)
+    , renderer(std::make_unique<OpenGLRenderer>())
     , _renderAlpha(0.f)
     , _renderWorld(nullptr)
     , cullOverride(false)
-    , map(renderer, _data)
+    , map(renderer.get(), _data)
     , water(this)
     , text(this) {
     logger->info("Renderer", renderer->getIDString());


vflyson:
Thank you for your contribution, @ShFil119!

I'd like to note that it would be much easier to manage the project and work with PRs if you could provide more definitive commit messages and PR titles.

There's an article that has once helped me understand the importance of being as verbose as possible while working with git, and I'd like to share it with you as I see your contributions being valuable, but the commit messages creating a bottleneck in reviewing and merging them in because it's a big project and it's not easy to maintain.

Here's the link to the article: https://chris.beams.io/posts/git-commit/

ShFil119:
Thanks, I really like programming, but I still have to learn more about team work. ;)

madebr:
Why use a `std::unique_ptr<Renderer>` inside GameRenderer when you're handing out raw pointers through the `getRenderer()` method?

I would suggest using a `std::shared_ptr<Renderer>` inside GameRenderer and hand out `std::weak_ptr<Renderer>`.

danhedron:
I'm happy with unique ptr. Nothing else should need to extended the lifetime.

paulsapps:
`getRenderer` could return a `Renderer&` which makes it clear you shouldn't delete it and that it can't be nullptr.

ShFil119:
Ok. It's good idea. How about returning const ref? It will prevent destroying.

madebr:
A const ref will disallow any non-const methods on the `Renderer` object, which is too strict atm.

JayFoxRox:
Bikeshed

rwengine/src/render/GameRenderer.hpp
```diff
@@ -140,8 +140,9 @@ class GameRenderer {
     void setupRender();
     void renderPostProcess();
 
-    Renderer* getRenderer() {
-        return renderer.get();
+    Renderer*& getRenderer() {
```
paulsapps: thats not quite right, should just be `Renderer& getRenderer() {`

And then anything that called it that was doing getRenderer->Something() should be changed to getRenderer.Something().

ShFil119: This pull request aims to prevent memory leak. 
Practically in the code are only pointers, so this refactor should implicate to change them.
And this is a lot of work to do, to much for this pull.

madebr: A reference to a pointer does not make sense here.
You're now basically returning a pointer to pointers in disguise.

paulsapps: It should just be a raw pointer if not returning as a reference otherwise this is very confusing and actually wrong as it allows the caller to change the pointer value.

danhedron: This should either return a reference or a pointer. If you don't want to update all calls to this function I would be happy with a comment on the function in the header that the pointer is owned by the GameRenderer.

ShFil119:
Should I "git reset --hard HEAD~1"?

madebr:
Hey @ShFil119 , I would suggest the approach in [this patch](https://github.com/madebr/openrw/commit/aee23f311ffc5ab578edc3e301b2e83047908306).

ShFil119:
Similar solution has been merged, using shared_ptr.
So this pull is unneeded.

PR #306, opened 2017-08-17
Replacing "new" operators with make_unique/shared

ShFil119:
I've found some snippets of code to edit in c++ 11 "style".
I've tried to keep readability as good as possible.

### DIFF
--- a/rwengine/src/BinaryStream.cpp
+++ b/rwengine/src/BinaryStream.cpp
@@ -20,7 +20,7 @@ std::unique_ptr<BinaryStream> BinaryStream::parse(const std::string &filename) {
     dfile.read(data, length);
     // std::cout << "File is " << length << " bytes" << std::endl << std::endl;
 
-    auto BS = std::unique_ptr<BinaryStream>(new BinaryStream);
+    auto BS = std::make_unique<BinaryStream>();
 
     // Set file's ACTUAL length
     auto header = reinterpret_cast<nativeSectionHeader_t *>(data);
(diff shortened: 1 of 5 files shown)


rwengine/src/BinaryStream.cpp
```diff
@@ -20,7 +20,7 @@ std::unique_ptr<BinaryStream> BinaryStream::parse(const std::string &filename) {
     dfile.read(data, length);
     // std::cout << "File is " << length << " bytes" << std::endl << std::endl;
 
-    auto BS = std::unique_ptr<BinaryStream>(new BinaryStream);
+    std::unique_ptr<BinaryStream> BS = std::make_unique<BinaryStream>();
```
paulsapps: might as well be `auto BS = std::make_unique<BinaryStream>();`

paulsapps: Also there is still `char *data = new char[length];` in here which could be vector<char>

rwengine/src/loaders/LoaderIDE.cpp
```diff
@@ -65,8 +65,7 @@ bool LoaderIDE::load(const std::string &filename, const PedStatsList &stats) {
                     break;
                 case OBJS:
                 case TOBJ: {  // Supports Type 1, 2 and 3
-                    auto objs =
-                        std::unique_ptr<SimpleModelInfo>(new SimpleModelInfo);
+                    std::unique_ptr<SimpleModelInfo> objs = std::make_unique<SimpleModelInfo>();
```
paulsapps: In this switch case nothing happens with the result obj... can't see why it needs to be on the heap?

ShFil119:
Do you prefer new pull, or rewrite commits?

danhedron:
Preferably rewrite commits unless the entire PR changes

ShFil119:
Done. ;)
"Have you run clang-format on the files? Some lines look like they're a bit longer than before."
No, but can you give a detail what/how change?

Issue #129, opened 2016-06-04
Text rendering is not accurate


JayFoxRox:
See #121 for more infos

(I'll eventually move the details from said post to this issue. I'll do it later though)

vflyson:
the dropbox link is dead Ö

JayFoxRox:
> Error (429)
This account's links are generating too much traffic and have been temporarily disabled!

It's still there but only accesible like once per week because dropbox sucks.
The same thing has been available on my website for a long time: http://www.jannikvogel.de/scratchpad/gta3-font/new-font.html

PR #307, opened 2017-08-22
Prevent occurrence text overlay

ShFil119:
Reedited, now it should handle overlay of HighPriority text. Description in comments of commits.
 
Ref:
#121 
#129


ShFil119:
Updated. ;) Or, do you prefer earlier version with push_back?

ShFil119:
After some testing. This type mechanism is in gta3, take a look at conversation with Luigi in mission "Drive Misty For Me".
What's missing is flushing after cutscenes and green text should also flushes conversation text.

danhedron:
LGTM.

PR #308, opened 2017-08-25
Various small changes

ShFil119:
Suggestions as usual are welcomed. ;)

This pull contains small improvements, but overall doesn't do anything big.


ShFil119:
Earlier:
xform.rotation = kf.rotation;
xform.translation = kf.position;
Now:
xform.rotation = kf.rotation;
xform.translation = kf.position;

I can be wrong, or something don't understand.
Can you show me the difference? Thx.

madebr:
You're right. Misread the diff. Ignore me 🙈 🌝

PR #312, opened 2017-09-08
Fix bugs in Windows build

madebr:
rwengine/openrw#286 only enables building OpenRW on Windows. The generated binary does not work.
This PR fixes some blatant bugs that will solve this problem.

The Windows performance is not good, but this allows implementing Windows integration like rwengine/openrw#276


danhedron:
Looks good to me. Thanks for separating these changes from the other windows changes.

PR #314, opened 2017-09-13
Plug some memory holes

madebr:
Hey,
Using Valgrind I plugged some memory holes.
These patches let OpenRW go from ([gist](https://gist.github.com/madebr/bfc3b31f97925e8f83422eb4bec4df06/1ae89d123c9d1c64650f3c54696d9ad1c69d08c2)):
```
==22737== LEAK SUMMARY:
==22737==    definitely lost: 561,692,681 bytes in 12,231 blocks
==22737==    indirectly lost: 70,419,415 bytes in 311,902 blocks
==22737==      possibly lost: 49,812 bytes in 148 blocks
==22737==    still reachable: 257,601 bytes in 1,283 blocks
==22737==         suppressed: 0 bytes in 0 blocks
```

to ([gist](https://gist.github.com/madebr/bfc3b31f97925e8f83422eb4bec4df06/1258c9e2f47d5036ed367ab145418b0807a08a60))

```
==26514== LEAK SUMMARY:
==26514==    definitely lost: 39,345 bytes in 18 blocks
==26514==    indirectly lost: 22,828 bytes in 65 blocks
==26514==      possibly lost: 800 bytes in 2 blocks
==26514==    still reachable: 225,491 bytes in 1,031 blocks
==26514==         suppressed: 0 bytes in 0 blocks
```

The remaining memory holes are in:
- the `SoundManager`, but that's being rewritten as we speak by @darkf
- createShader: which I'll leave to @ShFil119 (see #302 )
- memory allocated by SDL2/X/GL/... which is reclaimed by the OS at end of process.

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -29,9 +29,8 @@ bool CharacterController::updateActivity() {
     return false;
 }
 
-void CharacterController::setActivity(CharacterController::Activity *activity) {
-    if (_currentActivity) delete _currentActivity;
-    _currentActivity = activity;
+void CharacterController::setActivity(std::unique_ptr<Activity> activity) {
+    _currentActivity.swap(activity);
 }
 
 void CharacterController::skipActivity() {
@@ -43,14 +42,12 @@ void CharacterController::skipActivity() {
         setActivity(nullptr);
 }
 
-void CharacterController::setNextActivity(
-    CharacterController::Activity *activity) {
+void CharacterController::setNextActivity(std::unique_ptr<Activity> activity) {
     if (_currentActivity == nullptr) {
-        setActivity(activity);
+        setActivity(std::move(activity));
         _nextActivity = nullptr;
     } else {
-        if (_nextActivity) delete _nextActivity;
-        _nextActivity = activity;
+        _nextActivity.swap(activity);
     }
 }
 
@@ -97,13 +94,9 @@ void CharacterController::update(float dt) {
 
     if (updateActivity()) {
         character->activityFinished();
-        if (_currentActivity) {
-            delete _currentActivity;
-            _currentActivity = nullptr;
-        }
+        _currentActivity = nullptr;
         if (_nextActivity) {
-            setActivity(_nextActivity);
-            _nextActivity = nullptr;
+            setActivity(std::move(_nextActivity));
         }
     }
 }
@@ -282,7 +275,7 @@ bool Activities::EnterVehicle::update(CharacterObject *character,
             // out.
             character->playCycle(cycle_pullout);
             currentOccupant->controller->setNextActivity(
-                new Activities::ExitVehicle(true));
+                std::make_unique<Activities::ExitVehicle>(true));
         } else {
             character->playCycle(cycle_enter);
             character->enterVehicle(vehicle, seat);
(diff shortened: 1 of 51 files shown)


rwengine/src/ai/CharacterController.cpp
```diff
@@ -30,9 +29,8 @@ bool CharacterController::updateActivity() {
     return false;
 }
 
-void CharacterController::setActivity(CharacterController::Activity *activity) {
-    if (_currentActivity) delete _currentActivity;
-    _currentActivity = activity;
+void CharacterController::setActivity(std::unique_ptr<Activity> activity) {
+    _currentActivity.swap(activity);
```
paulsapps: why swap? Setting activity to _currentActivity seems redundant, could just be _currentActivity = std::move(activity);

tests/test_loaderdff.cpp
```diff
@@ -23,7 +23,7 @@ BOOST_AUTO_TEST_CASE(test_load_dff) {
         BOOST_REQUIRE(atomic->getGeometry());
         BOOST_REQUIRE(atomic->getFrame());
 
-        delete m;
+        m.reset();
```
paulsapps: redundant, will be reset/deleted because it goes out of scope here anyway

tests/test_animation.cpp
```diff
@@ -9,27 +9,30 @@ BOOST_AUTO_TEST_SUITE(AnimationTests)
...
-            1.0f,
-            AnimationBone::RT0,
-            {
-                {glm::quat(), glm::vec3(0.f, 0.f, 0.f), glm::vec3(), 0.f, 0},
-                {glm::quat(), glm::vec3(0.f, 1.f, 0.f), glm::vec3(), 1.0f, 1},
-            }};
-
-        animator.playAnimation(0, &animation, 1.f, false);
+        animation->duration = 1.f;
+        animation->bones["player"] = std::shared_ptr<AnimationBone>(
```
paulsapps: make_shared?

danhedron: Does this need to be shared? Is there a risk of bones being used out of scope if they are owned by the animation?

madebr: I made it a shared_pointer so it would be deleted automatically when the map would be destroyed (and to avoid adding a destructor).

I restored the plain pointer and added a destructor in the extra commit.

madebr:
I incorporated the comments in the extra commit. If you don't want that extra commit, please say so, and I'll merge it with the corresponding commit.

tests/test_animation.cpp
paulsapps: I think he means could it not be unique_ptr - i.e its not actually shared ownership?

madebr: The main reason I went for `std::shared_pointer<AnimationBone>` is because the [AnimationBone](https://github.com/rwengine/openrw/blob/11e90c61e56b60240251cd080f175ddff7d7a101/rwengine/src/loaders/LoaderIFP.hpp#L41) is referenced as key in [AnimationState](https://github.com/rwengine/openrw/blob/11e90c61e56b60240251cd080f175ddff7d7a101/rwengine/src/engine/Animator.hpp#L37).

rwengine/src/ai/CharacterController.cpp
ShFil119: Btw Isn't it better to use static union or newer c++ style variant?

PR #315, opened 2017-09-13
Implement FFmpeg audio decoder and remove dependencies on libMAD, libsndfile, and mman-win32 (on Windows)

darkf:
This PR rewrites the audio frontend to use libavcodec from FFmpeg instead of a combination of libsndfile and libMAD, which we unanimously wanted to get rid of.

It also deprecates streaming music in favor of just `Sound`, which for now works just fine (and we can add streaming in later, possibly in `Sound` instead of its own class.)

As a result of removing `MADStream`, we can also get rid of the Windows `mmap` polyfill `mman-win32`, which was an ugly hack.

### Caveats

I've only tested this on my machine (Windows 10 on x86_64, with MinGW/MSYS), I'd like others to test it on other platforms.

I only support planar signed 16-bit sample formats, which libavcodec hands to me. This makes it really easy to give to OpenAL.

I've only really tested the intro cutscenes, but they work fine and in the necessary format. I'd like someone to test and see if it breaks elsewhere (it will just not play the audio, and report an error to the console.)

In the future, if we want to support more formats, that is not hard to do so by converting via other FFmpeg APIs.

The wiki's build instructions will need to be updated for this post-merge, to drop the removed dependencies and add FFmpeg.


ShFil119:
On Arch Linux x64 it works well. ;)

danhedron:
Works on Fedora 24, but there's a deprecation warning:
```
SoundManager.cpp:97:49: warning: ‘AVStream::codec’ is deprecated [-Wdeprecated-declarations]
     AVCodecContext* codecContext = audioStream->codec;
                                                 ^~~~~
```
This is annoying, but the documentation is not clear on how this should be replaced.

Is there a way to suppress the output to stdout?
```
[mp3 @ 0xc29af40] Estimating duration from bitrate, this may be inaccurate
```

ShFil119:
It looks like AVCodecContext should be replaced with AVCodecParameters.
Examples:
https://github.com/mean00/avidemux2/commit/d2b1f44087b5774f00410db77fb9c560095939cc
https://code.videolan.org/libav/libav/commit/9200514ad8717c63f82101dc394f4378854325bf

darkf:
@danhedron Updated to use AVCodecParameters instead of `->codec` directly.

There is still a latent deprecation warning for `avcodec_decode_audio4` but I'm not going to bother with it because:

- It's not clear how to replace it, the docs say use `avcodec_receive_frame` but it's not a drop-in replacement (in particular, it has different semantics and it's not obvious how to get the length of the read packet.)

- The official audio decoding example still uses `avcodec_decode_audio4`, so obviously it's not a high priority for them.

If you want to figure that out, be my guest.

Note that because you wanted this to be updated it will no longer work on Travis because the Trusty version of FFmpeg is old, and it does not have these APIs. So, yeah, nothing I can do about that unless you feel like PPAing it (which I won't do.)

I also tuned the log level to errors on release mode, and warnings on debug+verbose mode.

ShFil119:
https://gist.github.com/ShFil119/cf1cfa4f8b4bbdeff8b2cd4da69255c2 
:)

danhedron:
Oh dear, well I'd rather live with the deprecation warnings than not have Travis. The API usage can be brought up to the new APIs at a later time.

The logging change is good.

darkf:
@ShFil119 Thanks. That is useful for the newer API and I'll keep it in mind, but currently we want to maintain API compatibility with older versions of FFmpeg (for systems like Travis which only have older versions.)
It's likely not going to go away any time soon.
Once Travis updates we can update to the newer API.

@danhedron Sounds good. I've dropped that change and split the rest out. Should be good now if nobody has found any issues.

ShFil119:
There are two more options. Build ffmpeg before openrw.
https://github.com/rg3/youtube-dl/issues/12366
Or with preprocessor hide newer version.

rwengine/src/audio/SoundManager.cpp
```diff
@@ -42,23 +54,124 @@ bool SoundManager::initializeOpenAL() {
...
+    AVFrame* frame = av_frame_alloc();
+    if (!frame) {
+        std::cerr << "Error allocating the audio frame" << std::endl;
+        return;
+    }
+
+    // Allocate formatting context
+    AVFormatContext* formatContext = nullptr;
+    if (avformat_open_input(&formatContext, filename.c_str(), nullptr, nullptr) != 0) {
+        av_frame_free(&frame);
```
danhedron: You can use unique ptr in this case to take advantage of RAII with a custom deleted.

PR #320, opened 2017-09-22
Replace static_cast and NULLs in renderer with their safer equivalents

TechnoCrunch:
Redundant c_str() were also removed and had no affect on the game during a test.

### DIFF
--- a/rwengine/src/render/ObjectRenderer.cpp
+++ b/rwengine/src/render/ObjectRenderer.cpp
@@ -32,7 +32,7 @@ RenderKey createKey(bool transparent, float normalizedDepth,
            uint32_t(0x7FFFFF *
                     (transparent ? 1.f - normalizedDepth : normalizedDepth))
                << 8 |
-           uint8_t(0xFF & (textures.size() > 0 ? textures[0] : 0)) << 0;
+           uint8_t(0xFF & (!textures.empty() ? textures[0] : 0)) << 0;
 }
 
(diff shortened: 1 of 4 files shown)


rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -73,7 +73,7 @@ void ObjectRenderer::renderGeometry(Geometry* geom,
                 dp.colour = mat.colour;
 
                 if (object && object->type() == GameObject::Vehicle) {
-                    auto vehicle = static_cast<VehicleObject*>(object);
+                    auto vehicle = dynamic_cast<VehicleObject*>(object);
```
paulsapps: This has run time cost, and it can fail so result should be asserted.

rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -349,24 +349,24 @@ void ObjectRenderer::buildRenderList(GameObject* object, RenderList& outList) {
     // Right now specialized on each object type
     switch (object->type()) {
         case GameObject::Instance:
-            renderInstance(static_cast<InstanceObject*>(object), outList);
+            renderInstance(dynamic_cast<InstanceObject*>(object), outList);
```
paulsapps: seems more like a design flaw, assuming there is a renderInstance for each type, then probably InstanceObject should have a renderInstance() which is passed an ObjectRenderer who can then call it without a cast.

PR #322, opened 2017-09-25
Implement math opcodes

ShFil119:
While searching opcodes connected with text, I found math opcodes which can be easily implemented.

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -8500,9 +8500,7 @@ void opcode_02f5(const ScriptArguments& args, const ScriptObject object, const S
 	@arg xOffset X offset
 */
 void opcode_02f6(const ScriptArguments& args, const ScriptFloat angle, ScriptFloat& xOffset) {
-	RW_UNIMPLEMENTED_OPCODE(0x02f6);
-	RW_UNUSED(angle);
-	RW_UNUSED(xOffset);
+	xOffset = glm::sin(glm::radians(angle));
 	RW_UNUSED(args);
 }
 
@@ -8514,9 +8512,7 @@ void opcode_02f6(const ScriptArguments& args, const ScriptFloat angle, ScriptFlo
 	@arg yOffset Y offset
 */
 void opcode_02f7(const ScriptArguments& args, const ScriptFloat angle, ScriptFloat& yOffset) {
-	RW_UNIMPLEMENTED_OPCODE(0x02f7);
-	RW_UNUSED(angle);
-	RW_UNUSED(yOffset);
+	yOffset = glm::cos(glm::radians(angle));
 	RW_UNUSED(args);
 }
 
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -12677,16 +12677,14 @@ bool opcode_0424(const ScriptArguments& args) {
 }
 
 /**
-	@brief convert_metres_to_feet %2d% store_to %1d%
+	@brief %2d% = meters_to_feet %1d% 
 
 	opcode 0425
 	@arg arg1 
 	@arg arg2 
 */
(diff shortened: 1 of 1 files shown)


rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -8500,9 +8480,7 @@ void opcode_02f5(const ScriptArguments& args, const ScriptObject object, const S
 	@arg xOffset X offset
 */
 void opcode_02f6(const ScriptArguments& args, const ScriptFloat angle, ScriptFloat& xOffset) {
-	RW_UNIMPLEMENTED_OPCODE(0x02f6);
-	RW_UNUSED(angle);
-	RW_UNUSED(xOffset);
+	xOffset = glm::sin(angle);
```
danhedron: The opcode argument is in degrees, but glm::sin takes radians. You can use glm::radians to convert the argument to rads.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -8514,9 +8492,7 @@ void opcode_02f6(const ScriptArguments& args, const ScriptFloat angle, ScriptFlo
 	@arg yOffset Y offset
 */
 void opcode_02f7(const ScriptArguments& args, const ScriptFloat angle, ScriptFloat& yOffset) {
-	RW_UNIMPLEMENTED_OPCODE(0x02f7);
-	RW_UNUSED(angle);
-	RW_UNUSED(yOffset);
+	yOffset = glm::cos(angle);
```
danhedron: Same issue as 02f6

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -12697,16 +12673,14 @@ bool opcode_0424(const ScriptArguments& args) {
...
 
 	opcode 0425
 	@arg arg1 
 	@arg arg2 
 */
 void opcode_0425(const ScriptArguments& args, const ScriptFloat arg1, ScriptFloat& arg2) {
-	RW_UNIMPLEMENTED_OPCODE(0x0425);
-	RW_UNUSED(arg1);
-	RW_UNUSED(arg2);
+	arg2 = arg1 * 3.2808398;
```
danhedron: This is actually too accurate, the behaviour is documented on https://www.gtamodding.com/wiki/0425

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -12793,16 +12767,14 @@ void opcode_042c(const ScriptArguments& args, const ScriptInt arg1) {
...
 
 	opcode 042d
 	@arg arg1 
 	@arg arg2 
 */
 void opcode_042d(const ScriptArguments& args, const ScriptInt arg1, ScriptInt& arg2) {
-	RW_UNIMPLEMENTED_OPCODE(0x042d);
-	RW_UNUSED(arg1);
-	RW_UNUSED(arg2);
+	arg2 = static_cast<ScriptInt>(std::round(static_cast<ScriptFloat>(arg1) * 3.2808398));
```
danhedron: Same as 0425, the actual behaviour is slightly less accurate: https://www.gtamodding.com/wiki/042D

rwengine/src/script/modules/GTA3ModuleImpl.inl
ShFil119: Multiplication with 3 will be enough?

rwengine/src/script/modules/GTA3ModuleImpl.inl
ShFil119: What do you think about floor(arg1 * 3.3333334)?

danhedron: That seems close to the expected behaviour

rwengine/src/script/modules/GTA3ModuleImpl.inl
danhedron: More like 3.333333, or division by 0.3

PR #323, opened 2017-10-01
Implement missing opcodes "Is in area?"

ShFil119:
Small pull, it's practically copy and paste.
Edit. I've added one more bigger commit. It's still calque, but I'm not sure about int arg6/8.
And how think about angled area.

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -4590,14 +4590,9 @@ void opcode_0196(const ScriptArguments& args, const ScriptCharacter character) {
 	@arg coord1 Coordinates
 	@arg arg6 Boolean true/false
 */
-bool opcode_0197(const ScriptArguments& args, const ScriptPlayer player, ScriptVec2 coord0, ScriptVec2 coord1, const ScriptBoolean arg6) {
-	RW_UNIMPLEMENTED_OPCODE(0x0197);
-	RW_UNUSED(player);
-	RW_UNUSED(coord0);
-	RW_UNUSED(coord1);
-	RW_UNUSED(arg6);
-	RW_UNUSED(args);
-	return false;
+bool opcode_0197(const ScriptArguments& args, const ScriptPlayer player, const ScriptVec2 coord0, const ScriptVec2 coord1, const ScriptBoolean arg6) {
+	auto condition = player->getCharacter()->getCurrentVehicle() == nullptr;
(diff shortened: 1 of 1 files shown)


rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -4792,13 +4712,12 @@ bool opcode_019f(const ScriptArguments& args, const ScriptPlayer player, const S
 	opcode 01a0
 	@arg player 
 	@arg coord0
-	@arg coord1
+	@arg coord1
```
JayFoxRox: unintended?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -4792,13 +4712,12 @@ bool opcode_019f(const ScriptArguments& args, const ScriptPlayer player, const S
...
 	@arg coord0
-	@arg coord1
+	@arg coord1 
 	@arg arg8 
 */
 bool opcode_01a0(const ScriptArguments& args, const ScriptPlayer player, const ScriptVec3 coord0, const ScriptVec3 coord1, const ScriptInt arg8) {
-	auto plyChar = player->getCharacter();
-	auto condition = plyChar->isStopped() && plyChar->getCurrentVehicle() != nullptr;
-	return script::objectInArea(args, plyChar, coord0, coord1, arg8, condition);
+	auto condition = player->getCharacter()->isStopped() && player->getCharacter()->getCurrentVehicle() != nullptr;
```
JayFoxRox: probably makes more sense to first check the vehicle state.
The `isStopped` might not even be necessary and probably depends on knowing if the char or their vehicle has stopped. You should also check what the precise condition for a vehicle having stopped is (forward / backward speed close to zero? Overall speed close to zero? Characters speed close to zero? Handbrake applied?).

If we can't figure it out now, we should at least keep an eye on those details (and track them).

ShFil119: As far I know function update will set flag stopped when speed is close to zero. I'll add extra condition checking vehicle.

rwengine/src/script/modules/GTA3ModuleImpl.inl
ShFil119: Pardon, file should be reformatted. Some lines are ended with space, some not.

ShFil119:
Updated:
Used JaxFoxRox's suggestions.
Converted arguments to be constant.
Fix : some of opcodes should use objectInRadius.

PR #324, opened 2017-10-10
Improve rendering performance on Intel

danhedron:
This is a simple change to get some extra performance where it matters, on integrated GPUs.

- Don't request a new buffer for every draw.
- Allocate as large a buffer as possible and copy per-draw data into it sequentially until it is full, then allocate a new buffer and start over.

- [x] respect GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT in entry count calculations.


ShFil119:
On sandy also performance improved ~30%.
(hd3000)

PR #327, opened 2017-10-13
Use memcpy instead of (glibc-specific) mempcpy.

christoph-heiss:
This fixes compilation macOS - it does not have mempcpy.

The return value is unused anyway, so no semantic change.


danhedron:
Oops. Thanks for fixing this quickly.

PR #330, opened 2017-10-19
Some potential deference of null pointers

ShFil119:
It should increase stability a bit.
- [x] cleanup


husho:
RW_CHECK presents only in debug builds and all it does is halt execution if conditions are met, not sure how it suppose to improve stability

ShFil119:
I have expressed myself badly. :( Actually almost all de-referencing are checked in debug, except these ones. So it should help debug openrw easier.

JayFoxRox:
Kind of pointless (I also dislike the `RW_CHECK` macro as I preffer `assert()` style conditions - but that's not your fault).

Otherwise LGTM (reviewed code, but did not test).

darkf:
LGTM.

PR #333, opened 2017-10-26
Modernify CMake Usage

danhedron:
- Update rwlib & rwengine to avoid using directory level commands.
- Import GLM as a subproject instead of using system package.

- [x] Improve Game, Library and Test target definitions.


danhedron:
Superseded by #335

PR #335, opened 2017-10-27
CMake rewrite

madebr:
- Require CMake version 3.2.2 (=version on Travis)
- Use a `rw_interface` target that will hold all compiler options.
Linking with that target will inherit the options.
- No more general `include_directories`/`add_definition`s
- `if(${XX})` --> `if(XX)`
- No more tabs in the cmake files, only spaces (4 spaces)
- Instead of using `dependency_INCLUDE_DIR` and `dependency_LIBRARIES` variables, use a `dependency::dependency` target to link with. (targets have an include and library property)'
  * Boost is the exception. Too much work to create a target for each component.
  * CMake 3.5 has a `FindBoost.cmake` module that creates exported targets. If we wait, Boost targets will come to us.
- Enable CTest
- Enable a CDash for OpenRW: http://my.cdash.org/index.php?project=OpenRW
  * Travis uses test scripts for continuous testing. (see `cmake/ctests/travis_script.cmake`)
    - A file `FATAL_ERROR` is created when an error occurs.
    - Fail the Travis build when that file exists.
  * This infrastructure can be used for the Windows build.
- I have capitalized all source files of the tests. To bring a bit of uniformity in it.
  * I wanted to be able to run each unit test separately using ctest.
  * That way, CDash will display the evolution of the test results.
  * But I ran into some problems. Boost Unit Test fails when a test suite has no unit tests. This causes Travis to fail. (I'm looking into it.)
  * Each test suite can be run separately using the CMake configuration option `SEPERATE_TEST_SUITES`
- The `cmake: move options and compiler configuration to separate files` patch is a **RFC**: I have extracted 2 files from the main `CMakeLists.txt` file. The main `CMakeLists.txt` looks more clean now:
  * `cmake_options.cmake`: all configure options of our projects are put there
  * `cmake_configure.cmake`: creation of the `rw_interface` target that will hold all compiler options.


madebr:
- I have added a patch to fix the `-Wunused-variable` warnings.
  * Travis builds now have 0 warnings: http://my.cdash.org/index.php?project=OpenRW&date=2017-10-27
  * The Continuous builds with site=Travis are those build by Travis.
- I have added a script (`cmake/ctest/experimental_script.ctest`) to build OpenRW and automatically upload the results to CDash.
  * Results of this script can be seen in the Experimental section of the CDash.
  * Building OpenRW on my local machine still shows some build warnings (ignoring the ffmpeg deprecation warnings).
  * The tests also fail.
    - Fixing those is future work :)

madebr:
- I have added a patch that adds coverage to the CTest `experimental_script.ctest`:
  * Output with one test suite (`SEPARATE_TEST_SUITES=0`): http://my.cdash.org/viewCoverage.php?buildid=1313521
  * Output with multiple test suite (`SEPARATE_TEST_SUITES=1`): http://my.cdash.org/viewCoverage.php?buildid=1313634

madebr:
- What is a good default test timeout?
  * Maybe set a test timeout per test suite? (now set a generous 300 seconds per suite)

danhedron:
This is looking good to me, how much more do you intend to change before it can be integrated?

madebr:
- Added [include-what-you-use](https://include-what-you-use.org/) static analyzer support:
  * still had many spurious warnings which should be silenced (using [mappings](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUMappings.md))
- I think the main CMake rewrite work is done,
In future PR's I would like to add support for other analyzers: Valgrind, linters, ...

PR #337, opened 2017-10-28
Avoid making unnecessary copies

ShFil119:
To think about:
- anonymous union
It prevents making reasonable contructor.
https://github.com/rwengine/openrw/blob/3c83e7cf52676dd27924bffb98713c712b4ded32/rwengine/src/script/ScriptMachine.cpp#L47
- opcode 0219
For example calling element nr size()-1  isn't thread safe.
https://github.com/rwengine/openrw/blob/512ac6f5eed3f6a356aac6bad3f28d8936faa25d/rwengine/src/script/modules/GTA3ModuleImpl.inl#L6280

### DIFF
--- a/rwengine/src/loaders/LoaderIFP.hpp
+++ b/rwengine/src/loaders/LoaderIFP.hpp
@@ -14,8 +14,19 @@ struct AnimationKeyframe {
     glm::quat rotation;
     glm::vec3 position;
     glm::vec3 scale;
-    float starttime;
-    int id;
+    float starttime = 0.f;
+    int id = 0;
+
+    AnimationKeyframe(glm::quat _rotation, glm::vec3 _position, glm::vec3 _scale, float _starttime, int _id)
+        : rotation(_rotation)
+        , position(_position)
+        , scale(_scale)
+        , starttime(_starttime)
+        , id(_id) {
+    }
+
+    AnimationKeyframe() {
+    }
 };
 
 struct AnimationBone {
--- a/rwengine/src/render/OpenGLRenderer.hpp
+++ b/rwengine/src/render/OpenGLRenderer.hpp
@@ -14,6 +14,22 @@ typedef uint64_t RenderKey;
 
 typedef std::uint32_t RenderIndex;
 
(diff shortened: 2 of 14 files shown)


rwengine/src/render/OpenGLRenderer.hpp
```diff
@@ -14,23 +14,41 @@ typedef uint64_t RenderKey;
 
 typedef std::uint32_t RenderIndex;
 
-struct VertexP3 {
+// just a handle
+struct VertexP {
```
danhedron: What is this for?

rwengine/src/loaders/LoaderIFP.hpp
```diff
@@ -16,6 +16,17 @@ struct AnimationKeyframe {
...
+
+    AnimationKeyframe(glm::quat _rotation, glm::vec3 _position, glm::vec3 _scale, float _starttime, int _id)
+        : rotation(_rotation)
+        , position(_position)
+        , scale(_scale)
+        , starttime(_starttime)
+        , id(_id) {
+    }
+
+    AnimationKeyframe() {
```
danhedron: Is it safe to leave the members uninitialised? Where is this ctor used

ShFil119: Here:
https://github.com/ShFil119/openrw/blob/0b11659c81b4ee3e8dcd813a84af6a4ff5bb3a7d/rwengine/src/loaders/LoaderIFP.cpp#L35
But few moment later it's overwritten here:
https://github.com/ShFil119/openrw/blob/0b11659c81b4ee3e8dcd813a84af6a4ff5bb3a7d/rwengine/src/loaders/LoaderIFP.cpp#L9
If you want, I can change.

rwengine/src/render/OpenGLRenderer.hpp
ShFil119: I was thinking it - "@todo normalize this to have the same interface as VertexP3" can mean there's possibility that in future one vector will store VertexP2 and VertexP3.

PR #339, opened 2017-10-29
Add build time option for CXX17 or boost filesystem

danhedron:
The existing filesystem usage has been adjusted to using the rwfs namespace.
The FILESYSTEM_LIBRARY option allows a choice between how the rwfs namespace will be defined:

- ___CXX17___: using the ``<filesystem>`` header & ``std::filesystem``.
- ___CXXTS___: using the ``<experimental/filesystem>`` header and ``std::experimental::filesystem``.
- ___BOOST___: using boost filesystem. The default for now as we still target C++14.

### DIFF
--- a/cmake_configure.cmake
+++ b/cmake_configure.cmake
@@ -38,6 +38,25 @@ else()
     message(FATAL_ERROR "Unknown platform \"${CMAKE_SYSTEM_NAME}\". please update CMakeLists.txt.")
 endif()
 
+if(FILESYSTEM_LIBRARY STREQUAL "CXX17")
+    target_compile_definitions(rw_interface INTERFACE "RW_FS_LIBRARY=0")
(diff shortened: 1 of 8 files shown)


madebr:
- I would move the `if`-ery you put in `rwlib/CMakeLists.txt` to `cmake_configure.cmake` and apply it to `rw_interface` instead of `rwlib`.
- `${Boost_SYSTEM_LIBRARY}` is a dependency of `${Boost_FILESYSTEM_LIBRARY}`. Normally the `FindBoost.cmake` bundled with CMake should add the system-library automatically.
  * But I've noticed (on Windows for instance) that when using a newer version of Boost, which CMake does not recognize, I still have to manually add the system library.
  * Therefore, I'ld add `${Boost_SYSTEM_LIBRARY}` along the `${Boost_FILESYSTEM_LIBRARY}`

PR #344, opened 2017-11-03
Miscellaneous patches

madebr:
This PR contains patches from all over the place. Changes are small, imho:
- Add `error_code` to `rwfs` namespace: some functions need it for the non-throwing version.
- Allow the configuration file to be passed by argument option
- `GameConfig` returns `rwfs::path` objects if paths are requested
- `GameData` and `LoaderIMG` accept `rwfs::path` objects
- Create strings and matrices in one go (less code)
- Remove some unused function definitions, function implementations and files
- Group the arguments thematically instead of one block


rwengine/src/engine/GameData.cpp
```diff
@@ -695,12 +695,14 @@ float GameData::getWaveHeightAt(const glm::vec3& ws) const {
     return (1 + std::sin(engine->getGameTime() + (ws.x + ws.y) * WATER_SCALE)) *
            WATER_HEIGHT;
 }
-
-bool GameData::isValidGameDirectory(const std::string& path) {
-    if (path.empty()) {
+#include <boost/system/error_code.hpp>
```
danhedron: What does this include do here?

madebr: Rest over include. Nothing apparently.

rwgame/states/IngameState.cpp
```diff
@@ -351,18 +351,7 @@ const ViewCamera& IngameState::getCamera(float alpha) {
...
-        m[0][2] = up.x;
-
-        m[1][0] = direction.y;
-        m[1][1] = right.y;
-        m[1][2] = up.y;
-
-        m[2][0] = direction.z;
-        m[2][1] = right.z;
-        m[2][2] = up.z;
+        glm::mat3 m(direction, right, up);
```
danhedron: This change totally jumbles the camera angles

rwlib/source/loaders/LoaderIMG.cpp
```diff
@@ -5,9 +5,9 @@
 LoaderIMG::LoaderIMG() : m_version(GTAIIIVC), m_assetCount(0) {
 }
 
-bool LoaderIMG::load(const std::string& filename) {
-    auto baseName = filename;
-    auto extpos = filename.find(".img");
+bool LoaderIMG::load(const rwfs::path& filepath) {
+    auto baseName = filepath.string();
+    auto extpos = baseName.find(".img");
```
danhedron: Could this be achieved with path::stem() instead?

rwgame/states/IngameState.cpp
madebr: Oops, screw up on my part!

rwlib/source/loaders/LoaderIMG.cpp
madebr: I changed the code with `replace_extension`.
Too bad C++ does not have `with_extension` (like Python has with [with_suffix](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_suffix))

rwgame/GameBase.cpp
```diff
@@ -1,26 +1,28 @@
...
 const std::string kBuildStr(kGitSHA1Hash, 8);
 const std::string kWindowTitle = "RWGame";
+const std::string kDefaultConfigFileName = "openrw.ini";
 constexpr int kWindowWidth = 800;
 constexpr int kWindowHeight = 600;
 
-GameBase::GameBase(Logger &inlog, int argc, char *argv[]) : log(inlog) {
+GameBase::GameBase(Logger &inlog, int argc, char *argv[]) :
+        log(inlog),
+        config(),
```
danhedron: Why put these empty constructors in the initaliser list?

madebr: I always try to be explicit about these things.

PR #346, opened 2017-11-08
Braced initialization

ShFil119:
Should increase readability a bit. ;)


rwlib/source/loaders/RWBinaryStream.hpp
```diff
@@ -67,7 +67,7 @@ class RWBStream {
      * @brief Returns a new stream for the data inside this one.
      */
     RWBStream getInnerStream() const {
-        return RWBStream(_dataCur, _currChunkSz);
+        return {_dataCur, _currChunkSz};
```
danhedron: This introduces a lot of warnings because of the ptrdiff_t -> size_t conversion. The type of _currChunkSz should be changed to size_t.

ShFil119: With static cast?

danhedron: No. The type of the member should be changed.

PR #349, opened 2017-11-17
Make vehicle wheels steer smoothly

drummyfish:
Hello, this is my first attempt at PR to this repo. Feel free to throw criticism at me :)

EDIT:

Unlike in vanilla engine, the vehicle wheels instantly turned to their limit angle when the player steered. I corrected this by smoothly changing the angle in each physics tick, independently of framerate. It's not just a change of animation but actual physics behavior that allows for finer steering.

### DIFF
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -262,14 +262,23 @@ void VehicleObject::tickPhysics(float dt) {
             physVehicle->setBrake(brakeReal * brakeF, w);
 
             if (wi.m_bIsFrontWheel) {
-                float sign = std::signbit(steerAngle) ? -1.f : 1.f;
-                physVehicle->setSteeringValue(
-                    std::min(glm::radians(info->handling.steeringLock),
-                             std::abs(steerAngle)) *
-                        sign,
-                    w);
-                // physVehicle->setSteeringValue(std::min(3.141f/2.f,
-                // std::abs(steerAngle)) * sign, w);
+                float currentVal = physVehicle->getSteeringValue(w);
+                float currentSign = std::signbit(currentVal) ? -1.f : 1.f;
+                float newVal;
+
+                if (std::abs(steerAngle) < 0.001f) {   // no steering?
+                    newVal = std::max(0.0f,std::abs(currentVal) - 4.f * dt) *
+                        currentSign; 
+                } else {
+                    newVal = currentVal + steerAngle * dt * 4.f;
+
+                    float limit = glm::radians(info->handling.steeringLock);
+
+                    if (std::abs(newVal) > limit)
+                        newVal = limit * currentSign;
+                }
+
+                physVehicle->setSteeringValue(newVal,w);
             }
         }
 


ShFil119:
I'll give you an advice, which I got. ;)
"Thank you for your contribution, @drummyfish!

I'd like to note that it would be much easier to manage the project and work with PRs if you could provide more definitive commit messages and PR titles.

There's an article that has once helped me understand the importance of being as verbose as possible while working with git, and I'd like to share it with you as I see your contributions being valuable, but the commit messages creating a bottleneck in reviewing and merging them in because it's a big project and it's not easy to maintain.

Here's the link to the article: https://chris.beams.io/posts/git-commit/" ;)

drummyfish:
Thank you! I try to be reasonable with the comments, but I'm still a newbie to contributing, so I will definitely read it. I noticed the code is quite nicely commented compared to some other projects, so I guess you emphasize good documentation, which is definitely good.

ShFil119:
Maintainers are just "picky". So if you want to make your code merged, you should try each available method to satisfy them. 

Good documentation is one best strategy. ;)

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -262,14 +262,23 @@ void VehicleObject::tickPhysics(float dt) {
...
-                float sign = std::signbit(steerAngle) ? -1.f : 1.f;
-                physVehicle->setSteeringValue(
-                    std::min(glm::radians(info->handling.steeringLock),
-                             std::abs(steerAngle)) *
-                        sign,
-                    w);
-                // physVehicle->setSteeringValue(std::min(3.141f/2.f,
-                // std::abs(steerAngle)) * sign, w);
+                float currentVal = physVehicle->getSteeringValue(w);
+                float currentSign = std::signbit(currentVal) ? -1.f : 1.f;
```
CookiePLMonster: `std::copysign(1.0f, currentVal)` may be handy here.

darkf:
Anyone tested this yet?

darkf:
I suppose I'll ask again -- @ShFil119, @husho, et al -- anyone up for testing and reviewing this?

husho:
I have tested it on master and it works fine.

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -262,14 +262,23 @@ void VehicleObject::tickPhysics(float dt) {
...
+                float newVal;
+
+                if (std::abs(steerAngle) < 0.001f) {   // no steering?
+                    newVal = std::max(0.0f,std::abs(currentVal) - 4.f * dt) *
+                        currentSign; 
+                } else {
+                    newVal = currentVal + steerAngle * dt * 4.f;
```
husho: Not sure if this is the right step value. My guess (from quick glance at decompiled code) it's  ```1 / 0.35 = 2.85714285714 ```

rwengine/src/objects/VehicleObject.cpp
JayFoxRox: Good suggestion; nit.

JayFoxRox:
I think this is fine as a temporary solution, but it feels slightly hacky / overcomplicated.

LGTM if @drummyfish is not around to address these issues.
In that case, an issue should be created to address some of the review issues (fine-tuning the interpolation, improving code quality)

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -262,14 +262,23 @@ void VehicleObject::tickPhysics(float dt) {
...
+
+                    float limit = glm::radians(info->handling.steeringLock);
+
+                    if (std::abs(newVal) > limit)
+                        newVal = limit * currentSign;
```
JayFoxRox: nit: [I like curly braces](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/). The google code style doesn't care, so this is a personal opinion.

dracc:
I'd be happy addressing these points and filing a 'fix' PR.

rwengine/src/objects/VehicleObject.cpp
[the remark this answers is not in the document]
dracc: Would you mind expand on how that would be accomplished? I don't see the case you're thinking about. :(

PR #353, opened 2017-11-23
Use Docker on Travis CI

madebr:
Hey,
By using Docker on Travis, we are no longer stuck on Ubuntu 14.04 and the older packages (e.g. older ffmpeg) there.

The penalty is an increase in build time to about 11 minutes.

I've tried caching the Docker image but that has no influence on build time.

This PR allows to remove the deprecated libavcodec calls.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+build
+.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,38 @@
-sudo: required
-dist: trusty
-language: generic
-env:
-  - CXX=g++-5 CC=gcc-5
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
-    packages:
-      - g++-5
-      - libbullet-dev
-      - libsdl2-dev
-      - libglm-dev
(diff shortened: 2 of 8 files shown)


ShFil119:
+1001 for docker. As we constantly make mac's build break, I was thinking about preparing mac build in travis. Also add clang and static analyzers. 

You can see my struggle here: https://github.com/ShFil119/openrw/pull/12
https://github.com/ShFil119/openrw/pull/11

For example:
Travis's version of clang doesn't want to work with stdlibc++... And libc++ doesn't work with bullet without rebuild. I think using docker is inevitable.

madebr:
- I've added tests for Arch Linux and Fedora Linux. Which is probably too much.
- The tests on Arch and Fedora fail because it needs #358 and #359.
- The MacOS image takes forever to boot. (currently not tested well enough because it just takes too long to start)

ShFil119:
It looks like xcode forces deeper changes. 
Here we gcc: https://travis-ci.org/ShFil119/openrw/jobs/320834101

Can we split it to another pull?

danhedron:
I merged these commits separately  🎉

PR #357, opened 2017-12-07
Fix cmake 3.10

ShFil119:
It's dump solution, but works.

Can someone with macOS check how it works?
It's preferred to use GLVND, but in case of inability to build I'll set legacy solution.


madebr:
What about this branch ShFil119? https://github.com/madebr/openrw/tree/cmake_optional_override

ShFil119:
Works and looks better. Please open pull.

PR #362, opened 2017-12-21
Small fixes

ShFil119:
In this pull I want to take care about one memory leak and undefined behavior.

Without virtual destructor: https://gist.github.com/ShFil119/a8dd2447957b9dc0dcd59f651315ebfc
With: https://gist.github.com/ShFil119/49403c30de189d07ec39d9c12065f6aa

Before: https://gist.github.com/ShFil119/d57738e91c9ee148689e85358f1241e1
After: https://gist.github.com/ShFil119/8c41e5c93392e7017b782f7eaecbc5ea
Edit I wanted open pull on my fork... I set tag work in progress.
To do:
- [x] cleanup
- [x] fix also commented code

### DIFF
--- a/rwengine/src/render/OpenGLRenderer.cpp
+++ b/rwengine/src/render/OpenGLRenderer.cpp
@@ -468,3 +468,6 @@ const Renderer::ProfileInfo& OpenGLRenderer::popDebugGroup() {
 #endif
     return profileInfo[0];
 }
+
+Renderer::ShaderProgram::~ShaderProgram() {
+}
--- a/rwlib/source/rw/bit_cast.cpp
+++ b/rwlib/source/rw/bit_cast.cpp
@@ -1,0 +1,16 @@
+#ifndef _LIBRW_BIT_CAST_CPP_
+#define _LIBRW_BIT_CAST_CPP_
+
+//Based on https://gist.github.com/socantre/3472964
+#include <cstring> // memcpy
+#include <type_traits> // is_trivially_copyable
+#include "rw/defines.hpp" // RW_ASSERT
+
+template <class Dest, class Source>
+inline Dest bit_cast(Source const &source) {
+    Dest dest = Dest{};
+    std::memcpy(&dest, &source, sizeof(Dest));
+    return dest;
+}
+
+#endif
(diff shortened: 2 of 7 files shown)


rwengine/src/render/OpenGLRenderer.cpp
```diff
@@ -468,3 +468,5 @@ const Renderer::ProfileInfo& OpenGLRenderer::popDebugGroup() {
 #endif
     return profileInfo[0];
 }
+
+Renderer::ShaderProgram::~ShaderProgram() = default;
```
danhedron: You already added the pure dtor in the header, what's the meaning behind this?

rwlib/source/rw/bit_cast.cpp
```diff
@@ -0,0 +1,19 @@
...
+#define _LIBRW_BIT_CAST_CPP_
+
+//Based on https://gist.github.com/socantre/3472964
+#include <cstring> // memcpy
+#include <type_traits> // is_trivially_copyable
+#include "rw/defines.hpp" // RW_ASSERT
+
+template <class Dest, class Source>
+inline Dest bit_cast(Source const &source) {
+    RW_ASSERT(std::is_trivially_copyable<Dest>::value);
```
danhedron: You can make these ``static_assert``, since they're know at compile time.

rwlib/source/rw/bit_cast.cpp
```diff
@@ -0,0 +1,19 @@
...
+    RW_ASSERT(std::is_trivially_copyable<Source>::value);
+
+    Dest dest = Dest{};
+    std::memcpy(&dest, &source, sizeof(Dest));
```
danhedron: You don't check that sizeof(Dest) <= sizeof(Source).

Is it necessary to specify the Source where bit_cast is used? It should be deduced.

rwlib/source/rw/bit_cast.cpp
ShFil119: I'll change.

rwengine/src/render/OpenGLRenderer.cpp
ShFil119: It's for generating dtor implementation. If you want I can replace with empty implementation.

PR #367, opened 2018-01-05
Remove refundant "virtual"

ShFil119:
Small cleanup. To think about - more functions are able to use c++11 const.
 Edit I've found some more in VehicleObject. Done.

### DIFF
--- a/rwengine/src/ai/DefaultAIController.hpp
+++ b/rwengine/src/ai/DefaultAIController.hpp
@@ -13,9 +13,9 @@ public:
         : CharacterController() {
     }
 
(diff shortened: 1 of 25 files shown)


CookiePLMonster:
It's a matter of preference really - what I'm seeing the most myself is **both** virtual and override specifiers. Not sure what's the standard for this codebase, however.

I do like the idea of using final (for entire classes if applicable) for applicable cases, too.

paulsapps:
hmm in my code I always use virtual and override.. I guess in the past I liked to make it clear it was virtual due to lack of override and got used to both or something..

PR #371, opened 2018-01-15
Use std::move

ShFil119:
One more emplace_back and usage std::move when possible.

- [x] remove unneeded std::move (POD/ptr)
- [x] avoid temporary zd
- [x] cleanup

### DIFF
--- a/rwengine/src/engine/SaveGame.cpp
+++ b/rwengine/src/engine/SaveGame.cpp
@@ -982,16 +982,9 @@ bool SaveGame::loadGame(GameState& state, const std::string& file) {
         Block11Zone& zone = zoneData.navZones[z];
         Block11ZoneInfo& day = zoneData.dayNightInfo[zone.dayZoneInfo];
         Block11ZoneInfo& night = zoneData.dayNightInfo[zone.nightZoneInfo];
-        ZoneData zd;
-        zd.name = zone.name;
-        zd.type = zone.type;
-        zd.min = zone.coordA;
-        zd.max = zone.coordB;
-        zd.island = zone.level;
         // @toodo restore gang density
-        zd.pedGroupDay = day.pedgroup;
-        zd.pedGroupNight = night.pedgroup;
-        gamezones.push_back(zd);
+        gamezones.emplace_back(zone.name, zone.type, zone.coordA, zone.coordB,
+                            zone.level, day.pedgroup, night.pedgroup);
     }
     // Re-build zone hierarchy
     for (ZoneData& zone : gamezones) {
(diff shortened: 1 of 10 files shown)


rwengine/src/engine/SaveGame.cpp
```diff
@@ -985,7 +985,7 @@ bool SaveGame::loadGame(GameState& state, const std::string& file) {
         // @toodo restore gang density
         zd.pedGroupDay = day.pedgroup;
         zd.pedGroupNight = night.pedgroup;
-        gamezones.push_back(zd);
+        gamezones.push_back(std::move(zd));
```
danhedron: This could be done with emplace_back, avoiding the temporary ``zd``.

danhedron:
A few of these are just PODs, where moving is the same as copying. You can just use ``move`` on the types that contain strings, vectors and other allocating types.

CookiePLMonster:
Does it matter if they are PODs? It's still a good practice, and emplacing is beneficial for pretty much any type (except fundamental types).

rwlib/source/data/Clump.hpp
```diff
@@ -149,6 +149,11 @@ struct Geometry {
         std::string name;
         std::string alphaName;
         TextureData::Handle texture;
+        Texture(std::string _name, std::string _alphaName, TextureData::Handle _texture)
+            : name(_name)
+            , alphaName(_alphaName)
```
CookiePLMonster: I believe those don't implicitly move - you might need to `std::move` them inside the initializer list.

rwlib/source/loaders/LoaderDFF.cpp
```diff
@@ -359,7 +359,7 @@ void LoaderDFF::readTexture(Geometry::Material &material,
 
     TextureData::Handle textureinst =
         texturelookup ? texturelookup(name, alpha) : nullptr;
-    material.textures.push_back({name, alpha, textureinst});
+    material.textures.emplace_back(name, alpha, textureinst);
```
CookiePLMonster: You should be using `std::move` here too, as now you're passing your strings by value. Looked up the full function and looks like moving those strings away is perfectly safe.

rwlib/source/data/Clump.hpp
ShFil119: I will try to use perfect forwarding later. Having quite busy week. ;)

CookiePLMonster: Cool =) Though in this case you don't need perfect forwarding - `std::move` alone does fine and is the best approach no matter if you're moving or copying strings into this constructor.

Issue #368, opened 2018-01-08
HUD element flashing


danhedron:
Used in the intro sequence and when you get out of hospital.

Related Opcodes:
* [03E7](http://www.gtamodding.com/wiki/03E7) Flash HUD Object.

Notes:
The game HUD is drawn in DrawUI.cpp.

CookiePLMonster:
Game flashes HUD elements by doing a `& 4` on an internal frame counter (so you get 3 frames of element not showing and 1 frame of element showing).

Since this behaviour is FPS dependant, I'd recommend making it timer based instead - 125ms off, 25ms on? Or 99 + 33 if you want to replicate default behaviour the most accurately.

PR #374, opened 2018-01-21
Map flashing

Florin9doi:


rwengine/src/engine/GameState.hpp
```diff
@@ -301,6 +301,7 @@ class GameState {
     float cutsceneStartTime;
     /** Flag for rendering cutscene letterbox */
     bool isCinematic;
+    int hudFlash;
```
ShFil119: I think we should initialize it. In order to prevent UB.

Florin9doi: done

rwgame/DrawUI.cpp
```diff
@@ -32,26 +32,29 @@ void drawMap(ViewCamera& currentView, PlayerController* player,
...
 
-    glm::quat camRot = currentView.rotation;
-
-    map.rotation = glm::roll(camRot) - glm::half_pi<float>();
-    map.worldSize = ui_worldSizeMin;
-    map.worldSize = ui_worldSizeMax;
-    if (player) {
-        map.worldCenter = glm::vec2(player->getCharacter()->getPosition());
-    }
+    bool displayMap = (fmod(world->getGameTime(), 0.5) >= .25);
```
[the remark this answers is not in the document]
Florin9doi: I'm not sure what's wrong with this one, but it comes from https://github.com/rwengine/openrw/blob/04ed374de86e72c7d2866238d8fc585f999e4310/rwgame/DrawUI.cpp#L108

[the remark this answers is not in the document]
ShFil119: I was thinking about small suggestion - changing "fmod" to "std::fmod". And 0.5, 0.25 to 0.5f, 0.25f. ;) 
Actually it forces 64bit mode.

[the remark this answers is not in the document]
Florin9doi: should I change both?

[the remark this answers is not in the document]
Florin9doi: done

rwengine/src/engine/GameState.hpp
danhedron: Instead of int, this could be an enum class so that the possible values are well named.

rwgame/DrawUI.cpp
```diff
@@ -32,26 +32,29 @@ void drawMap(ViewCamera& currentView, PlayerController* player,
...
 
-    glm::quat camRot = currentView.rotation;
-
-    map.rotation = glm::roll(camRot) - glm::half_pi<float>();
-    map.worldSize = ui_worldSizeMin;
-    map.worldSize = ui_worldSizeMax;
-    if (player) {
-        map.worldCenter = glm::vec2(player->getCharacter()->getPosition());
-    }
+    if (world->state->hudFlash != 8
```
danhedron: An enum class would make the meaning of this condition more obvious.

rwgame/DrawUI.cpp
```diff
@@ -32,26 +32,29 @@ void drawMap(ViewCamera& currentView, PlayerController* player,
...
 
-    glm::quat camRot = currentView.rotation;
-
-    map.rotation = glm::roll(camRot) - glm::half_pi<float>();
-    map.worldSize = ui_worldSizeMin;
-    map.worldSize = ui_worldSizeMax;
-    if (player) {
-        map.worldCenter = glm::vec2(player->getCharacter()->getPosition());
-    }
+    if (world->state->hudFlash != HudFlash::FlashRadar
```
danhedron: It would be tidier to do this check down in drawHUD, just to avoid changing the indentation here.

rwgame/DrawUI.cpp
```diff
@@ -104,8 +107,9 @@ void drawPlayerInfo(PlayerController* player, GameWorld* world,
 
     infoTextY += ui_textHeight;
 
-    if (player->getCharacter()->getCurrentState().health > ui_lowHealth
-        || fmod(world->getGameTime(), 0.5) >= .25) { // UI: Blinking health indicator if health is low
+    if ((world->state->hudFlash != HudFlash::FlashHealth
+        && player->getCharacter()->getCurrentState().health > ui_lowHealth)
+        || std::fmod(world->getGameTime(), 0.5f) >= .25f) { // UI: Blinking health indicator if health is low
```
danhedron: Based on [this](https://github.com/rwengine/openrw/issues/368#issuecomment-357550152) comment, the timer would be more accurate if it was ``fmod(t, 0.150f) <= 0.025f``.

rwengine/src/engine/GameState.hpp
```diff
@@ -243,6 +243,13 @@ struct GarageInfo {
     }
 };
 
+enum HudFlash {
```
danhedron: This could be ``enum class HudFlash``, as you're already referencing the values by their enum scope.

rwgame/DrawUI.cpp
Florin9doi: The map should flash about twice per second: https://youtu.be/klIcUiEw5jY?t=5m59s
The sum of shown and hiden durations should be about 500 ms. 
With fmod(t, 0.150f) the map will flash 6.7 times per second which is wrong.

danhedron:
Looks good to me.

PR #377, opened 2018-01-26
Use range loop

ShFil119:
Small clean up. It should increases readability a bit.  
- [x] Some of values are hardcoded, I mean written by a hand. Not sure, is it a part of gta3? 
- [x] some of them can be const
- [x] madebr's suggestions

And most important - has this pull a meaning or value? Do you want range loops?

### DIFF
--- a/rwengine/src/data/GameTexts.cpp
+++ b/rwengine/src/data/GameTexts.cpp
@@ -2,8 +2,8 @@
 
 GameString GameStringUtil::fromString(const std::string& str) {
     GameString s;
-    for (std::string::size_type i = 0u; i < str.size(); ++i) {
-        s += str[i];
+    for (const char &i : str) {
+        s += i;
     }
     return s;
 }
--- a/rwgame/MenuSystem.hpp
+++ b/rwgame/MenuSystem.hpp
@@ -126,9 +126,9 @@ public:
 
     void click(const float x, const float y) {
         glm::vec2 c(x - offset.x, y - offset.y);
-        for (auto it = entries.begin(); it != entries.end(); ++it) {
+        for (auto &entry : entries) {
             if (c.y > 0.f && c.y < size) {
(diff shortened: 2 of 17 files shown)


rwgame/MenuSystem.hpp
```diff
@@ -126,9 +126,9 @@ class Menu {
 
     void click(const float x, const float y) {
         glm::vec2 c(x - offset.x, y - offset.y);
-        for (auto it = entries.begin(); it != entries.end(); ++it) {
+        for (auto &entrie : entries) {
```
madebr: entry

rwlib/source/loaders/LoaderIMG.cpp
```diff
@@ -40,9 +40,9 @@ bool LoaderIMG::load(const rwfs::path& filepath) {
 /// Get the information of a asset in the examining archive
 bool LoaderIMG::findAssetInfo(const std::string& assetname,
                               LoaderIMGFile& out) {
-    for (size_t i = 0; i < m_assets.size(); ++i) {
-        if (boost::iequals(m_assets[i].name, assetname)) {
-            out = m_assets[i];
+    for (auto &m_asset : m_assets) {
```
madebr: m_asset -> asset

rwengine/src/data/GameTexts.cpp
```diff
@@ -2,8 +2,8 @@
 
 GameString GameStringUtil::fromString(const std::string& str) {
     GameString s;
-    for (std::string::size_type i = 0u; i < str.size(); ++i) {
-        s += str[i];
+    for (const char &i : str) {
+        s += i;
```
madebr: Is it possible to use `std::accumulate`?

ShFil119: yeah, but code becomes ugly(lambda)

Issue #360, opened 2017-12-14
Travis MacOS builds


madebr:
I've been trying to get Travis to build and test on MacOS without complete success.
Since I don't have access to a physical Mac, I don't know how to fix it.

* [This branch](https://github.com/madebr/openrw/tree/macos) contains the basic cruft to get MacOS builds on Travis going ([Travis log](https://travis-ci.org/madebr/openrw/jobs/318225814)).
* [This branch](https://github.com/madebr/openrw/tree/macos_unknown) contains an extra (bad) patch to get the MacOS builds proceed further ([Travis log](https://travis-ci.org/madebr/openrw/jobs/318225992))

The goal is to build OpenRW on MacOS X the same way as we do on Linux (using brew instead of apt-get).

I hope people with actual Apple hardware can solve this. (@aldoxtor @christoph-heiss )

Related: #343 and #354

aldoxtor:
It looks like the last PR fix building on macOS.

danhedron:
Travis is now building macOS.

Issue #385, opened 2018-02-06
Broken compatibility with libglm-dev >= 0.9.9


dracc:
libglm version 0.9.9 or greater breaks the game, menu background is black, most of the game is black (with occasional gray stripes). [Menu](http://i.imgur.com/ubRs4xv.png) [In-game](http://i.imgur.com/fR08hoy.png)
This seems to be due to glm removing default initialization of its data types.

Also worth noting that compilation requires the flag GLM_ENABLE_EXPERIMENTAL.

danhedron:
This should be fixed now.

PR #390, opened 2018-02-09
Miscellaneous macOS and rwviewer changes.

christoph-heiss:
Hi again!

Just some (I guess as usual you could say) some macOS-related stuff.
Maybe I get around sometime to do some hacking on the engine rather than just fixing build failures.

Anyway, a build-fixing commit, one for fixing DPI scaling in rwviewer and another fixing a rather minor inconvenient segfault.

I did not have the chance yet to test this changes on a linux box (although I do think nothing should have broken), so if someone could test that for me would be much appreciated! 

Thanks and with best regards!


danhedron:
There's an existing failure with the latest Fedora. If macOS is able to build on Travis, could you swap fedora for that?

christoph-heiss:
I pushed another commit to fix rwviewer-build on macOS and another to re-enable travis macOS builds as they seem to work fine (fixes #360).
Except one apparent segfault in "test_clump_clone", which I am going to have a look at (but this one is not really mission-critical I think).
`1: unknown location:0: fatal error: in "LoaderDFFTests/test_clump_clone": memory access violation at address: 0x00001420: no mapping at fault address
1: /Users/travis/build/rwengine/openrw/tests/test_LoaderDFF.cpp:66: last checkpoint`

But atleast with this we can know spot build regressions on macOS more easily and immediately.

And what do you mean with swapping fedora for it? I'm not quite sure. The fedora issue seems to be related to glm (probably #385 as well, I guess).

ShFil119:
Yeah, fedora contains outdated version of glm. Error is triggered by gcc 7.3. Arch and ubuntu have newer packages with fixes.

danhedron:
Looking into the failure, there's already a fix making its way through the fedora updates stream so it should go away within a few days.

danhedron:
Everything seems to be working on my end.

dracc:
What was the conclusion? Will the fedora build be swapped out in favor of a macos build, will a fourth build be added or is it status quo?

christoph-heiss:
Since danhedron pointed out that the Fedora build should fix itself in a few days, with 5f97386 I (re-)added a forth build for macOS.

I think build-testing on a variety of Linux distros is a good idea to get full coverage.

PR #392, opened 2018-02-10
Add experimental flag needed by glm 0.9.9

dracc:
In order to build OpenRW with libglm-dev version >= 0.9.9,
the flag GLM_ENABLE_EXPERIMENTAL needs to be set.

This is now done in cmake_configure.cmake , I'm not sure if it's the best spot but it works.


dracc:
To clarify, it is GLM_GTX_NORM (glm/gtx/norm.h) and GLM_GTX_QUATERNION (glm/gtx/quaternion.h) that requires the flag. These files are responsible for, among many others, glm::length2() and glm::quat().

dracc:
As far as I can tell, this PR fixes #385

Issue #202, opened 2016-08-12
Use lint / cpplint


JayFoxRox:
The source code should be vetted for/with:
* Pointless includes & missing includes (missing includes may be helped by MSVC)
* Static analysis issues
* Address sanitizer

PR #396, opened 2018-02-13
Add sanitizers

ShFil119:
UBSAN and LSAN cause almost no overhead, TSAN 5x-15x, ASAN 2x.

To do:
- [x] integration with travis? (to think about in future - checking memory leaks)
- [x] As far I remeber madebr suggested something instead of add_compile_options.

//There's also memory sanitizer for uninitialized reads. But it requires clang and rebuilding dependency. 
Ref:
#202


madebr:
I would add only one option to `cmake_options.cmake`: `ENABLE_SANITIZERS`, and let it accept a list of sanitizers to enable.
So this would work:
e.g.
```
cmake .. -DENABLE_SANITIZERS=address;leak
```
or
```
cmake .. -DENABLE_SANITIZERS=undefined
```
I prefer this over adding an option for every possible sanitizer.

PR #398, opened 2018-02-20
Various, small changes

ShFil119:
Small improvements. ;) 

- [x] fix appveyor
- [x] convert char[] to string

### DIFF
--- a/rwengine/src/ai/CharacterController.hpp
+++ b/rwengine/src/ai/CharacterController.hpp
@@ -20,8 +20,7 @@ public:
      * @brief The Activity struct interface
      */
     struct Activity {
-        virtual ~Activity() {
-        }
+        virtual ~Activity() = default;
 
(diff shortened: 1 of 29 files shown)


dracc:
Is constexpr not available for std::string or why change the version strings into cstring and then cast them back into std::string?
Other than that it looks really nice! :+1:

madebr:
A string cannot be constexpr as it allocates memory on the heap dynamically.

danhedron:
Leave ``std::string`` as they are for now. Everything else seems like a step in the right direction.

ShFil119:
Updated, I've removed constexpr from:
- ViewCamera defaultView (glm doesn't support constexpr for angleAxis)
- six other vectors which MSVC doesn't like

darkf:
`sspaceRect` should be changed to use `std::array` in a future PR, but LGTM.

PR #403, opened 2018-03-06
Add clang-tidy support

madebr:
This PR adds the clang-tidy linter to openrw.
All clang-tidy checks are enabled by default.

See http://clang.llvm.org/extra/clang-tidy/

Enabling clang-tidy autofix might result in incorrect source code, so manual checking is required.

### DIFF
--- a/cmake_configure.cmake
+++ b/cmake_configure.cmake
@@ -141,6 +141,9 @@ if(CHECK_CLANGTIDY)
         "-style=file"
         "-checks=*"
     )
+    if(CHECK_CLANGTIDY_FIX)
+        list(APPEND CLANGTIDY_ARGS "-fix")
+    endif()
 endif()
 
 function(openrw_target_apply_options)
--- a/cmake_configure.cmake
+++ b/cmake_configure.cmake
@@ -125,17 +125,39 @@ endif()
 
 include(CMakeParseArguments)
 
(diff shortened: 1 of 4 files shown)


cmake_configure.cmake
```diff
@@ -125,17 +125,42 @@ endif()
...
+if(CHECK_CLANGTIDY)
+    find_program(CLANGTIDY_PROGRAM
+        clang-tidy
+    )
+    if(NOT CLANGTIDY_PROGRAM)
+        message(FATAL_ERROR "Cannot find clang-tidy")
+    endif()
+    set(CLANGTIDY_ARGS "${CLANGTIDY_PROGRAM}")
+    list(APPEND CLANGTIDY_ARGS
+        "-style=file"
```
ShFil119: Shouldn't be here "-format-style=file"?

madebr: The docs on http://clang.llvm.org/extra/clang-tidy/ show a `-format-style=file` option.
The version installed by fedora 26 does not have that option (version 4.0.1).
This maybe requires a `FindClangTidy.cmake` module if this option has changed.

ShFil119: On arch I have to use "-format-style=file".

madebr: I believe to have fixed it.

PR #404, opened 2018-03-09
Allow to pause and resume sounds

ShFil119:
Exceptionally I serve you something finished. (I hope.)

For future task:
- Divide sounds into groups. For example we want to play menu's sounds in menu. ;)

to do:
- [x] Think about performance (passing sound by ref?),


darkf:
LGTM, pauses audio in cutscenes.

PR #407, opened 2018-03-11
Jump and vehicle start/stop improvements

paulkocialkowski:
This adds a few improvements that fix issues that were slightly annoying when playing the game:
- Jumps did not get inertia from the previous movement
- Cars were very slow to start and stop, which was very noticeable when going from forward to reverse

### DIFF
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -35,6 +35,7 @@ CharacterObject::CharacterObject(GameWorld* engine, const glm::vec3& pos,
     , m_look(0.f, glm::half_pi<float>())
     , running(false)
     , jumped(false)
+    , jumpAnimation(NULL)
     , jumpSpeed(DefaultJumpSpeed)
     , motionBlockedByActivity(false)
     , cycle_(AnimCycle::Idle)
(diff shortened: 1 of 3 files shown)


rwengine/src/objects/CharacterObject.cpp
```diff
@@ -35,6 +35,7 @@ CharacterObject::CharacterObject(GameWorld* engine, const glm::vec3& pos,
     , m_look(0.f, glm::half_pi<float>())
     , running(false)
     , jumped(false)
+    , jumpAnimation(NULL)
```
ShFil119: Instead of NULL you should use nullptr.

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -490,6 +504,12 @@ void CharacterObject::jump() {
         physCharacter->jump(btVector3(0.f, 0.f, 0.f));
 #endif
         jumped = true;
+        jumpAnimation = animator->getAnimation(AnimIndexMovement);
+
+        // There is no kinetic energy left after a jump
+        if (jumpAnimation == animations->animation(AnimCycle::JumpLand))
```
ShFil119: After character lands there's weird feeling, like character's body lose its ability move horizontally.

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -201,44 +248,11 @@ glm::vec3 CharacterObject::updateMovementAnimation(float dt) {
...
-            glm::vec3 d = (b - a);
-            animTranslate.y += d.y;
-
-            // Kludge: Drop y component of root bone
-            auto t = glm::vec3(root->getTransform()[3]);
-            t.y = 0.f;
-            root->setTranslation(t);
-        }
-    }
+    if (jumped == true && jumpAnimation != NULL)
```
ShFil119: Also here.

rwengine/src/objects/CharacterObject.cpp
paulkocialkowski: Ah thanks, it shows that I've used to C over C++...

dracc:
Is this PR dead or do you plan to fix the pointed out issues, @paulkocialkowski ? :+1:

darkf:
@paulkocialkowski Thanks for your contribution.

Since it seems there are some unresolved issues (+ it needs a rebase), I'm going to close this for now to avoid confusion (and taking space in the queue), however you are more than welcome to open a new PR when/if you fix those issues. Thanks. 👍

PR #408, opened 2018-03-12
Implemented some locate opcodes

husho:


JayFoxRox:
Why was this closed? Will the PR be renewed?

PR #419, opened 2018-03-31
[WIP]Add more bit_casts in loaders

ShFil119:
Nothing big.

### DIFF
--- a/rwengine/src/loaders/LoaderCOL.cpp
+++ b/rwengine/src/loaders/LoaderCOL.cpp
@@ -9,6 +9,7 @@
 #include <glm/glm.hpp>
 
 #include "data/CollisionModel.hpp"
+#include "rw/bit_cast.cpp"
 
 constexpr uint32_t kCollMagic = 0x4C4C4F43;
 
@@ -48,19 +49,19 @@ bool LoaderCOL::load(const std::string& path) {
         model->modelid = head.modelid;
 
         auto readFloat = [&]() {
-            auto f = (float*)d;
+            auto f = bit_cast<float>(*d);
             d += sizeof(float);
-            return *f;
+            return f;
         };
         auto readU8 = [&]() {
-            auto f = (uint8_t*)d;
+            auto f = bit_cast<uint8_t>(*d);
             d += sizeof(uint8_t);
-            return *f;
+            return f;
         };
         auto readU32 = [&]() {
-            auto f = (uint32_t*)d;
+            auto f = bit_cast<uint32_t>(*d);
             d += sizeof(uint32_t);
-            return *f;
+            return f;
         };
         auto readVec3 = [&]() {
             auto x = readFloat();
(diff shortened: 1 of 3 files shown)


rwengine/src/loaders/LoaderGXT.cpp
```diff
@@ -8,20 +8,21 @@
...
-    std::uint32_t blocksize = *(std::uint32_t *)data;
+    std::uint32_t blocksize = bit_cast<std::uint32_t>(*data);
 
     data += 4;
 
     auto tdata = data + blocksize + 8;
 
     for (size_t t = 0; t < blocksize / 12; ++t) {
-        size_t offset = *(std::uint32_t *)(data + (t * 12 + 0));
+        size_t offset = bit_cast<std::uint32_t>(data + (t * 12 + 0));
```
JayFoxRox: Aren't semantics changed here? I'm not familiar with the `bit_cast` function, but above you dereference data and pass that to the `bit_cast`. Here, you just calculate the address, don't dereference it and pass it to `bit_cast`

rwengine/src/loaders/LoaderCOL.cpp
```diff
@@ -48,19 +49,19 @@ bool LoaderCOL::load(const std::string& path) {
         model->modelid = head.modelid;
 
         auto readFloat = [&]() {
-            auto f = (float*)d;
+            auto f = bit_cast<float>(*d);
```
JayFoxRox: nit: `d` is `auto`, so if it's suddenly not `sizeof(float)` it *looks* like you've changed semantics. This is actually poor design in the `bit_cast` function in my opinion.

rwengine/src/loaders/LoaderGXT.cpp
ShFil119: You're right, I'll rebase.

(Casting ptr and dereferencing can cause UB, if padding is different. bit_cast handles it)

JayFoxRox: nit: who thought it was a good idea to do (presumably) pointer arithmetic on an `auto` variable?

rwengine/src/loaders/LoaderCOL.cpp
ShFil119: I'll add better docs.

rwengine/src/loaders/LoaderGXT.cpp
paulsapps: This whole thing seems a bit smelly IMO.. would it not be better to have a BinaryStream object or something that you can Seek/Read(T) etc instead?

ShFil119: Do you think about treating it as stream?

ShFil119:
I'll close it for now, I'll try later convert BinaryStream to safer ifstream.

PR #372, opened 2018-01-16
FFmpeg sdt sound

ShFil119:
There's still much work to do. But maybe someone wants to test it. ;)
Comments are welcome. 
- [ ] Cleanup
- [ ] Doppler effect - grab position from center
- [ ] Memory management
- [ ] Tests   
@danhedron, should I move code implementing binaural audio mode in another pull?

PR #420, opened 2018-04-16
[Read Only]Sfx sounds

ShFil119:
A lot of changes after #372.
To do: 
- [x]  fix memory leaks, It's hard to determine when should/can I remove bd and ioBuffer.
- [x] ScriptSoundType isn't sound's position in sfx file, so sounds are mixed,
- [x] fix build,
- [x] setting type ScriptSound(opcodes 018d, 018e),
- [x] drop deprecated methods,
- [x] cleanup, div code in commits (done?),
- [x] stl algorithms 
- [x] drop av_register_all()
- [x] opcode 03d7
- [x] remove mutexes,
- [ ] cleanup 2.
- [x] rwfs::path

To do (after?)
- [ ] updating listener position,
- [ ] RE calculating sound volume,
- [ ] some continuous sounds have break time (how long?), 
- [ ] resampling sfx.

As always feedback and criticism are welcome.

#edit1 implemented hardcoded table for translating between script and sfx numeration.
#edit2 plugged leaks.
#edit3 fixed build system.

### DIFF
--- a/rwengine/src/audio/Sound.hpp
+++ b/rwengine/src/audio/Sound.hpp
@@ -1,0 +1,125 @@
+#ifndef _RWENGINE_SOUND_HPP_
+#define _RWENGINE_SOUND_HPP_
+
+#include <memory>
+#include <vector>
+#include <string>
+#include <mutex>
+
+#include <al.h>
+#include <alc.h>
+#include <glm/glm.hpp>
+
+#include "loaders/LoaderSDT.hpp"
+#include "rw/defines.hpp"
+
+/// Opaque for raw sound,
+/// cooperate with ffmpeg
+/// (loading and decoding sound)
+class SoundSource {
+    friend class SoundManager;
+    friend class SoundBuffer;
+
+public:
+    /// Load sound from mp3/wav file
+    void loadFromFile(const std::string& filename);
+
+    /// Load sound from sdt file
+    void loadSfx(const std::string& path, const size_t& index);
+
+private:
+    /// Raw data
+    std::vector<int16_t> data;
+
+    size_t channels;
+    size_t sampleRate;
+};
+
+/// OpenAL tool for playing
+/// sound instance.
+class SoundBuffer {
+    friend class SoundManager;
+
+public:
+    SoundBuffer();
+    bool bufferData(SoundSource& soundSource);
+
+    bool isPlaying() const;
+    bool isPaused() const;
+    bool isStoped() const;
+
+    void play();
+    void pause();
+    void stop();
+
+    void setPosition(const glm::vec3 position);
+    void setLooping(const bool  looping);
+    void setPitch(const float  pitch);
+    void setGain(const float  gain);
+    void setMaxDistance(const float  maxDist);
+
+private:
+    ALuint source;
+    ALuint buffer;
+};
+
+/// Wrapper for SoundBuffer and SoundSource.
+/// Each command connected
+/// with playment is passed to SoundBuffer
+struct Sound {
+    std::unique_ptr<std::mutex> soundMutex = std::make_unique<std::mutex>(); // mutexes can't be copied
+    size_t id = 0;
+    std::shared_ptr<SoundSource> source = nullptr;
+    std::unique_ptr<SoundBuffer> buffer = nullptr;
+    bool isLoaded = false;
+
+    Sound() = default;
+
+    bool isPlaying() const {
+        return buffer->isPlaying();
+    }
+
+    bool isPaused() const {
+        return buffer->isPaused();
+    }
+    bool isStoped() const {
+        return buffer->isStoped();
+    }
+
+    void play()  {
+        buffer->play();
+    }
+
+    void pause() {
+        buffer->pause();
+    }
+
+    void stop() {
+        buffer->stop();
+    }
+
+    void setPosition(const glm::vec3 position) {
+        buffer->setPosition(position);
+    }
+
+    void setLooping(const bool  looping) {
+        buffer->setLooping(looping);
+    }
+
+    void setPitch(const float  pitch) {
+        buffer->setPitch(pitch);
+    }
+
+    void setGain(const float  gain) {
+        buffer->setGain(gain);
+    }
+
+    void setMaxDistance(const float  maxDist) {
+        buffer->setMaxDistance(maxDist);
+    }
+
+    int getScriptObjectID() const {
+        return id;
+    }
+};
+#endif
--- a/rwengine/src/audio/Sound.hpp
+++ b/rwengine/src/audio/Sound.hpp
@@ -12,6 +12,7 @@
 
 #include "loaders/LoaderSDT.hpp"
 #include "rw/defines.hpp"
+#include <rw/filesystem.hpp>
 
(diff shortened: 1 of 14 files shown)


rwlib/source/loaders/LoaderSDT.cpp
```diff
@@ -116,6 +117,46 @@ char* LoaderSDT::loadToMemory(size_t index, bool asWave) {
         return nullptr;
 }
 
+static int read_packet(void *opaque, uint8_t *buf, int buf_size) {
```
darkf: Can you make `opaque` a more useful name?

darkf:
You don't use `loadSound` yet so I'm assuming this is still majorly WIP, but make sure its returned context is freed -- I wonder if it would not be better to use an RAII wrapper anyway.

ShFil119:
Please check: https://github.com/rwengine/openrw/pull/420/files#diff-e9cc6b7c0e9b40ae49ea18ed92518bfcR205
Using RAII is quite hard, all code of ffmpeg is written in C. So it's hard to balance between transparency and convenience of using. I mean AVFormatContext or custom RAII class.

rwlib/source/loaders/LoaderSDT.cpp
```diff
@@ -116,6 +117,46 @@ char* LoaderSDT::loadToMemory(size_t index, bool asWave) {
...
+
+    bool found = findAssetInfo(index, assetInfo);
+    if (!found || !raw_sound) {
+        RW_ERROR("Error loading sound");
+        return nullptr;
+    }
+    ioBuffer = reinterpret_cast<uint8_t*>(av_malloc(ioBufferSize)); /// Will be freeded in SoundManager after buffering
+
+    bd = std::make_unique<bufferData>();
+    av_register_all();
```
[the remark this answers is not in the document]
ShFil119: Yes (in ffmpeg 4.0), it's also unneeded. To remove.

rwlib/source/loaders/LoaderSDT.cpp
ShFil119: Api forces `buf_size` to to be int.

rwengine/src/audio/Sound.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    /// Load sound from sdt file
+    void loadSfx(const std::string& path, const size_t& index);
+
+private:
+    /// Raw data
+    std::vector<int16_t> data;
+
+    size_t channels;
+    size_t sampleRate;
```
Lihis: Why these two are type of `size_t`?

ShFil119: WaveHeader is using uint.

rwlib/source/loaders/LoaderSDT.cpp
ShFil119: About upaque, I don't have better idea.

ShFil119:
Two questions:
Should this pull also handle synchronization?

Should I prepare high level function for each opcode in SoundManager, without grabbing reference of sound? (Example of that high level function is playSound for 018c)

rwengine/src/audio/Sound.hpp
Lihis: You're right, I missed that.

rwengine/src/audio/Sound.hpp
```diff
@@ -0,0 +1,125 @@
...
+class SoundBuffer {
+    friend class SoundManager;
+
+public:
+    SoundBuffer();
+    bool bufferData(SoundSource& soundSource);
+
+    bool isPlaying() const;
+    bool isPaused() const;
+    bool isStoped() const;
```
husho: Misspelled?

ShFil119: yeah

darkf:
Synchronization of what, exactly?

Please leave opcode stuff to a follow-up PR as this is already large.

ShFil119:
Yeah, pull is huge I'll try to extract/backport refactoring of code. 
 
About synchronization, I was thinking about preparing ground for future resource manager. But it looks there's no need for now.

rwlib/source/loaders/LoaderSDT.cpp
madebr: is it a 'handle'?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -17,6 +17,8 @@
 #include <data/CutsceneData.hpp>
 
 #include <boost/algorithm/string/predicate.hpp>
+#include <string>
```
ShFil119: To remove.

rwengine/src/audio/SoundManager.hpp
```diff
@@ -1,80 +1,239 @@
...
 #include <vector>
+#include <glm/glm.hpp>
 
 #include <al.h>
 #include <alc.h>
 
+#include "audio/Sound.hpp"
+
+/// Script is using different numeration of sounds
+/// than postion index in sfx file.
```
ShFil119: position

rwengine/src/audio/SoundManager.hpp
```diff
@@ -1,80 +1,239 @@
...
+/// Also it is needed to store range of sound.
+/// Struct is used by opcodes 018c, 018d.
+struct SoundInstanceData {
+    int id;
+    int sfx;
+    int range;
+};
+/// Hardcoded array for translation script index into sfx index
```
ShFil119: translating

rwengine/src/audio/SoundManager.hpp
```diff
@@ -1,80 +1,239 @@
...
-        friend class SoundBuffer;
-
-    public:
-        void loadFromFile(const std::string& filename);
+    /// Setting position of listener for openAL.
+    void setListenerPosition(const glm::vec3 position);
 
-    private:
-        std::vector<int16_t> data;
+    /// Setting velocity of velocity for openAL.
```
ShFil119: of listener

rwlib/source/loaders/LoaderSDT.cpp
ShFil119: Yeah.

JayFoxRox: If this is a callback the way it's implemented is already a good choice.
There should probably be a comment stating that this is an `// ffmpeg XYZ callback` (whatever XYZ is).

However, when casting the opaque you could give a more descriptive name than `bd` (although it's somewhat self documenting due to the cast)

rwengine/src/audio/SoundManager.hpp
```diff
@@ -1,6 +1,7 @@
 #ifndef _RWENGINE_SOUNDMANAGER_HPP_
 #define _RWENGINE_SOUNDMANAGER_HPP_
 
+#include <algorithm>
```
madebr: You should include only what you use in that file. So `algorithm` is better included in the cpp file.

ShFil119: IIRC some small function in headers are using it.

ShFil119:
Everything useful has been extracted. ;)

PR #422, opened 2018-04-18
Added transparency for pickups

NFSMONSTR:
How does it look

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -347,7 +347,11 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
     std::sort(renderList.begin(), renderList.end(),
               [](const Renderer::RenderInstruction& a,
                  const Renderer::RenderInstruction& b) {
-                  return a.sortKey < b.sortKey;
+                    if (!a.drawInfo.blend && b.drawInfo.blend) return true;
+                    if (a.drawInfo.blend && !b.drawInfo.blend) return false;
+                    if (a.sortKey < b.sortKey) return true;
+
+                    return false;
               });
     RW_PROFILE_END();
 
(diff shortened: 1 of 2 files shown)


NFSMONSTR:
I dont sure why AppVeyor build failed, because [here](https://ci.appveyor.com/project/NFSMONSTR/openrw/build/1.0.1) the same commit builds ok

ShFil119:
With pull is everything ok. ;)

We are using CDash for making stats from logs. At the end of a build CI (travis/appveyor) tries to upload, but sometimes connection fails.

NFSMONSTR:
Thank you for explanation

NFSMONSTR:
Sorry, it was an accidental
I just don't have enough experience with command line git :)

NFSMONSTR:
So what should I do? Try to reopen that PR or recreate it?

JayFoxRox:
See my comment. It's very possible that the original code in master is already fine, but the blendmode is / was misconfigured. I'd recommend checking the behaviour in GTA3 / VC and see what's supposed to happen.

This is the difference:
Mixing vs. Additive blending
Note that black / dark is invisible in additive mode. So the original code suggests that it's additive blending.

Screenshots like this with more overlap should help: 

Corona overlap

In case of additive blending the overlap will eventually turn completly white as more colors are being added to the pixel (rendering order does not matter). With normal alpha blending the overall result will be darker (and the rendering order has to depend on the distance to the camera, as each transparent object depends on the object behind it)

So if you line up to a lot of these weapon coronas and you look through them, you can tell wether it's additive or not depending on if it changes by the direction you look through them. Also the final color if there's hundreds of them. Ideally, however, you'd just use a graphics debugger like apitrace to look at the renderer state (which is very easy to do).
It's possible that this is different in GTA3 and VC though. So ideally document both.

---

If your change is still necessary (when it's actually normal alpha blending) then you can send a new PR. Ideally it also should contain a comment about GTA3 / VC differences.

Issue #380, opened 2018-01-30
Cutscenes continue playing through the game menu


vflyson:
When the `Esc` is pressed, the main cutscene continues to play instead of being put on pause.

PR #423, opened 2018-04-27
Fix pausing issues

Lihis:
Fixes issue #380 

Behavior described in the above issue was not limited only to cutscenes. In any case the game was paused, it "catch" up the time it was paused after unpausing. This can be noticed most easily by checking the in-game clock before pausing the game, waiting for a while and after unpausing the game clock fast forwarded regarding the time it was paused. This PR fixes that issue.


darkf:
Nevermind, #404 handles that.

LGTM. Thanks!

PR #433, opened 2018-05-01
Added transparency for pickups

NFSMONSTR:
It seems code of changing weapon corona color in master written to use additive blending
<p><details><summary>Also in GTA 3 and VC(at least in VC) using additive blending for weapon coronas(clickable)</summary><p>

 ---
  So there is a directx code that i got with MS PIX(part of directx sdk):
  image
  As you can see GTA VC is using additive blending
  Also you might notice that in one moment corona look the same as in GTA 3, so we can make a conclusion that in it used almost the same render code:
  image

  [Here](http://www.mediafire.com/file/f38uonxzkr8c3bj/VC.PIXRun) is output of PIX for more detail researching

 Also I tried to do the same for GTA 3 but all my tries(I tried to use MS PIX and apitrace) were unsuccessful

 ---

  </p></details></p>

So I added choosing of blending mode in setBlend and DisplayParameters: default(alpha-blending) and additive blending
But here are some problems:
1. Almost everywhere used DisplayParameters and setBlend, but in some places instead of they direcrly used glEnable(GL_BLEND) that causing problems with adding multiply blend modes
2. Also in some places render code expects that alpha-blending enabled
3. Also we have OpenGLRenderer class so I think is better to place all OpenGL-dependend code into it and add some wrappers to use it in GameRenderer, MapRenderer and other classes instead of using direct OpenGL calls, because if later we will decide to add other graphical backend it will create some additional difficulties(May be should add this to issues as improvement?)

To do:
- [x] Fix render bugs caused by 1 and 2

___

Special thanks to @JayFoxRox for [help](https://github.com/rwengine/openrw/pull/422#issuecomment-385413791)

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -341,7 +341,6 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
     culled += objectRenderer.culled;
     renderer->pushDebugGroup("Objects");
     renderer->pushDebugGroup("RenderList");
-
     // Also parallelizable
     RW_PROFILE_BEGIN("Sort");
     // Earlier position in the array means earlier object's rendering
@@ -349,12 +348,11 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
     std::sort(renderList.begin(), renderList.end(),
               [](const Renderer::RenderInstruction& a,
                  const Renderer::RenderInstruction& b) {
-                    if (!a.drawInfo.blend && b.drawInfo.blend) return true;
-                    if (a.drawInfo.blend && !b.drawInfo.blend) return false;
+                    if (a.drawInfo.blendMode==BLEND_NONE && b.drawInfo.blendMode!=BLEND_NONE) return true;
+                    if (a.drawInfo.blendMode!=BLEND_NONE && b.drawInfo.blendMode==BLEND_NONE) return false;
                     return (a.sortKey > b.sortKey);
               });
     RW_PROFILE_END();
-
     RW_PROFILE_BEGIN("Draw");
     renderer->drawBatched(renderList);
     RW_PROFILE_END();
@@ -519,8 +517,7 @@ void GameRenderer::renderEffects(GameWorld* world) {
         dp.colour = glm::u8vec4(particle.colour * 255.f);
         dp.start = 0;
         dp.count = 4;
-        dp.blend = true;
-        dp.blendMode = 1;
+        dp.blendMode = BLEND_ADDITIVE;
         dp.diffuse = 1.f;
 
         renderer->drawArrays(transformMat, &particleDraw, dp);
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -348,8 +348,8 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
     std::sort(renderList.begin(), renderList.end(),
               [](const Renderer::RenderInstruction& a,
                  const Renderer::RenderInstruction& b) {
-                    if (a.drawInfo.blendMode==BLEND_NONE && b.drawInfo.blendMode!=BLEND_NONE) return true;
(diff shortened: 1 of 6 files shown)


rwengine/src/render/OpenGLRenderer.hpp
```diff
@@ -75,6 +75,8 @@ class Renderer {
         Textures textures;
         /// Alpha blending state
         bool blend;
+        /// Blending mode 0 - glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA); 1 - glBlendFunc(GL_ONE, GL_ONE);
+        size_t blendMode;
```
JayFoxRox: Please create a new enum if we are going to support multiple blending modes. We could even consider removing the bool for blending and using the enum values to disable blending.

NFSMONSTR: Good idea. Thank you

rwengine/src/render/OpenGLRenderer.hpp
```diff
@@ -341,38 +347,34 @@ class OpenGLRenderer : public Renderer {
...
+                case BLEND_NONE:
+                    glDisable(GL_BLEND);
+                    break;
+                case BLEND_ALPHA:
+                    glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+                    break;
+                case BLEND_ADDITIVE:
+                    glBlendFunc(GL_ONE, GL_ONE);
+                    break;
+
```
JayFoxRox: I'd rather see

```c
default:
  assert(false);
  break;
```

We'll have undefined behaviour anyway in that case - so might as well catch it in the debugger

rwengine/src/render/OpenGLRenderer.hpp
```diff
@@ -53,6 +53,15 @@ struct VertexP3 {
     VertexP3() = default;
 };
 
+/**
+ * Enum used to determine which blending mode to use
+ */
+enum BlendMode {
```
JayFoxRox: We use modern C++. Consider `enum class`

JayFoxRox:
I've only skimmed over the code, and aside from what I've noted, this LGTM.
Thanks for taking the time to use PIX and all - greatly appreciated :)

rwengine/src/render/GameRenderer.cpp
```diff
@@ -341,20 +341,18 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
...
-
     // Also parallelizable
     RW_PROFILE_BEGIN("Sort");
     std::sort(renderList.begin(), renderList.end(),
               [](const Renderer::RenderInstruction& a,
                  const Renderer::RenderInstruction& b) {
-                    if (!a.drawInfo.blend && b.drawInfo.blend) return true;
-                    if (a.drawInfo.blend && !b.drawInfo.blend) return false;
+                    if (a.drawInfo.blendMode==BLEND_NONE && b.drawInfo.blendMode!=BLEND_NONE) return true;
+                    if (a.drawInfo.blendMode!=BLEND_NONE && b.drawInfo.blendMode==BLEND_NONE) return false;
```
JayFoxRox: Interestingly, we could probably even use the actual value of the blendMode as the sort-key now (or a lookup table to re-order the blendMode values in case we re-order them).

Please do *not* change this in this PR. It's just something I wanted to note for the future.
Might be an idea to create an issue about this.

NFSMONSTR:
I dont sure what to do with code like this:
(GameRenderer.cpp)
```c++

void GameRenderer::drawTexture(TextureData* texture, glm::vec4 extents) {
    glUseProgram(ssRectProgram);
    ...
    glEnable(GL_BLEND);
    glUniform2f(ssRectOffset, extents.x, extents.y);
    glUniform2f(ssRectSize, extents.z, extents.w);
    ...

```

It uses OpenGL calls directly instead of calling Renderer class
Also it just enables GL_BLEND expecting that alpha blending already enabled

So what should I do in this PR?
Should I rewrite it to use Renderer class or just add `glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);`?

ShFil119:
Using rebase you won't have extra commits. ;)

NFSMONSTR:
Thank you. I'll try to do that

PR #434, opened 2018-05-03
Implemented some locate opcodes

ShFil119:
Original author: @husko
Ref: #408

### DIFF
--- a/scripts/docker/fedora_latest.docker
+++ b/scripts/docker/fedora_latest.docker
@@ -1,4 +1,4 @@
-FROM fedora:branched
+FROM fedora:rawhide
 
 # ffmpeg-devel needs rpmfusion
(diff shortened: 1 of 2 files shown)


scripts/docker/fedora_latest.docker
```diff
@@ -1,4 +1,4 @@
-FROM fedora:branched
+FROM fedora:rawhide
```
JayFoxRox: Is this supposed to be here?

ShFil119: Nope, branch was outdated.

JayFoxRox:
Just for the record: The official development channel is GitHub.

To be more open to new users, you should mirror important IRC conversations on GitHub, so even people who don't use IRC can join the development discussions and understand decisions.

IRC is good for quick discussions, but anything which should be on record, should be posted here, too.
Whenever you close a PR there should be a small statement (here, on GitHub) so other people know what's going on.

PR #411, opened 2018-03-12
Implemented various opcodes

husho:

PR #435, opened 2018-05-03
Implemented various opcodes.

dracc:
As written by @husho in #411

Edit: Modified according to discussion, code style and how the original game works.
Removed placeholder opcode descriptions ("Description") as function names were rather descriptive.

### DIFF
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -209,6 +209,7 @@ glm::vec3 CharacterObject::updateMovementAnimation(float dt) {
         if (it != movementAnimation->bones.end()) {
             auto rootBone = it->second;
             float step = dt;
+            RW_CHECK(animator->getAnimation(AnimIndexMovement), "Failed to read animation using index " << AnimIndexMovement);
             const float duration =
                 animator->getAnimation(AnimIndexMovement)->duration;
             float animTime =


rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -1762,8 +1787,7 @@ void opcode_0097(const ScriptArguments& args, ScriptFloat& arg1L) {
     @arg arg1G 
 */
 void opcode_0098(const ScriptArguments& args, ScriptFloat& arg1G) {
-    RW_UNIMPLEMENTED_OPCODE(0x0098);
-    RW_UNUSED(arg1G);
+    arg1G = static_cast <float> (std::rand()) / static_cast <float> (RAND_MAX);
```
ShFil119: We should use c++11 random.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -1774,8 +1798,7 @@ void opcode_0098(const ScriptArguments& args, ScriptFloat& arg1G) {
     @arg arg1G 
 */
 void opcode_0099(const ScriptArguments& args, ScriptInt& arg1G) {
-    RW_UNIMPLEMENTED_OPCODE(0x0099);
-    RW_UNUSED(arg1G);
+    arg1G = std::rand() % 65536;
```
ShFil119: Here also we should use c++11 random.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6046,12 +6045,9 @@ bool opcode_0207(const ScriptArguments& args, const ScriptCharacter character, c
...
     @arg arg3 
 */
-void opcode_0208(const ScriptArguments& args, const ScriptFloat arg1, const ScriptFloat arg2, ScriptFloat& arg3) {
-    RW_UNIMPLEMENTED_OPCODE(0x0208);
-    RW_UNUSED(arg1);
-    RW_UNUSED(arg2);
-    RW_UNUSED(arg3);
+void opcode_0208(const ScriptArguments& args, const ScriptFloat low, const ScriptFloat high, ScriptFloat& random) {
     RW_UNUSED(args);
+    random = low + static_cast <float> (std::rand()) / (static_cast <float> (RAND_MAX / (high - low)));
```
ShFil119: Here also we should use c++11 random.

dracc: You are totally righ, but these three have since been taken care of in a tasty manner by @christoph-heiss in #439 ! :smile:

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -211,6 +213,26 @@ inline void addObjectToMissionCleanup(const ScriptArguments& args,
         args.getState()->missionObjects.push_back(object);
     }
 }
+
+inline void getClosestNode(const ScriptArguments& args, ScriptVec3& coord, AIGraphNode::NodeType type,
+                           ScriptFloat& xCoord, ScriptFloat& yCoord, ScriptFloat& zCoord) {
+    coord = script::getGround(args, coord);
+    float closest = 2000.f;
```
JayFoxRox: This feels kind of arbitrary.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2234,8 +2262,8 @@ bool opcode_00c2(const ScriptArguments& args, ScriptVec3 coord, const ScriptFloa
     opcode 00c3
 */
 void opcode_00c3(const ScriptArguments& args) {
-    RW_UNIMPLEMENTED_OPCODE(0x00c3);
-    RW_UNUSED(args);
+    args.getVM()->setDebugFlag(true);
+
```
JayFoxRox: What does this even do? Are we sure it should be part of the SCM interpreter / `ScriptMachine`?
Maybe this debug flag is for the gamestate instead? Is there any documentation for this?

rwengine/src/script/ScriptMachine.hpp
```diff
@@ -153,7 +164,8 @@ class ScriptMachine {
     SCMFile* file;
     ScriptModule* module;
     GameState* state;
-
+    bool debugFlag;
```
JayFoxRox: Are we sure this should be in `ScriptMachine`? Do we know that the opcodes don't modify a game-wide debug mode? Is there any documentation for this?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -4978,35 +4979,43 @@ void opcode_01b4(const ScriptArguments& args, const ScriptPlayer player, const S
...
+    args.getState()->basic.lastWeather = weatherID;
     args.getState()->basic.nextWeather = weatherID;
+    args.getState()->basic.forcedWeather = weatherID;
 }
 
 /**
-    @brief release_weather
+    @brief RELEASE_WEATHER
+
+    opcode 01B7 / 439
```
JayFoxRox: Again, unnecessary decimal + renamed opcode to something which is not giving any benefit. Please review other instances too.

rwengine/src/script/modules/GTA3ModuleImpl.inl
JayFoxRox: In that case, I'd suggest to remove these opcodes from this patch to avoid merge conflicts and changing back and forth.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6399,10 +6391,10 @@ void opcode_0223(const ScriptArguments& args, const ScriptCharacter character, c
     @arg vehicle Car/vehicle
     @arg arg2 
 */
-void opcode_0224(const ScriptArguments& args, const ScriptVehicle vehicle, const ScriptInt arg2) {
+void opcode_0224(const ScriptArguments& args, const ScriptVehicle vehicle, const ScriptFloat health) {
```
JayFoxRox: Urgh? This suggests that the SCM documentation we used for the opcode was bad. Please review or even unit-test this.
If you can't do this, please add an issue that we should review the type of this argument.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6413,11 +6405,9 @@ void opcode_0224(const ScriptArguments& args, const ScriptVehicle vehicle, const
     @arg player Player
     @arg arg2 
 */
-void opcode_0225(const ScriptArguments& args, const ScriptPlayer player, ScriptInt& arg2) {
-    RW_UNIMPLEMENTED_OPCODE(0x0225);
-    RW_UNUSED(player);
-    RW_UNUSED(arg2);
+void opcode_0225(const ScriptArguments& args, const ScriptPlayer player, ScriptFloat& health) {
```
JayFoxRox: Same as above: changing types can be dangerous.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6427,11 +6417,9 @@ void opcode_0225(const ScriptArguments& args, const ScriptPlayer player, ScriptI
     @arg character Character/ped
     @arg arg2 
 */
-void opcode_0226(const ScriptArguments& args, const ScriptCharacter character, ScriptInt& arg2) {
-    RW_UNIMPLEMENTED_OPCODE(0x0226);
-    RW_UNUSED(character);
-    RW_UNUSED(arg2);
+void opcode_0226(const ScriptArguments& args, const ScriptCharacter character, ScriptFloat& health) {
```
JayFoxRox: Another instance with Int to Float change. Please review other instances too

rwengine/src/script/ScriptFunctions.hpp
husho: IIRC, original game uses arbitrary value of something like 99999.f too.

JayFoxRox: If it does, we should figure it out (and document it), or choose a large enough value / always accept the first entry to set the baseline. I feel 2000.0f might be too small.

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: This is part of SCM interpreter. Though, this is disabled in original game and completely removed from Vice City. It is still functional if you patch 1 asm instruction. IIRC, it is used to draw boundaries for locate opcodes.

rwengine/src/script/ScriptMachine.hpp
husho: Read above.

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: Documentation seemed messy so I basically tried to mess around with it.

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: Yeah, not sure if is there any up to date proper documentation anywhere. Most popular sources like gtamodding.com or gtagmodding.com have mistakes.

rwengine/src/script/ScriptFunctions.hpp
husho: I think it should be within game map boundaries. Isn't III map 4000x4000, is it?

rwengine/src/script/modules/GTA3ModuleImpl.inl
JayFoxRox: Still weird that it's part of the `ScriptMachine` in my opinion.
Now that I know what it does, I think we should definitely implement it.

*(Although we should probably also expose it via the debug menu later, which could be a follow-up issue)*

rwengine/src/script/ScriptFunctions.hpp
JayFoxRox: Wouldn't it make more sense to use a value which is definitely *not* in the game bounds [, but way larger]?
if you are in the center (2000,2000), and you look for objects within a 2000 unit radius in a 4000x4000 square, then you won't be able to find objects near the corners of the game world.

husho: Yeah, sorry for misunderstanding. I meant that it should cover map boundaries.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -892,22 +892,47 @@ void opcode_0051(const ScriptArguments& args) {
...
+    RW_UNUSED(arg2);
+    RW_UNUSED(arg3);
+    RW_UNUSED(arg4);
+    RW_UNUSED(arg5);
+    RW_UNUSED(arg6);
+    RW_UNUSED(args);
+}
+
+/**
+    @brief CREATE_PLAYER index x y z outPlayerHandle
```
JayFoxRox: Please don't rename these. Issue still remains in the Cleanup commit.
Other instances too.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -892,22 +892,47 @@ void opcode_0051(const ScriptArguments& args) {
 }
 
 /**
-    @brief  %5d% = create_player %1o% at %2d% %3d% %4d%
+    @brief LINE
```
JayFoxRox: Consider giving this a new name, or avoid implementing this in the first place.
If the command is not used, there is no point in implementing (or even stubbing) it either *imo*.

husho: This one is unused.

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: '''
float dist  = std::abs(coord.x - node->position.x);
      dist += std::abs(coord.y - node->position.y);
      dist += std::abs(coord.z - node->position.z) * 3.f;

rwengine/src/script/ScriptFunctions.hpp
dracc: This value was found to be 10000.f in @husho 's research.

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -211,6 +213,26 @@ inline void addObjectToMissionCleanup(const ScriptArguments& args,
         args.getState()->missionObjects.push_back(object);
     }
 }
+
+inline void getClosestNode(const ScriptArguments& args, ScriptVec3& coord, AIGraphNode::NodeType type,
```
husho: Distance calculation should be changed to this (https://pastebin.com/X3hCXE05)
```
float dist  = std::abs(coord.x - node->position.x);
      dist += std::abs(coord.y - node->position.y);
      dist += std::abs(coord.z - node->position.z) * 3.f;
```

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -223,7 +223,9 @@ inline void getClosestNode(const ScriptArguments& args, ScriptVec3& coord, AIGra
     
     for (const auto &node : nodes) {
         if (node->type == type) {
-            float dist = glm::distance2(coord, node->position);
+            float dist = std::abs(coord.x - node->position.x);
```
JayFoxRox: please add a comment that this is how the game does it (manhattan-distance and has to be 3x as close on the vertical axis to be considered)

dracc: Above the commented line or in the description?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -892,22 +892,26 @@ void opcode_0051(const ScriptArguments& args) {
 }
 
 /**
-    @brief  %5d% = create_player %1o% at %2d% %3d% %4d%
+    @brief %5d% = create_player %1o% at %2d% %3d% %4d%
```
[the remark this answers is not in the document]
dracc: With your comments in #438 in mind, it'd be best if I restored this space, right?

dracc:
I somehow managed to destroy my branch.
git and I were not friends today.
Will re-open based on new branch.

JayFoxRox:
This should almost never be necessary - unless you have previously closed the PR (as github then stops following the pushes / git history, so fixing it becomes nearly impossible)
In the future, you can just `git push --force` if you accidentally uploaded a wrong version (but please be careful when force pushing away code which already got reviews).
If you have trouble locally, check out `git reflog`.
Also make sure to reference this in your new PR *description* so people can see the full review log: `Continuation of #435`.

PR #436, opened 2018-05-03
Implemented blip opcodes

dracc:
As written by @husho in #409

### DIFF
--- a/rwengine/src/script/ScriptFunctions.hpp
+++ b/rwengine/src/script/ScriptFunctions.hpp
@@ -139,6 +139,39 @@ inline GameString gxt(const ScriptArguments& args, const ScriptString id) {
     return args.getWorld()->data->texts.text(id);
 }
 
+inline BlipData& createBlip(const ScriptArguments& args, const ScriptVec3& coord,
+                            BlipData::BlipType type) {
+    BlipData data;
+    data.coord = coord;
+    data.type = type;
+    switch (type) {
+        case BlipData::Contact:
+            data.colour = 2;
+            break;
+        case BlipData::Coord:
+            data.colour = 5;
+            break;
+        default:
+            RW_ERROR("Unhandled blip type");
+            break;
+    }
+    data.target = 0;
+    data.display = BlipData::ShowBoth;
+    data.texture = "";
+    data.size = 3;
+    auto blip = args.getState()->addRadarBlip(data);
+    return args.getState()->radarBlips[blip];
+}
+
+const char* getBlipSprite(ScriptRadarSprite sprite);
+
+inline BlipData& createBlipSprite(const ScriptArguments& args, const ScriptVec3& coord,
+                                  BlipData::BlipType type, int sprite) {
+    auto& data = script::createBlip(args, coord, type);
+    data.texture = getBlipSprite(sprite);
+    return data;
+}
+
 inline BlipData& createObjectBlip(const ScriptArguments& args,
                                   GameObject* object) {
     BlipData data;
@@ -160,7 +193,7 @@ inline BlipData& createObjectBlip(const ScriptArguments& args,
             data.colour = 6;  // @todo 4 in Vice City
             break;
         default:
-            data.type = BlipData::Location;
+            data.type = BlipData::None;
             RW_ERROR("Unhandled blip type");
             break;
     }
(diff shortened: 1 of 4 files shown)


rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -139,32 +139,23 @@ inline GameString gxt(const ScriptArguments& args, const ScriptString id) {
...
             break;
-        case GameObject::Pickup:
-            data.type = BlipData::Pickup;
-            data.colour = 6;  // @todo 4 in Vice City
-            break;
-        case GameObject::Instance:
-            data.type = BlipData::Instance;
-            data.colour = 6;  // @todo 4 in Vice City
+        case BlipData::Coord:
+            data.colour = 5;
```
JayFoxRox: nit: The Vice City comments were usefull in my opinion.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -3878,31 +3878,53 @@ void opcode_0160(const ScriptArguments& args, ScriptVec3 coord, const ScriptChan
...
 /**
-    @brief %4d% = create_marker_above_actor %1d% color %2d% visibility %3d%
+    @brief ADD_BLIP_FOR_CHAR_OLD pedHandle colour display outblipHandle
 
-    opcode 0162
-    @arg character Character/ped
-    @arg arg2 
-    @arg arg3 
-    @arg blip Blip
+    opcode 0162 / 354
```
JayFoxRox: The decimal is very confusing here

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -3878,31 +3878,53 @@ void opcode_0160(const ScriptArguments& args, ScriptVec3 coord, const ScriptChan
...
+    auto& data = script::createObjectBlip(args, vehicle);
+    data.colour = colour;
+    /// @todo change ScriptBlipDisplay to BlipData::DisplayMode
+    data.display = static_cast <BlipData::DisplayMode> (display);
+    blip = &data;
 }
 
 /**
-    @brief %4d% = create_marker_above_actor %1d% color %2d% visibility %3d%
+    @brief ADD_BLIP_FOR_CHAR_OLD pedHandle colour display outblipHandle
```
JayFoxRox: Is this change really useful? The comment on the left has been auto-generated. This change is fine if we did this renaming anyway.

However, I'm also worried *this* might even be the R* function name, which we tried to avoid for legal reasons.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -3878,31 +3878,53 @@ void opcode_0160(const ScriptArguments& args, ScriptVec3 coord, const ScriptChan
...
-    @arg arg3 
-    @arg blip Blip
+    opcode 0162 / 354
+
+    Creates and attaches blip for an ped
+
+    @arg pedHandle Ped
+    @arg colour Blip colour
+    @arg display Blip display mode
+    @arg outblipHandle Created blip
```
JayFoxRox: The `@arg` was previously easily parsable by a computer. Now it's not. Check other instances and how we handled this in the past.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -3878,31 +3878,53 @@ void opcode_0160(const ScriptArguments& args, ScriptVec3 coord, const ScriptChan
...
+    data.colour = colour;
+    /// @todo change ScriptBlipDisplay to BlipData::DisplayMode
+    data.display = static_cast <BlipData::DisplayMode> (display);
+    blip = &data;
+}
+
+/**
+    @brief ADD_BLIP_FOR_OBJECT_OLD objectHandle colour display outblipHandle
+
+    opcode 0163 / 355
```
JayFoxRox: For this comment block, see other instance above.
Also take care of other instances which follow.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -11717,8 +11716,8 @@ void opcode_03dc(const ScriptArguments& args, const ScriptPickup pickup, ScriptB
     @arg arg2 
     @arg blip 
 */
-void opcode_03dd(const ScriptArguments& args, const ScriptPickup pickup, const ScriptRadarSprite arg2, ScriptBlip& blip) {
-    auto data = script::createObjectBlipSprite(args, pickup, arg2);
+void opcode_03dd(const ScriptArguments& args, const ScriptPickup pickup, const ScriptRadarSprite blipSprite, ScriptBlip& blip) {
```
JayFoxRox: Argument renamed but unchanged in comment. Please review other instances too.

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: In fact, it is original R* function name.

rwengine/src/script/ScriptFunctions.hpp
husho: Ugh, it was a long time ago, so I don't really remember what was the purpose of this change. BUT I think there was missing bilp type. Although, these values are from original game.

dracc:
Great input @jayfoxrox and @husho. Thanks!
I'll look into it ASAP.

rwengine/src/script/modules/GTA3ModuleImpl.inl
dracc: I'm not sure I follow.
Was the name "argX" (where X is a number) easier to parse, or what is the actual difference?

JayFoxRox: Take my commnt as an example on how *not* to write a review comment. I don't know what I meant myself.
But I *think* it has to do with the actual argument being called `blip` when this one refers to `outblipHandle`

rwengine/src/script/ScriptFunctions.hpp
dracc: It seems a function was added and the "deleted" values have been revised and git has, instead of being easy-to-read chosen to be efficient and reuse as many lines as possible. Looks a mess, but if one reads more code it stands clear that nothing was really lost, just re-organized. :)

dracc:
Comments have been addressed and taken care of.

PR #412, opened 2018-03-12
Improved mission clean up logic and implemented related opcodes

husho:

PR #437, opened 2018-05-03
[Ready] Improved mission clean up logic and implemented related opcodes

dracc:
As written by @Husho in #412

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -1789,7 +1789,11 @@ void opcode_009a(const ScriptArguments& args, const ScriptPedType pedType, const
 
     coord = script::getGround(args, coord);
     character = args.getWorld()->createPedestrian(model, coord + script::kSpawnOffset);
-    script::addObjectToMissionCleanup(args, character);
+    character->setLifetime(GameObject::MissionLifetime);
+
+    if (args.getThread()->isMission) {
+        script::addObjectToMissionCleanup(args, character);
+    }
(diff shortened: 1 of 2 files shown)


dracc:
LGTM.
No weird comments and the logic seems, well, logical.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -10696,23 +10728,27 @@ void opcode_0394(const ScriptArguments& args, const ScriptInt arg1) {
     @arg clearParticles Boolean true/false
 */
 void opcode_0395(const ScriptArguments& args, ScriptVec3 coord, const ScriptFloat radius, const ScriptBoolean clearParticles) {
-    GameWorld* gw = args.getWorld();
+    auto gw = args.getWorld();
```
husho: this should be a pointer

dracc: Seems to me I ought to revert it back to `GameWorld* gw`. Do you agree?

husho: yes

dracc:
Opcode 02dd looks like it was removed and rewritten with the same code. Reason for this is that it mixed tabs and spaces and looked horrid when using other tab-width than the original authors.
I'm sure there are more instances of this, but I stumbled upon this one and figured I "might as well".

dracc:
Hmm... This one's a little odd.
<a href="https://i.imgur.com/SZJCH9p.png">
  
</a>

PR #438, opened 2018-05-03
[Ready] Implemented some locate opcodes

dracc:
As written by @husho in #408


dracc:
Placeholders for ScriptVec3, would that (if it's the third argument) be %3d% or %3o%?

husho:
neither, o is for models and d is for numbers

dracc:
That sounds sane. But what would be the right one?

husho:
v2, v3? but then positions would misalign

dracc:
@jayfoxrox clarified on irc that the placeholders are generated, so I will leave them be (mostly).

dracc:
So, instead of "mostly", it would be preferable to not touch them at all?
What about the `@arg` lines, are those only related to OpenRW?
Great write-up, thank you @jayfoxrox!

PR #439, opened 2018-05-03
Implement random number opcodes [0098, 0099, 0208, 0209].

christoph-heiss:
Hi!

Not much to say, really. Implemented using C++ std::{random_device,mt19937}.

I originally wrote this about a month ago or two, and thought I would just PR this now - I do have finals at the moment and not really a lot of time. And before this code rots ..

Greets!


darkf:
LGTM, thanks.

PR #440, opened 2018-05-05
Make buoys float again

mole99:
Until now buoys would just fall down, we need to apply the force with an impulse to make it work. To update the objects in debug mode, a tickPhysics function had to be created, like the vehicles already use.

### DIFF
--- a/rwengine/src/objects/InstanceObject.cpp
+++ b/rwengine/src/objects/InstanceObject.cpp
@@ -46,6 +46,11 @@ InstanceObject::InstanceObject(GameWorld* engine, const glm::vec3& pos,
 InstanceObject::~InstanceObject() = default;
 
 void InstanceObject::tick(float dt) {
+    RW_UNUSED(dt);
+    // Moved to tickPhysics
+}
+
+void InstanceObject::tickPhysics(float dt) {
     if (animator) animator->tick(dt);
 
     if (!body || !dynamics) {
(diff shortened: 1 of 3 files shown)


ShFil119:
After some testing, it works fine for buoys.

As street lights are also InstanceObjects doesn't it cause side effects?
screenshot_20180506_004749
Actually I don't have physical access to windows. :(
 Does street lights also float on water?

mole99:
You are right, I haven't even thought about that!

With this change only buoys will float on water, later on we can add other objects which should float, but the issue for now was only concerning the buoys.

Unfortunately I had to hardcode the model name, but this already happens with the IslandLOD and should be therefore okay.

I do my best to improve this project ;)

rwengine/src/objects/InstanceObject.cpp
```diff
@@ -64,57 +75,62 @@ void InstanceObject::tick(float dt) {
...
-        if (hI < NO_WATER_INDEX) {
-            wH = engine->data->waterHeights[hI];
-            wH += engine->data->getWaveHeightAt(ws);
-            inWater = vH <= wH;
-        } else {
-            inWater = false;
+
+    // Only certain objects should float on water
+    if ( floating )
+    {
```
ShFil119: Can you reformat this?

darkf:
Thanks

PR #447, opened 2018-05-09
[Ready] Perfect forwarding and some forgotten stuff

ShFil119:
I forgot about this branch. ;)
#edit, now it should it be mergeable.

### DIFF
--- a/rwengine/src/TextureArchive.cpp
+++ b/rwengine/src/TextureArchive.cpp
@@ -46,7 +46,7 @@ std::unique_ptr<TextureArchive> TextureArchive::create(
                    bufSize);
         }
 
-        textureArchive->textures.push_back(std::move(texture));
+        textureArchive->textures.push_back(texture);
 
         section = section->next;  // Extension
     }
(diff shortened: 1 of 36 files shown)


rwengine/src/TextureArchive.cpp
```diff
@@ -46,7 +46,7 @@ std::unique_ptr<TextureArchive> TextureArchive::create(
                    bufSize);
         }
 
-        textureArchive->textures.push_back(std::move(texture));
+        textureArchive->textures.push_back(texture);
```
husho: Why remove ```std::move``` here

rwengine/src/loaders/LoaderIDE.cpp
```diff
@@ -237,7 +237,7 @@ bool LoaderIDE::load(const std::string &filename, const PedStatsList &stats) {
                         getline(buffstream, buff, ',');
                         node.other_thing2 = atoi(buff.c_str());
 
-                        path.nodes.push_back(std::move(node));
+                        path.nodes.push_back(node);
```
husho: Why remove ```std::move``` here

rwengine/src/TextureArchive.cpp
ShFil119: Trivially copyable.

rwengine/src/loaders/LoaderIDE.cpp
ShFil119: Also trivially copyable.

PR #448, opened 2018-05-10
[Ready] Vehicle health

husho:
Minor changes, but this should make more missions playable, because now it's possible to "destroy" vehicles.

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -4327,7 +4327,7 @@ bool opcode_0184(const ScriptArguments& args, const ScriptCharacter character, c
 */
 bool opcode_0185(const ScriptArguments& args, const ScriptVehicle vehicle, const ScriptInt arg2) {
     RW_UNUSED(args);
-    return vehicle->getHealth() <= arg2;
+    return vehicle->getHealth() > arg2;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -4324,11 +4326,8 @@ bool opcode_0184(const ScriptArguments& args, const ScriptCharacter character, c
     @arg arg2 
 */
 bool opcode_0185(const ScriptArguments& args, const ScriptVehicle vehicle, const ScriptInt arg2) {
-    RW_UNIMPLEMENTED_OPCODE(0x0185);
-    RW_UNUSED(vehicle);
-    RW_UNUSED(arg2);
     RW_UNUSED(args);
-    return true;
+    return vehicle->getHealth() <= arg2;
```
danhedron: This comparison should be ``>`` based on the description of the opcode.

husho: https://hastebin.com/pixogesato

husho: So I decided to recheck this opcode in scm and logically it seems that it should be ```>```
Lesson learned - don't trust hex-rays decompiler. Sadly I can't really make sense out of FPU instructions, here's disassembly if someone's want to verify https://pastebin.com/raw/EQGyCKTB

PR #450, opened 2018-05-12
[Ready] Fixed script objects falling through the ground

husho:
This fixes issue when during the game intro few objects fall through the ground.

- [x] 0107 opcode
- [x] 0392 opcode​

Issues to fix:
- [x] Lamposts act weirdly now, seems like they only fall down after 2nd hit
- [x] Hydrants don't longer break

### DIFF
--- a/rwengine/src/render/ObjectRenderer.cpp
+++ b/rwengine/src/render/ObjectRenderer.cpp
@@ -35,10 +35,9 @@ constexpr float kMagicLODDistance = 330.f;
 constexpr float kVehicleLODDistance = 70.f;
 constexpr float kVehicleDrawDistance = 280.f;
 
-RenderKey createKey(float normalizedDepth,
-                    Renderer::Textures& textures) {
+RenderKey createKey(float normalizedDepth, Renderer::Textures& textures) {
     return (uint32_t(0x7FFFFF * normalizedDepth) << 8 |
-           uint8_t(0xFF & (!textures.empty() ? textures[0] : 0)));
+            uint8_t(0xFF & (!textures.empty() ? textures[0] : 0)));
 }
 
 void ObjectRenderer::renderGeometry(Geometry* geom,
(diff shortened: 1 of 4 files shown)


mythrillex:
After the fix, some hydrants dont break when u hit, its normal ?

husho:
I'll test

husho:
What you described happens in master branch, in my branch hydrants don't even break

rwengine/src/render/ObjectRenderer.cpp
```diff
@@ -308,7 +308,8 @@ void ObjectRenderer::renderVehicle(VehicleObject* vehicle,
 void ObjectRenderer::renderPickup(PickupObject* pickup, RenderList& outList) {
     if (!pickup->isEnabled()) return;
 
-    glm::mat4 modelMatrix = glm::translate(glm::mat4(1.0f), pickup->getPosition());
+    glm::mat4 modelMatrix =
+        glm::translate(glm::mat4(1.0f), pickup->getPosition());
```
danhedron: Looks like there's just formatting changes in this file

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -10395,10 +10396,8 @@ void opcode_0381(const ScriptArguments& args, const ScriptObject object, ScriptV
     @arg arg2 Boolean true/false
 */
 void opcode_0382(const ScriptArguments& args, const ScriptObject object, const ScriptBoolean arg2) {
-    RW_UNIMPLEMENTED_OPCODE(0x0382);
-    RW_UNUSED(object);
-    RW_UNUSED(arg2);
     RW_UNUSED(args);
+	object->setStatic(!arg2);
```
danhedron: The description of this opcode seems to be that it turns collision detection on and off, not change static status.

rwengine/src/render/ObjectRenderer.cpp
husho: getVisible was placed with isVisible

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: You right. Will fix.

PR #464, opened 2018-05-16
Debug menu off by 1 error fix

husho:


rwgame/states/DebugState.cpp
```diff
@@ -311,7 +311,7 @@ std::shared_ptr<Menu> DebugState::createMissionsMenu() {
         "The Exchange",
     };
 
-    for (int i = 0; i < 79; ++i) {
+    for (int i = 0; i < 80; ++i) {
```
ShFil119: Maybe it's better to grab size of `w`?

ShFil119:
Btw `w` can be const. 
Also I'd replace c style array with std::array. (you can also change array in `createWeatherMenu` )

rwgame/states/DebugState.cpp
husho: Yeah

PR #465, opened 2018-05-16
Increase tickrate to 60 times per second

ShFil119:
It should reduce level of perceptibility physics bugs and tearing non-interpolated objects.

### DIFF
--- a/rwgame/game.hpp
+++ b/rwgame/game.hpp
@@ -1,6 +1,6 @@
 #ifndef GAME_HPP
 #define GAME_HPP
 
-#define GAME_TIMESTEP (1.f / 30.f)
+#define GAME_TIMESTEP (1.f / 60.f)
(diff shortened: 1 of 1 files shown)


ShFil119:
Can you share how it looks like and your setup (hardware)?

SonofUgly:
I was using rivatuner to limit the framerate, but I can't recreate it so either something else had changed or I was mistaken, so nevermind about that.

dracc:
I tried it, albeit running with more or less constant 75fps.
Smooth as butter and car physics are suddenly manageable?
The Kumho didn't flip in the first corner at the end of the bridge and the Banshee only flips if you hit a curb at a bad angle.

ShFil119:
Cars can still flop if one of wheels sinks in fender.
fender

PR #466, opened 2018-05-16
[Ready]Deploy appveyor builds

ShFil119:
Not sure if we also want to deploy builds from pulls.


.appveyor.yml
```diff
@@ -38,3 +38,25 @@ init:
 
 build_script:
   - ctest -VV -S cmake/ctest/script_ci.ctest
+
+# scripts to run after build
+after_build:
+  - 7z a openrw-%appveyor_repo_branch%.zip %APPVEYOR_BUILD_FOLDER%\build\rwgame\Release\
```
husho: Consider adding ```*``` in the end of the path to get rid of unnecessary folders in archive

ShFil119:
@madebr suggested that it's good step to move setting CMAKE_INSTALL_PREFIX into ctest. 
I'm trying to do this. https://github.com/ShFil119/openrw/pull/35/files

Which version do you prefer?

husho:
I'm not familiar with cmake, so I can't speak for myself.

.appveyor.yml
```diff
@@ -38,11 +38,11 @@ init:
 
 build_script:
   - ctest -VV -S cmake/ctest/script_ci.ctest
+  - msbuild %APPVEYOR_BUILD_FOLDER%\build\rwgame\INSTALL.vcxproj
```
madebr: I believe calling `msbuild` explicitly can be changed to 
```
cmake --target install --config Release "%APPVEYOR_BUILD_FOLDER%\build"
```
(or something similar).
That would make it build system agnostic.

ShFil119: We want to install all or "only" rwgame?

darkf:
Is this ready?

madebr:
Not pr ready, but I've tried creating archives using CPack.

https://github.com/madebr/openrw/commit/02866b9b91f17c91c75de4ab2de2813ae936034c

`cpack -G TXZ`

ShFil119:
@madebr, do you want to include using cpack in this pull?

madebr:
CPack does not need any CMAKE_INSTALL_PREFIX fu.
It can also help create installers. (Qt IFW)

ShFil119:
Succeeded by #503.

PR #471, opened 2018-05-19
Temporary partial fix for duped cars & peds in first mission

husho:
Fixes duped cars and peds in first mission only if you skip cutscene. Watching whole intro movie still creates duplicates.

### DIFF
--- a/rwengine/src/script/modules/GTA3ModuleImpl.inl
+++ b/rwengine/src/script/modules/GTA3ModuleImpl.inl
@@ -2421,13 +2421,23 @@ bool opcode_00e0(const ScriptArguments& args, const ScriptPlayer player) {
     @arg player Player
     @arg buttonID Button ID
 */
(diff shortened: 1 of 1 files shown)


dracc:
I've tested it and it seems to do the trick. Although I don't fancy hacks, this might be a good compromise until we find what actually causes it.

darkf:
Instead of a hack I'd rather the real solution be implemented, unless it's super difficult.

husho:
It's good enough for now, because duped cars are very annoying.

husho:
Implementing controls system would take a while.

JayFoxRox:
The entire function is a hack, so I'm fine with adding another temporary hack to fix an issue which might scare people away from openrw during the first minutes of trying it.

I did not test this myself, but will trust @husho .

LGTM.

*(I'd merge this, but will honor @darkf request for @danhedron review)*

PR #473, opened 2018-05-19
Fixed: mission is not restarting if player is in vehicle

### LINKED ISSUE #469: Mission does not restart
Give Me Liberty and Luigi's Girls mission does not restart if you stay in vehicle after you fail it

husho:
Closes #469


dracc:
LGTM.

PR #474, opened 2018-05-19
Fixed loading ZoneCylinder texture

NFSMONSTR:
It seems like that texture for ZoneCylinder loaded correctly but `texturelookup ` can't find texture because wrong texture slot is set. So in `loadClump` we should set it manually

### DIFF
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -409,6 +409,15 @@ ClumpPtr GameData::loadClump(const std::string& name) {
     return m;
 }
 
+ClumpPtr GameData::loadClump(const std::string& name, const std::string& textureSlot) {
+    std::string currentSlot = currenttextureslot;
+    if (textureSlot.size()>0)
+        currenttextureslot = textureSlot;
+    ClumpPtr result = loadClump(name);
+    currenttextureslot = currentSlot;
+    return result;
+}
+
 void GameData::loadModelFile(const std::string& name) {
     auto file = index.openFilePath(name);
     if (!file) {
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -411,7 +411,7 @@ ClumpPtr GameData::loadClump(const std::string& name) {
 
 ClumpPtr GameData::loadClump(const std::string& name, const std::string& textureSlot) {
     std::string currentSlot = currenttextureslot;
-    if (textureSlot.size()>0)
+    if (!textureSlot.empty())
         currenttextureslot = textureSlot;
     ClumpPtr result = loadClump(name);
     currenttextureslot = currentSlot;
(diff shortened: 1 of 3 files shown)


rwengine/src/engine/GameData.cpp
```diff
@@ -409,6 +409,15 @@ ClumpPtr GameData::loadClump(const std::string& name) {
     return m;
 }
 
+ClumpPtr GameData::loadClump(const std::string& name, const std::string& textureSlot) {
+    std::string currentSlot = currenttextureslot;
+    if (textureSlot.size()>0)
```
ShFil119: Maybe better `!textureSlot.empty()`?

rwgame/RWGame.cpp
```diff
@@ -26,10 +26,10 @@
 #include <iomanip>
 #include <iostream>
 
-std::map<GameRenderer::SpecialModel, std::string> kSpecialModels = {
-    {GameRenderer::ZoneCylinderA, "zonecyla.dff"},
-    {GameRenderer::ZoneCylinderB, "zonecylb.dff"},
-    {GameRenderer::Arrow, "arrow.dff"}};
+std::map<GameRenderer::SpecialModel, std::pair<std::string,std::string>> kSpecialModels = {
```
ShFil119: Can it be set as const?

NFSMONSTR: You mean the whole `std::map`?

ShFil119: yeah

NFSMONSTR: I'll try. Thank you

rwengine/src/engine/GameData.cpp
```diff
@@ -409,6 +409,15 @@ ClumpPtr GameData::loadClump(const std::string& name) {
     return m;
 }
 
+ClumpPtr GameData::loadClump(const std::string& name, const std::string& textureSlot) {
+    std::string currentSlot = currenttextureslot;
+    if (!textureSlot.empty())
```
ShFil119: Btw, does this `if` change anything? I mean you save current state of `currenttextureslot` in `currentSlot`. After that you override it or not. But at the end you use value from `currentSlot`. (or I don't notice something?)

NFSMONSTR: `currenttextureslot` is global variable. It used inside of loadClump to determine which texture slot to use to search for texture for this model




```c++
(GameData.cpp)
...
GameData::GameData(Logger* log, const rwfs::path& path)
    : datpath(path), logger(log), engine(nullptr) {
    dffLoader.setTextureLookupCallback(
        [&](const std::string& texture, const std::string&) {
            return findSlotTexture(currenttextureslot, texture);
        });
}
```

ShFil119: Thx for explanation.

PR #475, opened 2018-05-19
Added window configs

mole99:
As the configuration of OpenRW is rather basic, this commit extends the configuration file with a **[window]** section. It is now possible to set the **width** and **height** of the window and choose whether it should start in **fullscreen**.

This is the content of my openrw.ini file:

```
[game]
path=/path/to/game/data/
language=american
[input]
invert_y=0
[window]
width=1920
height=1080
fullscreen=0
```


dracc:
Lgtm.

darkf:
Thanks

PR #479, opened 2018-05-20
[WIP] Add icon using SDL2_image

ShFil119:
Works, but probably @madebr will kill me for this ugly FindSDL2_image.cmake (without targets), so it's WIP.

SDL2_image should be useful for adding splash screen.

- [ ] update deps/CI


LinuxDonald:
I dont know if it helps but this is our find sdl2 image from fifengine: https://github.com/fifengine/fifengine/blob/master/cmake/FindSDL2_image.cmake

darkf:
Are you really pulling in SDL2_image just for an icon? Surely there's a better way if we're not going to use the lib otherwise (ICO? XPM? Something?)

ShFil119:
I was thinking about using it for load screen. Or maybe better use ffmpeg for decoding video (intro)?

madebr:
I would never kill you. :kissing_heart:

Also:
This pr adds a run time dependency on a `logo.png` file in the current working directory.
This pr only sets the window icon. The icon of the executable is not set.

vflyson:
@ShFil119 I've had a PR open but my implementation ended up being way too old-school. Here it is: https://github.com/rwengine/openrw/pull/275 (note how @darkf didn't like there how it was an XPM, and here he's suggesting it to be an XPM). Not saying he's a bad person, perhaps he just finally got inspired by @tsjost's old-ass computers and is going back in time, realizing that old tech was more reliable.

darkf:
@vflyson That's kinda funny, but you're misinterpreting my argument on both sides. XPM isn't great because you can't easily edit it without recompiling, and there is no reason to bring in a huge dep just for an icon. stb_image is small, or BMPs/ICOs are easy to parse, although worst case XPM/ICO seems fine to me.

JayFoxRox:
@ShFil119 

> I was thinking about using it for load screen. Or maybe better use ffmpeg for decoding video (intro)?

I'm not sure what you mean. I feel this doesn't belong here either.
If you have reasons to include a dependency you should have mentioned them in the PR description already.
The GTA intros are movie files which won't be decoded with SDL2_image.
If OpenRW was ever to get a custom intro (which is not planned), it could load raw pixels or whatever.
We also always have the option to package our files in the same formats as used by the game (such as txd).

---

I also agree that SDL2_image is overkill.

[SDL_LoadBMP_RW](https://wiki.libsdl.org/SDL_LoadBMP_RW) and [SDL_LoadBMP](https://wiki.libsdl.org/SDL_LoadBMP) seem to be in SDL core.

I also feel this PR has been open for too long already which suggest design issues which could have been avoided if design decisions were discussed before work starts. Adding a new dependency is always tricky and should be avoided in most cases.

This, combined with #275 just has that taste of licking the gold plated walls of a bikeshed again.
So this is the last attempt I give to fix it: get rid of SDL2_image and find a minimalistic loader (single header or existing in codebase).
If that doesn't work out, we should close this PR and wait until someone comes up with a proper solution.

darkf:
Okay let me just suggest something obvious then:

Since we're already using SDL why don't we just use SDL_LoadBMP, which is in core SDL. There, no more deps, and no need to write anything.

dracc:
I like the idea of using an array of pixel data put inside a header or something.
Sure, it's a bit painful to edit, but it adds no file dependencies at runtime.

Lihis:
XPM, a header and can be edited easily?

madebr:
On Windows, adding a `rc` file with a ico file is sufficient.
Linux needs some icon embedded (or external) and a desktop file + icon image.

danhedron:
As has been said previously, importing SDL_Image just to read a single image is overkill.

There's examples of how to bake an image into a source file, like [this one](https://caedesnotes.wordpress.com/2015/04/13/how-to-integrate-your-sdl2-window-icon-or-any-image-into-your-executable/). I'd rather do this with a simple 32x32 image. Anything platform specific for high-resolution artwork can be done later.

ShFil119:
I'll open new pull in a moment.

PR #482, opened 2018-05-21
Added support for variable game speed

husho:
Ripped off from pickups PR and I did improve it a little

### DIFF
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -32,9 +32,8 @@ std::map<GameRenderer::SpecialModel, std::string> kSpecialModels = {
     {GameRenderer::Arrow, "arrow.dff"}};
 
 namespace {
-  constexpr float kPhysicsTimeStep = 1.0f/30.0f;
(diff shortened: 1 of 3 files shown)


rwgame/RWGame.cpp
```diff
@@ -412,37 +414,48 @@ int RWGame::run() {
...
-        lastFrame = now;
+        auto currentFrame = chrono::steady_clock::now();
+        auto frameTime =
+            chrono::duration<float>(currentFrame - lastFrame).count();
+        lastFrame = currentFrame;
 
-        if(!world->isPaused()) {
-            accumulatedTime += deltaTime;
-            world->dynamicsWorld->stepSimulation(deltaTime * timescale, kMaxPhysicsSubSteps, kPhysicsTimeStep);
+        // Clamp frameTime, so we won't freeze completely
```
darkf: Shouldn't this be a higher constant? 100ms is not a huge amount, that's less than two frames. It should be more around the 5 frame ballpark IMHO.

husho: Agreed

PR #485, opened 2018-05-23
Fixed a bunch of warnings

husho:


darkf:
Shouldn't some of those (like "max flips") still be unsigned? Rather, have you checked if the actual game format is signed or unsigned?

husho:
I think all of the stats are signed integers, because script argument is signed. As for max flips - definitely. Here's disassembly:
```
.text:00447E61 220 8B 1D D8 25 8F 00                       mov     ebx, ds:_ZN6CStats16MaximumJumpFlipsE ; CStats::MaximumJumpFlips
.text:00447E67 220 39 C3                                   cmp     ebx, eax
.text:00447E69 220 7E 05                                   jle     short loc_447E70
.text:00447E6B 220 89 D8                                   mov     eax, ebx
.text:00447E6D 220 EB 06                                   jmp     short loc_447E75
```
jle
```It performs a signed comparison jump after a cmp if the destination operand is less than or equal to the source operand.```

darkf:
I'm not sure that necessarily implies it's signed, since jle works on both types of operands of course, it could just mean that max jump flips is effectively signed, but I digress, it doesn't really matter.

LGTM, thanks.

PR #488, opened 2018-05-26
[Ready] Refactor: moved getPlayer to World class

husho:
ready

### DIFF
--- a/rwengine/src/engine/GameWorld.cpp
+++ b/rwengine/src/engine/GameWorld.cpp
@@ -1023,3 +1023,12 @@ void GameWorld::clearObjectsWithinArea(const glm::vec3 center,
     // @todo Remove all temp objects, extinguish all fires, remove all
     // explosions, remove all projectiles
 }
+
+PlayerController* GameWorld::getPlayer() {
+    auto object = pedestrianPool.find(state->playerObject);
+    if (object) {
+        auto controller = static_cast<CharacterObject*>(object)->controller;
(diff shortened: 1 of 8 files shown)


rwengine/src/script/ScriptMachine.cpp
```diff
@@ -12,11 +12,9 @@
 #include "script/ScriptModule.hpp"
 
 void ScriptMachine::executeThread(SCMThread& t, int msPassed) {
-    auto& players = getState()->world->players;
+    auto player = state->world->getPlayer();
 
-    if (!players.empty()) {
-        // @todo Add support for multiple players
```
ShFil119: Maybe it's better not to remove this comment?

husho: Don't think it's possible to support multiple players anyway, because of the way how certain scm commands work. If we want to implement some sort of coop, missions have to be rewritten in some other scripting language.

rwengine/src/engine/GameWorld.cpp
```diff
@@ -1022,4 +1022,13 @@ void GameWorld::clearObjectsWithinArea(const glm::vec3 center,
 
     // @todo Remove all temp objects, extinguish all fires, remove all
     // explosions, remove all projectiles
+}
+
+PlayerController* GameWorld::getPlayer() {
```
darkf: When can this method fail? In other words, why wouldn't there always be a player when it's called, could it not return a reference instead and throw if one isn't found?

PR #490, opened 2018-05-31
[WIP] Garages continuation

husho:
- [ ] Opcodes (most of them)
- [x] Fix swing garages
- [x] Door height based on model size
- [x] Fix isObjectInsideGarage
- [x] Second door support (partially)
- [ ] Fix segfault
- [ ] Fix double doors


rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,887 @@
...
+    }
+
+    if (needsToUpdate) {
+        updateDoor();
+    }
+}
+
+void Garage::updateDoor() {
+    if (swingType) {
+        doorObject->setRotation(glm::angleAxis(fraction * 1.57079632679f,
```
ShFil119: Hardcoded value?

husho: yes, 90 degrees

paulkocialkowski: Maybe it would be clearer to put it as either "pi / 2", "2 * pi * 90 / 360" or "pi * 90 / 180"?

ShFil119: You can use glm::half_pi. ;)

ShFil119:
Code looks good, but in my opinion you should add some (dump) unit tests. Without them it's hard to check if code is working or not.

husho:
I'll do tests in separate PR as well as extend functionality that can be actually tested.

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,887 @@
...
+            // Find out real value
+            garageTimer = engine->getGameTime() + 1.5f;
+
+            break;
+        }
+
+        case Type::Respray: {
+            // Find out real value
+            garageTimer = engine->getGameTime() + 2.f;
+            playerVehicle->setHealth(1000.f);
```
ShFil119: Shouldn't we somehow assert/check there's playerVehicle?

husho: debug assert will do

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,887 @@
...
+    }
+
+    if (needsToUpdate) {
+        updateDoor();
+    }
+}
+
+void Garage::updateDoor() {
+    if (swingType) {
+        doorObject->setRotation(glm::angleAxis(fraction * glm::radians(90.f),
```
ShFil119: Maybe here also?

husho: It would be better to replace runtime check with debug assert in constructor when it tries to find door object

husho: Don't know what error handling pattern we should use, maybe it even better to use ```std::optional```

.gitignore
```diff
@@ -5,3 +5,4 @@ documentation/
 *swp
 *.user*
 .DS_Store
+.vscode/
```
JayFoxRox: Such things don't belong in a project gitignore, but your private git configuration.
As vscode is a popular editor, I'd be willing to allow this here. But ideally it wasn't here.

rwengine/src/engine/Garage.hpp
```diff
@@ -0,0 +1,111 @@
...
+
+    void doOnOpenEvent();
+    void doOnCloseEvent();
+    void doOnStartOpeningEvent();
+    void doOnStartClosingEvent();
+
+    void updateDoor();
+
+public:
+    enum class Type {
```
JayFoxRox: nit: if these are constants the real game uses there could be a comment about this here.

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+bool Garage::isObjectInsideGarage(GameObject* object) {
+    auto p = object->getPosition();
+
+    // Do basic check first
+    if (p.x < min.x) return false;
+    if (p.y < min.y) return false;
+    if (p.z < min.z) return false;
+    if (p.x > max.x) return false;
+    if (p.y > max.y) return false;
+    if (p.z > max.z) return false;
```
JayFoxRox: nit: this sort of check is a recurring pattern and I wonder if we should have a function for it (if there isn't one already somewhere)

rwengine/src/engine/Garage.cpp
JayFoxRox: There doesn't seem to be one? Only `RW_UNUSED`. I think we should get rid of `playerIsInVehicle` in this function, and explicitly check if `playerVehicle` is valid here.

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+}
+
+void Garage::doOnStartOpeningEvent() {
+    auto player = engine->getPlayer();
+    auto plyChar = player->getCharacter();
+    auto playerPosition = plyChar->getPosition();
+    auto playerVehicle = plyChar->getCurrentVehicle();
+    bool playerIsInVehicle = playerVehicle != nullptr;
+    RW_UNUSED(playerPosition);
+    RW_UNUSED(playerIsInVehicle);
```
JayFoxRox: See other comment about `playerIsInVehicle`. Not sure how useful creating all those variables is when there is no user (and it's unlikely there'll be one).
It's probably more important to make it obvious that we check for the valitity of playerVehicle when it's being used.

rwengine/src/engine/GarageController.cpp
```diff
@@ -1,702 +0,0 @@
-#include "GarageController.hpp"
```
JayFoxRox: For future changes: When renaming files manually git does not automatically keep track of previous history. If you had used `git mv GarageController.cpp Garage.cpp`, then git should have created a much much nicer history / diff.
That would also potentially have helped in review.

Fixing this retroactively is a bit harder, but theoretically possible (essentially manually changing it back, and then using `git mv`, then force push - do at your own risk)

JayFoxRox: Nevermind, GitHub might just handle this poorly - in the "Big Refactor" commit this is displayed properly.

rwengine/src/engine/Garage.cpp
husho: Probably it would be used in the future for other garage types.

rwengine/src/engine/Garage.cpp
husho: Probably it would be used in the future for other garage types.

JayFoxRox:
> I'll do tests in separate PR as well as extend functionality that can be actually tested.

Ideally, this should have been in the PR description, so everyone knows your intentions and can focus review on what's important.

---

The git history is not ideal (looking at the commit names alone, it's very hard to tell what's going on).
Unfortunately the move of the source file in particular makes review a lot harder than it should be.

However, skimming over the code it looks mostly fine. I have not tried to run this yet.
If it doesn't break the game, I'd like to merge this soon, as it's better-than-master, and @husho promised unit tests, so we don't want to delay those.

So if someone could test this PR, that would be nice. In order to help with testing, it would be nice if @husho could add more information how to test this to the first post / PR description too (and I'd expect that for future PRs to speed up review): which missions, which garages, ...

---

I suggest creating an issue about missing garage features or unit tests after merge, so that this is not forgotten in case @husho disappears for whatever reasons.

I'd also appreciate if someone could create replacement issues for #107 which reflects the changes in this code. I'd like to close #107 as it's a meta issue which is hard to maintain.

*(Ideally these requests for updated issues should also have been mentioned in the first post / PR description, so it's easier for maintainers to keep track of what to do / look for after a merge)*

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -10957,9 +10928,8 @@ void opcode_03ba(const ScriptArguments& args, ScriptVec3 coord0, ScriptVec3 coor
     @arg garage 
 */
 void opcode_03bb(const ScriptArguments& args, const ScriptGarage garage) {
-    RW_UNIMPLEMENTED_OPCODE(0x03bb);
-    RW_UNUSED(garage);
     RW_UNUSED(args);
+   	garage->makeDoorSwing();
```
ShFil119: Unneeded tab?

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+    }
+}
+
+void Garage::close() {
+    if (state == State::Opened || state == State::Opening) {
+        state = State::Closing;
+    }
+}
+
+float Garage::getDistanceToGarage(glm::vec3 point) {
```
ShFil119: Arg can be const.

rwengine/src/engine/GameWorld.cpp
```diff
@@ -386,77 +386,11 @@ PickupObject* GameWorld::createPickup(const glm::vec3& pos, int id, int type) {
...
-    }
-
-    // Create controller
-    garageControllers.push_back(
-        std::make_unique<GarageController>(GarageController(this, info, door)));
-
-    return info;
+Garage* GameWorld::createGarage(const glm::vec3 coord0, const glm::vec3 coord1,
+                                Garage::Type type) {
+    int id = garages.size();
```
ShFil119: const

rwengine/src/engine/GameWorld.hpp
```diff
@@ -17,8 +17,8 @@
 #include <ai/AIGraphNode.hpp>
 #include <audio/SoundManager.hpp>
 
-#include <engine/GarageController.hpp>
 #include <engine/Payphone.hpp>
+#include <engine/Garage.hpp>
```
ShFil119: order

rwengine/src/engine/GameWorld.hpp
```diff
@@ -35,8 +35,8 @@ class btSequentialImpulseConstraintSolver;
 struct btDbvtBroadphase;
 
 class GameState;
-class GarageController;
 class Payphone;
+class Garage;
```
ShFil119: Also order.

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+    : engine(engine_), id(id_), type(type_) {
+    min.x = std::min(coord0.x, coord1.x);
+    min.y = std::min(coord0.y, coord1.y);
+    min.z = std::min(coord0.z, coord1.z);
+
+    max.x = std::max(coord0.x, coord1.x);
+    max.y = std::max(coord0.y, coord1.y);
+    max.z = std::max(coord0.z, coord1.z);
+
+    glm::vec2 midpoint;
```
ShFil119: It can also be const `glm::vec2 midpoint{(min.x + max.x) / 2, (min.y + max.y) / 2};`

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+    midpoint.x = (min.x + max.x) / 2;
+    midpoint.y = (min.y + max.y) / 2;
+
+    // Find door objects for this garage
+    for (auto p : engine->instancePool.objects) {
```
ShFil119: Also?

rwengine/src/engine/Garage.cpp
```diff
@@ -0,0 +1,889 @@
...
+        if (!inst->getModel()) {
+            continue;
+        }
+
+        if (!SimpleModelInfo::isDoorModel(
+                inst->getModelInfo<BaseModelInfo>()->name)) {
+            continue;
+        }
+
+        auto instPos = inst->getPosition();
```
ShFil119: And here and below?

rwengine/src/engine/Garage.hpp
```diff
@@ -0,0 +1,112 @@
...
+    // @todo use model type
+    int targetModel = 0;
+
+    State state = State::Closed;
+
+    bool resprayDone = false;
+
+    Garage(GameWorld* engine_, const int id_, const glm::vec3 coord0,
+               const glm::vec3 coord1, const Type type_);
+    ~Garage();
```
ShFil119: `= default;`?

rwengine/src/engine/Garage.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    void makeDoorSwing();
+
+    bool isObjectInsideGarage(GameObject* object);
```
ShFil119: Can this and below be const?

PR #492, opened 2018-06-01
Fix vehicle damage

husho:


ShFil119:
Can you add in commit message short note what didn't work and how you fixed it?

JayFoxRox:
nit: Commit title is slightly wrong, I think the real message should convey that this was about *visual* damage / *strong impacts*.

For future PRs: Always add a PR description to explain the changes - regardless how small the change is.
I think there might be a value in having a threshold for damage (and might already be done elsewhere in the code?). I worry this change might result in too much damage in rare cases (sliding along walls possibly). However, I do not know how damage is calculated in openrw - I think the PR description would have been a good room to explain this, and why we don't need the threshold (which might have been an intentional decision).

Anyhow, I also think this was a bug, and I'd assume this change doesn't break anything.
If it does, we can always fix it again later.

LGTM; Merging.

PR #495, opened 2018-06-01
Implement some cheats and showing cheat activated message

NFSMONSTR:


rwgame/RWGame.cpp
```diff
@@ -212,31 +216,40 @@ void RWGame::handleCheatInput(char symbol) {
...
+                                           20, //shotgun
+                                           5,  //grenade
+                                           5,  //molotov
+                                           5,  //rocket launcher
+                                           20, //flamethrower
+                                           200,//ak47
+                                           200,//m16
+                                           5   //sniper rifle
+                                           };
+            for (unsigned long i = 0; i < ammo.size(); i++)
```
husho: better to use ```std::size_t```

husho: or even ```size_type```

NFSMONSTR: Thank you

NFSMONSTR: It seems like that using `size_type` will cause warning about comparison of integer expressions of different signedness. So I'll think it better to use something like `std::vector<size_type>::size_type`

husho: Yeah, that's what I meant, it should be  ```std::vector<int>::size_type```

rwgame/RWGame.cpp
```diff
@@ -212,31 +216,40 @@ void RWGame::handleCheatInput(char symbol) {
...
-            // @todo ShowHelpMessage("CHEAT4"); // III / VC: Inputting armor
-            // cheat.
+            showHelpMessage("CHEAT4"); // III / VC: Inputting armor cheat.
         });
 
         checkForCheat("GUNSGUNSGUNS", [&] {
-            // @todo give player weapons
-            // @todo ShowHelpMessage("CHEAT2"); // III / VC: Inputting weapon
-            // cheats.
+            const std::vector<int> ammo = {1,  //baseball bat
```
ShFil119: It can even be a std::array.

rwgame/RWGame.cpp
```diff
@@ -245,51 +258,50 @@ void RWGame::handleCheatInput(char symbol) {
...
+         uint16_t vehicleModel = 110;  // @todo Which cars are spawned?!
 #else
-        // uint16_t vehicleModel = 122;
+         uint16_t vehicleModel = 122;
 #endif
-            // @todo Spawn rhino
-            // @todo ShowHelpMessage("CHEAT1"); // III / VC: Inputting most
-            // cheats.
+            auto ch = getPlayer()->getCharacter();// @todo Change spawn place to be more like in original game
+            auto playerRot = ch->getRotation();
```
ShFil119: Can this and below vars be const?

rwgame/RWGame.cpp
```diff
@@ -160,6 +160,11 @@ void RWGame::startScript(const std::string& name) {
     }
 }
 
+void RWGame::showHelpMessage(const GameStringKey& id) {
```
husho: This should be a member of GameState

husho:
Needs rebase

NFSMONSTR:
I will just recreate it normally and updated

JayFoxRox:
This was still fixable - in the future, avoid closing PRs unless there is a fundamental design issue or it's not necessary anymore.
By closing the PR you broke GitHubs internal mechanism, so you can't re-open this one without actually causing issues.

NFSMONSTR:
Sorry, I still weak with GitHub and Git

PR #476, opened 2018-05-19
Redid opcodes related to player input and cutscenes

husho:
Removed wrong functionality from 0373, added 03ef.

PR #498, opened 2018-06-02
[WIP] Payphones

husho:


rwengine/src/engine/GameWorld.cpp
```diff
@@ -459,6 +459,15 @@ GarageInfo* GameWorld::createGarage(const glm::vec3 coord0,
     return info;
 }
 
+Payphone* GameWorld::createPayphone(const glm::vec2 coord) {
+    int id = payphones.size();
+    Payphone* payphone = new Payphone(this, id, coord);
```
ShFil119: Memory leak?

rwengine/src/ai/PlayerController.hpp
```diff
@@ -58,6 +66,23 @@ class PlayerController : public CharacterController {
     // @todo not implemented yet
     bool isBusted() const;
 
+    void pickUpPayphone();
```
ShFil119: Here and below some kind of comments?

rwengine/src/engine/Payphone.hpp
```diff
@@ -0,0 +1,45 @@
...
+    int id;
+
+    int getScriptObjectID() const {
+        return id;
+    }
+
+    Payphone(GameWorld* engine_, const int id_, const glm::vec2 coord);
+    ~Payphone();
+
+    void enable();
```
ShFil119: Here and below some kind of comments?

rwengine/src/engine/GameWorld.cpp
ShFil119: Maybe it's better to keep in vector unique_ptrs?

rwengine/src/engine/SaveGame.cpp
```diff
@@ -1257,6 +1257,12 @@ bool SaveGame::loadGame(GameState& state, const std::string& file) {
         }
     }
 
+    // @todo restore properly
+    for (size_t p = 0; p < payphoneData.numPayphones; ++p) {
```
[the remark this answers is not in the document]
husho: I don't really wanna touch SaveGame.cpp, as it was being refactored by @danhedron back in February. But there is no progress since :(

rwengine/src/engine/GameWorld.cpp
husho: But it's not uniquely-owned

ShFil119: You can create unique_ptr with this memory when inserting into vector. Other way is usage of `std::move`.

husho: But we still need to return back valid object after inserting into vector, and it's no longer valid after ```std::move``` Or I guess you can just do ```back``` on vector to return newly created object.

husho: That's all nice, but it doesn't really go well with all this ScriptObjectType stuff

ShFil119: Raw ptr to memory allocated by unique_ptr you can get using `get`.

husho: Doesn't that break the whole purpose of having a unique_ptr then?

ShFil119: Nope, there's only one owner of memory(so overhead as raw ptr) and you have quarantine that memory is freed at the end. 

Good habit is to use ref instead of raw ptr when borrowing memory. As I suggested on IRC lately.

ShFil119:
Also here unit tests would be useful.

rwengine/src/engine/Payphone.hpp
```diff
@@ -0,0 +1,49 @@
...
+    State state = State::Idle;
+
+    int id;
+
+    int getScriptObjectID() const {
+        return id;
+    }
+
+    Payphone(GameWorld* engine_, const int id_, const glm::vec2 coord);
+    ~Payphone();
```
ShFil119: maybe better `~Payphone() = default;`?

rwengine/src/engine/Payphone.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (glm::distance(coord, glm::vec2(o->getPosition())) < 2.f) {
+            object = static_cast<InstanceObject*>(o);
+        }
+    }
+
+    position = object->getPosition();
+    message.clear();
+}
+
+Payphone::~Payphone() {
```
ShFil119: And this remove.

darkf:
Pinging to remind you that the build fails and there are unaddressed comments

husho:
@darkf review #476 then. And again if SaveGame is going to be refactored to use streams why bother with it?

darkf:
@husho What's the relation between this and #476? ... And you can always change the save stuff later.

husho:
I used code from #476, it needs to get merged first, so this can be rebased.

husho:
Also it's not me who worked on save stuff. Check danhedron save-streams branch

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6610,70 +6610,64 @@ void opcode_0249(const ScriptArguments& args, const ScriptModel model) {
...
+    @arg payphone 
+    @arg text 
 */
-void opcode_024c(const ScriptArguments& args, const ScriptPhone phone, const ScriptString arg2) {
-    RW_UNIMPLEMENTED_OPCODE(0x024c);
-    RW_UNUSED(phone);
-    RW_UNUSED(arg2);
+void opcode_024c(const ScriptArguments& args, const ScriptPayphone payphone, const ScriptString text) {
     RW_UNUSED(args);
+    payphone->setMessageAndStartRinging(text);
```
ShFil119: Does it work, I mean payphone is const.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -6610,70 +6610,64 @@ void opcode_0249(const ScriptArguments& args, const ScriptModel model) {
...
     opcode 024e
-    @arg phone 
+    @arg payphone 
 */
-void opcode_024e(const ScriptArguments& args, const ScriptPhone phone) {
-    RW_UNIMPLEMENTED_OPCODE(0x024e);
-    RW_UNUSED(phone);
+void opcode_024e(const ScriptArguments& args, const ScriptPayphone payphone) {
     RW_UNUSED(args);
+    payphone->disable();
```
ShFil119: Does it work (I mean payphone is const)?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -11827,15 +11819,14 @@ void opcode_0404(const ScriptArguments& args) {
...
     opcode 0405
-    @arg phone Handle
+    @arg payphone Handle
 */
-void opcode_0405(const ScriptArguments& args, const ScriptPhone phone) {
-    RW_UNIMPLEMENTED_OPCODE(0x0405);
-    RW_UNUSED(phone);
+void opcode_0405(const ScriptArguments& args, const ScriptPayphone payphone) {
     RW_UNUSED(args);
+    payphone->enable();
```
ShFil119: also

husho: Pointer to the object is unchangeable, not the object itself, object method declaration implies that object can be mutated (no const).

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: Pointer to the object is unchangeable, not the object itself, object method declaration implies that object can be mutated (no const).

rwengine/src/script/modules/GTA3ModuleImpl.inl
husho: Pointer to the object is unchangeable, not the object itself, object method declaration implies that object can be mutated (no const).

rwengine/src/script/ScriptTypes.cpp
```diff
@@ -220,6 +220,18 @@ ScriptObjectType<GarageInfo> ScriptArguments::getScriptObject(
     return {param.handleValue(), garage};
 }
 template <>
+ScriptObjectType<Payphone> ScriptArguments::getScriptObject(
+    unsigned int arg) const {
+    auto& param = (*this)[arg];
+    RW_CHECK(param.isLvalue(), "Non lvalue passed as object");
+    auto& payphones = getWorld()->payphones;
+    Payphone* payphone = nullptr;
+    if (size_t(*param.handleValue()) < payphones.size()) {
```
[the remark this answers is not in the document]
husho: size_t is unsigned, also this code was copied from other function, so what? do I change names in all similar functions then?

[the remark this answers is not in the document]
darkf: At least here because it's used more than once and it makes it clear it's being used as an index.

And what I mean is the index (`*param.handleValue()`) could be < 0 from the script, before the cast. Should probably be an assertion.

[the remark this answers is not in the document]
husho: How about using ```std::vector::at```

rwengine/src/engine/GameWorld.cpp
```diff
@@ -461,7 +461,7 @@ GarageInfo* GameWorld::createGarage(const glm::vec3 coord0,
 
 Payphone* GameWorld::createPayphone(const glm::vec2 coord) {
     int id = payphones.size();
-    payphones.emplace_back(this, id, coord);
+    payphones.emplace_back(make_unique<Payphone>(this, id, coord));
```
ShFil119: Didn't you forget include memory?

husho: It's in the header file

husho: I dont understand why it was merged before #476 even

PR #500, opened 2018-06-03
Implement script timer

husho:


rwengine/src/engine/GameState.hpp
```diff
@@ -386,6 +386,12 @@ class GameState {
      */
     std::vector<TextDisplayData> texts;
 
+    /**
+     * Script timer
+     */
+    int32_t* scriptTimerVariable = nullptr;
```
darkf: `ScriptInt*` please

PR #503, opened 2018-06-04
Use CPack to create archives on appveyor

madebr:
- Install executables/... in GNU default directories: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
- Use CPack to generate archive: https://cmake.org/cmake/help/latest/module/CPack.html
- Use CPack to create an artifact on appveyor.

I'm using TXZ because that one has the highest compression ratio. 7z would also be a great choice.
```
$ du -b OpenRW-13946234.* | sort
1286216	OpenRW-13946234.tar.xz
1286970	OpenRW-13946234.7z
1905566	OpenRW-13946234.tar.bz2
2430744	OpenRW-13946234.tar.gz
2432541	OpenRW-13946234.zip
2469706	OpenRW-13946234.sh
3539451	OpenRW-13946234.tar.Z
```


rwengine/CMakeLists.txt
```diff
@@ -159,3 +159,10 @@ target_include_directories(rwengine
     )
 
 openrw_target_apply_options(TARGET rwengine)
+
+if(INSTALL_LIBS)
+    install(TARGETS rwengine
```
danhedron: Is there a reason we'd want to install the static library?

madebr: Support for third person total conversions :rofl: .

Shared libraries should be installed.
We can switch the test to `BUILD_SHARED_LIBS`.

ShFil119:
lgtm, should I close #466?

darkf:
Needs a rebase.

madebr:
rebased and tested artifact on appveyor using wine

darkf:
Thanks.

PR #507, opened 2018-06-06
Fix: terminate only mission scripts when starting new mission via debug menu

husho:
and clear unnecessary blips

### DIFF
--- a/rwengine/src/engine/SaveGame.cpp
+++ b/rwengine/src/engine/SaveGame.cpp
@@ -1223,7 +1223,7 @@ bool SaveGame::loadGame(GameState& state, const std::string& file) {
     for (size_t s = 0; s < numScripts; ++s) {
         state.script->startThread(scripts[s].programCounter);
         SCMThread& thread = threads.back();
-        // thread.baseAddress // ??
+        // no baseAddress in III and VC
         strncpy(thread.name, scripts[s].name, sizeof(SCMThread::name) - 1);
         thread.conditionResult = scripts[s].ifFlag;
         thread.conditionCount = scripts[s].ifNumber;
--- a/rwgame/states/DebugState.cpp
+++ b/rwgame/states/DebugState.cpp
@@ -317,18 +317,22 @@ std::shared_ptr<Menu> DebugState::createMissionsMenu() {
 
             if (vm) {
                 std::list<SCMThread>& threads = vm->getThreads();
+                const auto& offsets = vm->getFile()->getMissionOffsets();
+
+                RW_ASSERT(!offsets.empty());
 
                 for (auto& thread : threads) {
-                    if (thread.isMission) {
+                    if (thread.baseAddress >= offsets[0]) {
                         thread.wakeCounter = -1;
                         thread.finished = true;
                     }
                 }
 
-                RW_ASSERT(i < vm->getFile()->getMissionOffsets().size());
+                game->getState()->radarBlips.clear();
 
-                auto offset = vm->getFile()->getMissionOffsets()[i];
-                vm->startThread(offset, true);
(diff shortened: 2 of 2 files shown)


rwengine/src/engine/SaveGame.cpp
```diff
@@ -1226,6 +1226,7 @@ bool SaveGame::loadGame(GameState& state, const std::string& file) {
         state.script->startThread(scripts[s].programCounter);
         SCMThread& thread = threads.back();
         // thread.baseAddress // ??
+        // husho: baseAddress does not exist in III and VC
```
JayFoxRox: nit: comment should be on same line or above.
Also "// Only in SA" would be a better comment (and LCS / VCS if that affects it).

husho: I didnt write only in SA, because I dont know how its implemented in SA

rwgame/states/DebugState.cpp
```diff
@@ -317,18 +317,22 @@ std::shared_ptr<Menu> DebugState::createMissionsMenu() {
 
             if (vm) {
                 std::list<SCMThread>& threads = vm->getThreads();
+                const auto& offsets = vm->getFile()->getMissionOffsets();
+
+                RW_ASSERT(!offsets.empty());
 
                 for (auto& thread : threads) {
-                    if (thread.isMission) {
+                    if (thread.baseAddress >= offsets[0]) {
```
JayFoxRox: The comment above said that the baseAddress does not exist in III and VC. But now it's used.
I'm confused.

JayFoxRox: Ah, I assume the comment above referred to Savegames only.

husho: openrw implements scripts differently, instead of loading mission script bytecode into fixed .bss array, it just loads whole scm and uses baseAddress for mission relative jumps, meanwhile VC for example just uses fixed offset 225512 or for III - 0x20000

JayFoxRox:
This has been around for almost 2 weeks.
Final call for review: I want someone to test this before merge.

---

Some more comments about PR quality:

- Commit: "Improved missions debug menu" - title: how? Attempt to be more precise. Once merged, most people will not read commits grouped by PR, but stand-alone, so the context is lost.
- Commit: "Don't use hardcoded value" - title: which value? Attempt to be more precise; You can also just squash this into the previous commit as the intermediate step with the hardcoded value serves no real use.
- Commit "Changed comment": can be squashed; also nitpicking some more: adding "Unknown for other GTA games" is more clear, as the current comment could imply that it *does* exist in all others - which was my mistake when I made my other comment. This also feels like a fixup and can be squashed into the first commit.

All of this is nitpicking and optional. But I'd prefer if you fixed it so we can raise our history quality (which sets an example for future work and helps when debugging issues, so it will save time in the long run).

rwgame/states/DebugState.cpp
danhedron: Is the meaning of the field documented anywhere? It would be good to correct the discrepancy.

husho: what field?

danhedron: baseAddress

danhedron: Ah, then I just misunderstood. I thought you meant that it just isn't in the save game?

danhedron:
LGTM

rwgame/states/DebugState.cpp
husho: If mission thread needs to do relative jump it subtracts negative jump offset from fixed offset, there is a special buffer for mission bytecode.

Issue #317, opened 2017-09-20
Replace boost::filesystem with std::filesystem


darkf:
A post-C++14 TS and now standard in C++17 adds `std::experimental::filesystem::v1` and `std::filesystem` respectively, which is basically `boost::filesystem`.

As Boost is large and a nonstandard dependency, we'd like to move as much of its use as possible to the STL, so this would be a good change to start with.

Since we should target effectively modern compilers, I wager most systems would have this library already -- although we can `#ifdef` it to use Boost as a fallback.

danhedron:
The usual pattern seems to be defining a fs namespace that aliases either the std filesystem or boost filesystem in a header, limiting the checking to a single definition. It should be simple to replace the existing usage

PR #509, opened 2018-06-08
Refactor LoaderCutsceneDAT and fix cutscenes on macOS

### LINKED ISSUE #378: Cutscenes not being rendered on MacOS
I get this on MacOS 10.13.2. I can't do anything until a cutscene ends. The sound is working. I can't skip a cutscene either.

screen shot 2018-01-29 at 23 40 49

emilisb:
Refactored Cutscene DAT Loader. Fixes #378 . The old code wasn't working on macOS.

screen shot 2018-06-08 at 19 31 34

### DIFF
--- a/rwengine/src/loaders/LoaderCutsceneDAT.cpp
+++ b/rwengine/src/loaders/LoaderCutsceneDAT.cpp
@@ -4,8 +4,10 @@
 #include <limits>
 #include <sstream>
 #include <string>
+#include <vector>
 
 #include <glm/glm.hpp>
+#include <boost/algorithm/string.hpp>
 
 #include <rw/defines.hpp>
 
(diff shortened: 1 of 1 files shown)


rwengine/src/loaders/LoaderCutsceneDAT.cpp
```diff
@@ -16,17 +18,21 @@ void LoaderCutsceneDAT::load(CutsceneTracks &tracks, const FileHandle& file) {
...
-        ss >> t;
-        ss.ignore(2, ',');
-        ss >> z;
-        ss.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+        std::string line;
+        ss >> line;
+
+        boost::split(lineParts, line, boost::is_any_of(","));
+
+        float t = stof(lineParts[0]);
```
husho: Better to use ```std::vector::at``` to check bounds, don't trust input data

emilisb: Good point. Fixed, thanks :)

rwengine/src/loaders/LoaderCutsceneDAT.cpp
```diff
@@ -16,17 +18,21 @@ void LoaderCutsceneDAT::load(CutsceneTracks &tracks, const FileHandle& file) {
...
-        float t = 0.f;
-        float z = 0.f;
-        ss >> t;
-        ss.ignore(2, ',');
-        ss >> z;
-        ss.ignore(std::numeric_limits<std::streamsize>::max(), '\n');
+        std::string line;
+        ss >> line;
+
+        boost::split(lineParts, line, boost::is_any_of(","));
```
christoph-heiss: Wasn't there once some endeavor (#317) for getting rid of the boost dependency, at least for anything except the testcases? 

(IMHO, I do not really see the advantage here over using e.g. `std::getline` with `,` as separator, or just fixing the existing code.)

darkf:
I agree that boost should be avoided where possible

JayFoxRox:
@emilisb it has been almost 2 weeks since you've sent this PR and more than 1 week since the first feedback:

- What's your plan with this PR?
    - Are you still around?
    - Are you going to address the feedback?
- Also, is there an issue for this problem (if so, please mention it in the PR description)?
    - What is the actual problem with the current code?

JayFoxRox:
Closed.

Continued in #533 which does the same boost-less, has better documentation and active code author.
Thanks to @emilisb for pointing out a possible solution and motivating the work :)

PR #510, opened 2018-06-08
Don't draw asterisks in subtitles

NFSMONSTR:
img

In English version GTA 3 some subtitles missing and replaced with asterisks, which original game don't draw

### DIFF
--- a/rwengine/src/render/TextRenderer.cpp
+++ b/rwengine/src/render/TextRenderer.cpp
@@ -163,6 +163,9 @@ void TextRenderer::renderText(const TextRenderer::TextInfo& ti,
     for (size_t i = 0; i < text.length(); ++i) {
         char16_t c = text[i];
 
+        if (c == '*')
+           break;
+
         // Handle any markup changes.
         if (c == '~' && text.length() > i + 1) {
             switch (text[i + 1]) {


JayFoxRox:
I wonder if this change is such a good idea. It is a logic change as part of the renderer.
If this change is bad, it might be hard to track down what causes bugs in the future.
This change could lead to text not appearing when it should, and it might be difficult to tell that text is even missing. This might result in confusing game dialog / logic which could accidentally be blamed on something else.

Do we know that this game handles this the same way?
What about strings like " * " (are spaces still drawn? with the textboxes it would change the size) or "foo*bar" (will it display as "foo", "foo bar", ...)

So I'm wondering if we should `break`, `continue`, `if (str == "*") { return; }` or maybe something entirely different.

If you can't be bothered to check what the original game does (or at least some testing with the original game), I'd suggest creating an easy-to-find issue about this, so we consider it an option when problems arise.

husho:
```return``` would be better https://pastebin.com/d3EQ155N

**Comment by NFSMONSTR** [comment]
You right. The game really checks only the first char. My bad

That how looks "foo*bar" in game:
img

File: rwengine/src/render/TextRenderer.cpp [hunk 1]
```diff
@@ -141,6 +141,9 @@ void TextRenderer::setFontTexture(int index, const std::string& texture) {
 
 void TextRenderer::renderText(const TextRenderer::TextInfo& ti,
                               bool forceColour) {
+    if ((ti.text.empty()) || (ti.text[0] == '*'))
```
darkf: What's with the unnecessary parens?

PR #516, opened 2018-06-18
Fix panic on new game 

husho:
Payphone pool is empty while ScriptPayphone object is getting unpacked during payphone creation caused out of range panic

master was broken by #498

### DIFF
--- a/rwengine/src/script/ScriptTypes.cpp
+++ b/rwengine/src/script/ScriptTypes.cpp
@@ -224,10 +224,13 @@ ScriptObjectType<Payphone> ScriptArguments::getScriptObject(
     unsigned int arg) const {
     auto& param = (*this)[arg];
     RW_CHECK(param.isLvalue(), "Non lvalue passed as object");
-    auto index = *param.handleValue();
-    RW_CHECK(index >= 0, "Object index is negative");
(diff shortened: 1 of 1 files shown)


JayFoxRox:
I'm not sure what would happen if the index is negative, but I guess it does not matter as the `RW_CHECK` would be triggered.
I also wonder why the code isn't closer to that of `VehicleGenerator` or `GarageInfo` which seems to do the same thing (but still lacks a check for negative indices).
I guess even if it's bad, those changes can be delayed as this is a quickfix.

Aside from that LGTM.
Did only skim over code, but did not test.

---

@husho In the future I'd recommend to:

1. Create an issue if something broke master and state if and when you are intending to fix it. The title should make clear that master is broken ("Crash when starting a new game")
2.  Respond on the PR which broke master that it has accidentally broke it (referencing the new issue).
3. When creating a Fix-PR, referece the original PR and the issue in the Fix-PR, and possibly have something like "Quick-Fix" in the title, to suggest that this is a small change which should be applied sooner than others.

Nothing in this PR (or issues) really suggests that master is currently broken or that this fixes it.

JayFoxRox:
Not a big fan of the commit title, but this will do. Thanks @husho for the quick fix.

Merged.

PR #522, opened 2018-06-21
Prepare sdt loader for ffmpeg

ShFil119:
Extracted from #420.


rwlib/source/loaders/LoaderSDT.hpp
```diff
@@ -33,16 +57,23 @@ class LoaderSDT {
...
+    /// Destructor
+    ~LoaderSDT() = default;
 
     /// Load the structure of the archive
     /// Omit the extension in filename
-    bool load(const std::string& filename);
+    bool load(const rwfs::path& path);
 
     /// Load a file from the archive to memory and pass a pointer to it
     /// Warning: Please delete[] the memory in the end.
```
ShFil119: to remove

ShFil119:
@madebr has suggested on IRC to prevent linking FFmpeg to rwlib. I'll try something like this: sdt loader provides raw data and SoundManager deals with it.

ShFil119:
Updated, please review. ;) 
(adding extra comment, because not sure if forced push will trigger message)

rwengine/src/audio/SoundManager.cpp
```diff
@@ -173,6 +178,234 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+        avcodec_close(codecContext);
+        avcodec_free_context(&codecContext);
+        avformat_close_input(&formatContext);
+        RW_ERROR("Couldn't open the audio codec context");
+        return;
+    }
+#endif
+
+    // Expose audio metadata
+    channels = codecContext->channels;
```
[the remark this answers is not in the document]
ShFil119: Here you think about casting or changing some variable's type?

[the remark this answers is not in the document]
Lihis: I was thinking of casting. Tho, as `uint16_t` is being used in `WaveHeader` is `size_t` really needed.

ShFil119:
Done. ;)

rwengine/src/audio/SoundManager.cpp
```diff
@@ -173,6 +178,226 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+    auto inputDataStart = std::make_unique<uint8_t[]>(input.size); /// Store start ptr of data to be able freed memory later
+    input.ptr = inputDataStart.get();
+
+    /// Alocate memory for buffer
+    /// Memory freeded at the end
+    static constexpr size_t ioBufferSize = 4096;
+    auto  ioBuffer = static_cast<uint8_t*>(av_malloc(ioBufferSize));
+
+    /// Convert memory, in order to match required layout for ffmpeg
+    memcpy(input.ptr, raw_sound.get(), input.size);
```
danhedron: Why does this need to be copied? Can't the memory for raw_sound be given to the input?

rwengine/src/audio/SoundManager.cpp
```diff
@@ -173,6 +178,226 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+    /// Prepare input
+    InputData input{};
+    input.size = sizeof(WaveHeader) + sdt.assetInfo.size;
+    auto inputDataStart = std::make_unique<uint8_t[]>(input.size); /// Store start ptr of data to be able freed memory later
+    input.ptr = inputDataStart.get();
+
+    /// Alocate memory for buffer
+    /// Memory freeded at the end
+    static constexpr size_t ioBufferSize = 4096;
+    auto  ioBuffer = static_cast<uint8_t*>(av_malloc(ioBufferSize));
```
danhedron: You can use ``unique_ptr`` with a custom deleter to handle specific dealloc functions from C APIs.

rwengine/src/audio/SoundManager.cpp
ShFil119: I don't like pointer casting, but memory layout should match. I'll check.

rwengine/src/audio/SoundManager.cpp
ShFil119: I think it'll be confusing for reading.

Also, actually buffer is deleted before context, this change will flip order. FFmpeg really likes to change api, so I think it's better to be as close as possible to "template" use.

rwengine/src/audio/SoundManager.cpp
```diff
@@ -173,6 +178,226 @@ void SoundManager::SoundSource::loadFromFile(const rwfs::path& filePath) {
...
+
+    /// Finally prepare our "custom" format context
+    AVIOContext* avioContext = avio_alloc_context(ioBuffer, ioBufferSize, 0, &input, &read_packet, nullptr, nullptr);
+    AVFormatContext* formatContext = avformat_alloc_context();
+    formatContext->pb = avioContext;
+
+    if (avformat_open_input(&formatContext, "nothint", nullptr, nullptr) != 0) {
+        av_frame_free(&frame);
+        RW_ERROR("Error opening audio file (" << index << ")");
+        return;
```
danhedron: Should ``formatContext->pb`` be freed in these early branches?

ShFil119: should be

ShFil119: Done.

danhedron:
LGTM

PR #523, opened 2018-06-21
[Ready]Usage of default member initializer

ShFil119:
Pros:
- Cleanup: actually some classes as initialized in headers, some in constructor in cpp. 
- Should make headers more readable.
- Reduces time of deps to rebuild.
- Don't need to remember to set values when we are adding more ctors.
- compiler checks if is value valid (see https://github.com/rwengine/openrw/commit/625ebdcb90f4f7e720793458ba056a1704d63218)


husho:
Needs rebase

ShFil119:
Rebased.

rwengine/src/engine/GameData.hpp
```diff
@@ -60,7 +60,7 @@ class GameData {
     GameData(Logger* log, const rwfs::path& path);
     ~GameData();
```
Lihis: This should be `default`.

ShFil119: Done. ;)

rwengine/src/render/WaterRenderer.hpp
```diff
@@ -41,19 +41,19 @@ class WaterRenderer {
     void render(GameRenderer* renderer, GameWorld* world);
 
 private:
-    std::unique_ptr<Renderer::ShaderProgram> waterProg;
-    std::unique_ptr<Renderer::ShaderProgram> maskProg;
+    std::unique_ptr<Renderer::ShaderProgram> waterProg = nullptr;
+    std::unique_ptr<Renderer::ShaderProgram> maskProg = nullptr;
```
danhedron: nit: This is already the default value for ``unique_ptr``

PR #525, opened 2018-06-21
Fix #349 "Smooth vehicle steering"

### LINKED ISSUE #518: "Smooth steering" PR needs cleanup
The PR #349 was left with a few comments but its author went AWOL.

Unless someone else gets to it earlier, I'll start digging into it in the next 24h.

dracc:
There were some minor points of improvement pointed out in the PR #349 . Here I've tried addressing those.

Closes #518 .

### DIFF
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -310,6 +310,8 @@ void VehicleObject::tick(float dt) {
 void VehicleObject::tickPhysics(float dt) {
     RW_UNUSED(dt);
 
+    static constexpr float steeringWeight = 1.f/0.35f;
+
     if (physVehicle) {
         // todo: a real engine function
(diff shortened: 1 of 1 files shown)


rwengine/src/objects/VehicleObject.cpp
```diff
@@ -310,6 +310,8 @@ void VehicleObject::tick(float dt) {
 void VehicleObject::tickPhysics(float dt) {
     RW_UNUSED(dt);
 
+    const float steeringWeight = 1.f/0.35f;
```
ShFil119: It can be `static constexpr`.

dracc: There, I fixeded it.

dracc:
There was a comment in #349 about the special case "no steering". I have been busy the last few days, have not tried

PR #526, opened 2018-06-21
Rwlib changes

madebr:
- `rw/defines.hpp`  --> `rw/debug.hpp`
- `bit_cast` is a header, not a source file
- get the path to the background file from FileIndex
- refactor FileIndex a bit:
    - `rwfs::path` -> a path on the file system
       `std::string` -> a relative path, indexed by `FileIndex`
    - make `FileIndex` const correct (std::map::operator[] modifies the std::map)
    - use a single map internally
    - keys of this map are paths relative in the game data directory + filenames
      (so paths like "data/main.scm" can be found twice in the map: "data/main.scm" and "main.scm")
    - normalization of the file paths is done inside `FileIndex`,
      so there shouldn't be any ::tolower's used anywhere else anymore.
    - added a normalizer helper function `normalizeFilePath`
    - updated documentation
    - updated tests

### DIFF
--- a/rwengine/src/BinaryStream.cpp
+++ b/rwengine/src/BinaryStream.cpp
@@ -2,7 +2,7 @@
 
 #include <cstring>
 #include <fstream>
-#include <rw/defines.hpp>
+#include <rw/debug.hpp>
 
 namespace RW {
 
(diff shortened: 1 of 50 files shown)


husho:
Needs rebase

madebr:
Rebased.

PR #533, opened 2018-06-26
Fix cutscene only showing a grey screen on macOS.

### LINKED ISSUE #378: Cutscenes not being rendered on MacOS

christoph-heiss:
[Fixes #378]

Since the author of #509 apparently went MIA, I though I would quickly fix this up since I own a macOS machine.

After some debugging across two machines, I found the root cause.
Basically the issue was caused by a difference how the different C++ standard library implementations process formatted input.

Here is a minimal reproducer of the problem:
```
#include <iostream>
#include <sstream>
int main()
{
        float f;
        std::stringstream{"1234f"} >> f;
        std::cout << f << std::endl;
        return 0;
}
```
On a macOS machine, this returns 0. On a linux box, 1234 (as the previous code assumed).

I also included a small build failure fix when using a combination of gcc/clang and the different filesystem libraries, which I had laying around for a while now.


madebr:
LGTM, is the `CMAKE_CXX_COMPILER_ID` on mac equal to `Clang`?
Can the test for `AppleClang` be removed from line 8 from `cmake_configure.cmake`?

christoph-heiss:
Just tested it out, `CMAKE_CXX_COMPILER_ID` is also just `Clang` on macOS.
Do you want me to push another commit to remove that?

madebr:
That would be nice. Since the topic is Apple. :green_apple:

christoph-heiss:
No, it is `AppleClang` (at least for the system compiler provided by XCode).

For clang installed by Homebrew, it is just `Clang` (which is an obscure edge-case).
I came across this when I test-compiled using g++ on macOS. 

(Apple really has some platform-specific weirdness.)

madebr:
Thanks for verifying!
Did not know this.

JayFoxRox:
@madebr review sounds like it only reviewed the code, much like mine.
Aside from @christoph-heiss , did any other macOS user (and at least one gcc Linux user) *run* this for testing?

I'll probably hit merge once CI passed, if we have confirmed these touched platforms still *run* fine.

madebr:
Just tested it on linux, gcc 8. rwgame still works, rwtests shows no difference in behavior.

tsjost:
Tested on MacOS 10.13.4 with "_Apple LLVM version 9.1.0 (clang-902.0.39.1)_"; current master 7061813a indeed has broken initial cutscene, while this patch fixes it 👍

PR #536, opened 2018-07-03
Usage of resampler for sounds

### LINKED ISSUE #458: No sound in Cutscenes, affected only FFmpeg 4.0?
I'll check if ffmpeg changed api behavior. Please answer if you also encounter this problem. ;)

ShFil119:
Commit fixes #458.

It also loosens the connection
between decoder and player (openAL).

It allows to easier replace
player if there's need.
To do later:
IIRC @madebr has said something about using rwfs?

### DIFF
--- a/rwengine/src/audio/SoundManager.cpp
+++ b/rwengine/src/audio/SoundManager.cpp
@@ -7,6 +7,8 @@ extern "C" {
 #include <libavformat/avformat.h>
 #include <libavformat/avio.h>
 #include <libavutil/avutil.h>
+#include <libavutil/opt.h>
+#include <libswresample/swresample.h>
 }
 //ab
 #include <rw/debug.hpp>
@@ -20,6 +22,9 @@ extern "C" {
 #if LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(57,80,100)
 #define avio_context_free av_freep
 #endif
+
+constexpr int kNumOutputChannels = 2;
+constexpr AVSampleFormat kOutputFMT = AV_SAMPLE_FMT_S16;
 
 SoundManager::SoundManager() {
     initializeOpenAL();
@@ -58,8 +63,6 @@ bool SoundManager::initializeOpenAL() {
 }
 
 bool SoundManager::initializeAVCodec() {
-    av_register_all();
-
 #if RW_DEBUG && RW_VERBOSE_DEBUG_MESSAGES
         av_log_set_level(AV_LOG_WARNING);
 #else
@@ -93,8 +96,7 @@ void SoundManager::SoundSource::loadFromFile(const rwfs::path& filePath) {
     }
 
     // Find the audio stream
-    AVCodec* codec = nullptr;
-    int streamIndex = av_find_best_stream(formatContext, AVMEDIA_TYPE_AUDIO, -1, -1, &codec, 0);
+    int streamIndex = av_find_best_stream(formatContext, AVMEDIA_TYPE_AUDIO, -1, -1, nullptr, 0);
     if (streamIndex < 0) {
         av_frame_free(&frame);
         avformat_close_input(&formatContext);
@@ -103,6 +105,9 @@ void SoundManager::SoundSource::loadFromFile(const rwfs::path& filePath) {
     }
 
     AVStream* audioStream = formatContext->streams[streamIndex];
+    AVCodec* codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+
+#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(57,5,0)
     AVCodecContext* codecContext = audioStream->codec;
     codecContext->codec = codec;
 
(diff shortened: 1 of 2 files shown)


rwengine/src/audio/SoundManager.cpp
```diff
@@ -16,6 +18,9 @@ extern "C" {
 #define av_frame_free   avcodec_free_frame
 #endif
 
+constexpr int kNrOfOutputChannels = 2;
```
Lihis: nit: At first look I didn't understand what was the purpose of this, so `kNumOutputChannels` maybe?

rwengine/src/audio/SoundManager.cpp
```diff
@@ -108,9 +113,36 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+        avformat_close_input(&formatContext);
+        RW_ERROR("Couldn't open the audio codec context");
+        return;
+    }
+#endif
 
     // Expose audio metadata
-    channels = codecContext->channels;
+    channels = kNrOfOutputChannels;
     sampleRate = codecContext->sample_rate;
```
Lihis: `WaveHeader` seems to use `uint32_t`, should this reflect it and cast or just cast?

Lihis: `WaveHeader` seems to use `uint32_t`, should this reflect it and cast or just cast?

rwengine/src/audio/SoundManager.cpp
```diff
@@ -108,9 +113,36 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
 
     // OpenAL only supports mono or stereo, so error on more than 2 channels
```
Lihis: Is the `channels > 2` check below necessary as it's `2` anyway?

ShFil119:
Done. ;)

rwengine/src/audio/SoundManager.cpp
```diff
@@ -166,10 +185,82 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+    AVFrame* resampled = nullptr;
+
+    while (av_read_frame(formatContext, &readingPacket) == 0) {
+        if (readingPacket.stream_index == audioStream->index) {
+
+            int sendPacket = avcodec_send_packet(codecContext, &readingPacket);
+            av_packet_unref(&readingPacket);
+            int receiveFrame = 0;
+
+
```
Lihis: Two empty lines?

rwengine/src/audio/SoundManager.cpp
```diff
@@ -166,10 +185,82 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+                        return;
+                    }
+                    swr_init(swr);
+                    if (!swr_is_initialized(swr)) {
+                        RW_ERROR("Resampler has not been properly initialized.");
+                        return;
+                    }
+                }
+                // Decode audio packet
+
```
Lihis: I assume this empty line was supposed to be above the comment, not after it.

rwengine/src/audio/SoundManager.cpp
```diff
@@ -16,6 +18,9 @@ extern "C" {
 #define av_frame_free   avcodec_free_frame
 #endif
 
+constexpr int kNumOutputChannels = 2;
```
Lihis: Why this is now hard-coded?

ShFil119: Mono sound you will hear only from one channel. It will sound weird in binaural mode. Const var is to make clear that it's converted into stereo.

Lihis: Yes all of your reply makes sense.

But does this mean that GTA sound files doesn't have more than 2 channels? Did GTA3 (or VC/SA) support surround sound?

ShFil119: I'm not familiar with VC/SA, but wav files and sfx are mono. Mp3 files are stereo.

Lihis: Thanks, I don't have issues with this then.

ShFil119: If you think about input data, it will (or should) be converted to stereo.
If output data, due to bad quality source it'll sound as normal stereo.

rwengine/src/audio/SoundManager.cpp
ShFil119: Reformatted it a bit. Should now be a little better. :)

rwengine/src/audio/SoundManager.cpp
```diff
@@ -166,10 +185,81 @@ void SoundManager::SoundSource::loadFromFile(const std::string& filename) {
...
+    /// Free resampler
+    swr_free(&swr);
+
+    /// Close the context and free all data associated to it, but not the context itself.
     avcodec_close(codecContext);
+
+    /// Free the context itself.
+    avcodec_free_context(&codecContext);
+
+    /// We are done here. Close the input.
```
danhedron: These comments are just repeating what the code says

danhedron:
Code seems OK. Seems like this will be difficult to verify without the whole sound pipeline running.

rwengine/src/audio/SoundManager.cpp
ShFil119: To remove?

PR #537, opened 2018-07-03
[Ready] Basic traffic

mole99:
This is my first rather long pull request, so criticism is as always appreciated.

The main purpose of this commit is to implement, as already said, basic traffic.
I did change some other sections, which i will explain.

Basically the vehicles spawn similar like the pedestrians, but this time in between two nodes, so that we can set the orientation.

I don't know whether there is also a file like `pedgrp.dat` for vehicles, so for now I hardcoded a list with vehicles to spawn.

A new goal - TrafficDriver was added, and a new Activity - DriveTo.
For the rudimentary control I needed two new functions `isInFront(...)` and `isOnSide(...)` which test whether a point is in front or at the side of the vehicle.

As updating the angle of the steering takes too long for this rudimentary control to work, I changed `setSteeringAngle(...)` to also set directly the steering angle.

In order for the character to plan ahead when driving I added `nextTargetNode` and `lastTargetNode` which can be helpful when improving pedestrian behavior.

`gatherExternalNodesNear(...)` now also takes the NodeType as input, as it is pointless to always sort the list afterwards.

Opcode 00ae was also rudimentary implemented, the police cars at the start of the game will now wander around.

I had to increase the minDist to 15.f from 10.f for the vehicles in order to not spawn in one another.

Also when in debug mode vehicles continue driving but the controller won't update, so the vehicles will miss their nodes and go crazy, maybe vehicles should be prevented moving when in debug mode.

Tasks to implement in the future:

- [ ] Improve current driving behavior
    - [ ] Vehicles sometimes lock
    - [ ] Implement dead end
- [ ] Implement different driving behaviors
- [ ] Vehicles sometimes spawn in view
- [ ] Create list with vehicles to spawn (zones?)
- [ ] Implement opcodes
- [ ] One-way streets

So please test this pull request and give me some feedback.

PS: If you want to have a little fun set the vehicle list to: `std::vector<uint16_t> cars = {131};`

### DIFF
--- a/rwengine/src/ai/AIGraph.cpp
+++ b/rwengine/src/ai/AIGraph.cpp
@@ -1,5 +1,6 @@
 #include "ai/AIGraph.hpp"
 
+#include <algorithm>
 #include <cstddef>
 
 #include <glm/gtx/norm.hpp>
@@ -46,8 +47,8 @@ void AIGraph::createPathNodes(const glm::vec3& position,
             ainode->nextIndex = node.next >= 0 ? startIndex + node.next : -1;
             ainode->flags = AIGraphNode::None;
             ainode->size = node.size;
-            ainode->other_thing = node.other_thing;
-            ainode->other_thing2 = node.other_thing2;
+            ainode->leftLanes = node.leftLanes;
+            ainode->rightLanes = node.rightLanes;
             ainode->position = nodePosition;
             ainode->external = node.type == PathNode::EXTERNAL;
             ainode->disabled = false;
@@ -96,7 +97,8 @@ glm::ivec2 worldToGrid(const glm::vec2& world) {
 
 void AIGraph::gatherExternalNodesNear(const glm::vec3& center,
                                       const float radius,
-                                      std::vector<AIGraphNode*>& nodes) {
+                                      std::vector<AIGraphNode*>& nodes,
+                                      AIGraphNode::NodeType type) {
     // the bounds end up covering more than might fit
     auto planecoords = glm::vec2(center);
     auto minWorld = planecoords - glm::vec2(radius);
@@ -111,11 +113,12 @@ void AIGraph::gatherExternalNodesNear(const glm::vec3& center,
                 continue;
             }
             auto& external = gridNodes[i];
-            for (AIGraphNode* node : external) {
-                if (glm::distance2(center, node->position) < radius * radius) {
-                    nodes.push_back(node);
-                }
-            }
+            copy_if(external.begin(), external.end(), back_inserter(nodes),
+                    [&center, &radius, &type](const AIGraphNode* node) {
+                        return node->type == type &&
+                               glm::distance2(center, node->position) <
+                                   radius * radius;
+                    });
         }
     }
 }
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -99,6 +99,7 @@ void CharacterController::update(float dt) {
         _currentActivity = nullptr;
         if (_nextActivity) {
             setActivity(std::move(_nextActivity));
+            _nextActivity = nullptr;
         }
     }
 }
@@ -139,6 +140,255 @@ bool Activities::GoTo::update(CharacterObject *character,
 
     controller->setMoveDirection({1.f, 0.f, 0.f});
     controller->setRunning(sprint);
+
+    return false;
+}
+
+glm::vec3 CharacterController::calculateRoadTarget(const glm::vec3 &target,
+                                                   const glm::vec3 &start,
+                                                   const glm::vec3 &end) {
+    // @todo set the real value
+    static constexpr float roadWidth = 5.f;
+
+    static const glm::vec3 up = glm::vec3(0.f, 0.f, 1.f);
+    const glm::vec3 dir = glm::normalize(start - end);
+
+    // Calculate the strafe vector
+    glm::vec3 strafe = glm::cross(up, dir);
+
+    const glm::vec3 laneOffset =
+        strafe *
+        (roadWidth / 2 + roadWidth * static_cast<float>(getLane() - 1));
+
+    return target + laneOffset;
+}
+
+void CharacterController::steerTo(const glm::vec3 &target) {
+    // We can't drive without a vehicle
+    VehicleObject* vehicle = character->getCurrentVehicle();
+    if (vehicle == nullptr) {
+        return;
+    }
+
+    // Calculate the steeringAngle
+    float steeringAngle = 0.0f;
+    float deviation = glm::abs(vehicle->isOnSide(target)) / 5.f;
+
+    // If we are almost at the right angle, decrease the deviation to reduce wiggling
+    if (deviation < 1.f) deviation = deviation / 5.f;
+
+    // Make sure to normalize the value
+    deviation = glm::clamp(deviation, 0.f, 1.f);
+
+    // Set the right sign
+    steeringAngle = std::copysign(deviation, -vehicle->isOnSide(target));
+
+    vehicle->setSteeringAngle(steeringAngle, true);  
+}
+
+// @todo replace this by raytest/raycast logic
+bool CharacterController::checkForObstacles()
+{
+    // We can't drive without a vehicle
+    VehicleObject* vehicle = character->getCurrentVehicle();
+    if (vehicle == nullptr) {
+        return false;
+    }
+
+    // The minimal distance we test for objects
+    static constexpr float minColDist = 20.f;
+
+    // Try to stop before pedestrians
+    for (const auto &obj : character->engine->pedestrianPool.objects) {
+        // Verify that the character isn't the driver and is walking
+        if (obj.second != character && ((CharacterObject*)obj.second)->getCurrentVehicle() == nullptr) {
+            // Only check characters that are near our vehicle
+            if (glm::distance(vehicle->getPosition(), obj.second->getPosition()) <= minColDist) {
+                // Check if the character is in front of us and in our way
+                if ( vehicle->isInFront(obj.second->getPosition()) > -3.f 
+                    && vehicle->isInFront(obj.second->getPosition()) < 10.f 
+                    && glm::abs(vehicle->isOnSide(obj.second->getPosition())) < 3.f ) {
+                    return true;
+                }
+            }
+        }
+    }
+
+    // Brake when a car is in front of us and change lanes when possible
+    for (const auto &obj : character->engine->vehiclePool.objects) {
+        // Verify that the vehicle isn't our vehicle
+        if (obj.second != vehicle) {
+            // Only check vehicles that are near our vehicle
+            if (glm::distance(vehicle->getPosition(), obj.second->getPosition()) <= minColDist) {
+                // Check if the vehicle is in front of us and in our way
+                if ( vehicle->isInFront(obj.second->getPosition()) > 0.f 
+                    && vehicle->isInFront(obj.second->getPosition()) < 10.f 
+                    && glm::abs(vehicle->isOnSide(obj.second->getPosition())) < 2.5f ) {
+                    // Check if the road has more than one lane
+                    // @todo we don't know the direction of the road, so for now, choose the bigger value
+                    int maxLanes = targetNode->rightLanes > targetNode->leftLanes ?
+                                    targetNode->rightLanes : targetNode->leftLanes;
+                    if (maxLanes > 1) {
+                        // Change the lane, firstly check if there is an occupant
+                        if (((VehicleObject *)obj.second)->getDriver() !=
+                            nullptr) {
+                            // @todo for now we don't know the lane where the player is currently driving
+                            // so just slow down, in the future calculate the lane
+                            if (((VehicleObject *)obj.second)
+                                    ->getDriver()
+                                    ->isPlayer()) {
+                                return true;
+                            }
+                            else {
+                                int avoidLane = ((VehicleObject *)obj.second)
+                                                    ->getDriver()
+                                                    ->controller->getLane();
+
+                                // @todo for now just two lanes
+                                if (avoidLane == 1) character->controller->setLane(2);
+                                else character->controller->setLane(1);
+                            }
+                        }					
+                    }
+                    else {
+                        return true;
+                    }
+                }
+            }
+        }
+    }
+
+    return false;
+}
+
+bool Activities::DriveTo::update(CharacterObject *character,
+                              CharacterController *controller) {
+
+    // We can't drive without a vehicle
+    VehicleObject* vehicle = character->getCurrentVehicle();
+    if (vehicle == nullptr) {
+        return true;
+    }
+
+    // Get the nodes from the controller
+    AIGraphNode* lastTargetNode = controller->lastTargetNode;
+    AIGraphNode* nextTargetNode = controller->nextTargetNode;
+
+
+    // That's the position the vehicle is actually targeting
+    // depending on the lane, we have to shift its position
+    glm::vec3 roadTarget;
+
+    // A list of nodes we can choose from
+    std::vector<AIGraphNode*> potentialNodes = targetNode->connections;
+
+    // Make sure that we have a lastTargetNode
+    if (lastTargetNode == nullptr) {
+        for (const auto &node : potentialNodes) {
+            if (vehicle->isInFront(node->position) < 0.f) {
+                lastTargetNode = node;
+                break;
+            }
+        }
+    }
+    if (lastTargetNode == nullptr) {
+        return false;
+    }
+
+    // Remove unwanted nodes
+    for( auto i = potentialNodes.begin(); i != potentialNodes.end(); ) {
+        // @todo we don't know the direction of the road, so for now, choose the bigger value
+        int maxLanes = (*i)->rightLanes > (*i)->leftLanes ? (*i)->rightLanes : (*i)->leftLanes;
+
+        // We don't want roads with lanes <= 0, also ignore the lastTargetNode
+        if ( (*i) == lastTargetNode || maxLanes <= 0) {
+            i = potentialNodes.erase(i);
+        }
+        else {
+            ++i;
+        }
+    }
+
+    // That's a dead end, try to turn around
+    if (potentialNodes.empty()) {
+        //@todo try to turn around
+    }
+    // Just a normal road
+    if (potentialNodes.size() == 1) {
+        roadTarget = controller->calculateRoadTarget(
+            targetNode->position, lastTargetNode->position,
+            potentialNodes.at(0)->position);
+    }
+    // Intersection, choose a direction
+    else if (potentialNodes.size() > 1) {
+        // Choose the next node randomly
+        if(nextTargetNode == nullptr) {
+            auto& random = character->engine->randomEngine;
+            int i = std::uniform_int_distribution<>(0, potentialNodes.size() - 1)(random);
+            nextTargetNode = potentialNodes.at(i);
+        }
+
+        // Set the nextTargetNode to make sure we go this direction
+        controller->nextTargetNode = nextTargetNode;
+
+        roadTarget = controller->calculateRoadTarget(targetNode->position,
+                                                     lastTargetNode->position,
+                                                     nextTargetNode->position);
+    }
+    // Otherwise set the target to the current node
+    else {
+        roadTarget = targetNode->position;
+    }
+
+    // Check whether we reached the node
+    const auto targetDistance = glm::vec2(vehicle->getPosition() - roadTarget);
+
+    static constexpr float reachDistance = 5.0f;
+
+    if (glm::length(targetDistance) <= reachDistance) {
+        // Finish the activity
+        return true;
+    } 
+
+    // @todo set real values
+    static constexpr float maximumSpeed = 10.f;
+    static constexpr float intersectionSpeed = 3.5f;
+
+    float currentSpeed = 0.f;
+
+    // Set the speed depending on where we are driving
+    if ( potentialNodes.size() == 1 ) {
+        currentSpeed = maximumSpeed;
+    }
+    else {
+        currentSpeed = intersectionSpeed;
+    }
+
+    // Check whether a pedestrian or vehicle is in our way
+    if (controller->checkForObstacles() == true) {
+        currentSpeed = 0.f;
+    }
+
+    // Is the vehicle slower than it should be
+    if (vehicle->getVelocity() < currentSpeed) {
+        vehicle->setHandbraking(false);
+
+        // The vehicle is driving backwards, accelerate
+        if (vehicle->getVelocity() < 0) {
+            vehicle->setThrottle(1.f);
+        }
+        // Slowly accelerate until we reach the designated speed
+        else {
+            vehicle->setThrottle(1.f - (vehicle->getVelocity() / currentSpeed));
+        }
+    }
+    // We are to fast, activate the handbrake - works better
+    else {
+        vehicle->setHandbraking(true);
+    }
+
+    // Steer to the target
+    controller->steerTo(roadTarget);
 
     return false;
 }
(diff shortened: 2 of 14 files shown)


rwengine/src/ai/AIGraph.cpp
```diff
@@ -113,7 +114,10 @@ void AIGraph::gatherExternalNodesNear(const glm::vec3& center,
             auto& external = gridNodes[i];
             for (AIGraphNode* node : external) {
```
danhedron: This loop looks like a good candidate for using ``copy_if``.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,222 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+    }
+    else {
+        currentSpeed = intersectionSpeed;
+    }
+
+    // The minimal distance we test for objects
+    float minColDist = 20.f;
+
+    // Try to stop before pedestrians
+    for (auto& obj : character->engine->pedestrianPool.objects) {
```
danhedron: This is quite an expensive way of checking for an obstruction. This could use the collision detection system to detect pedestrians/vehicles.

SonofUgly:
>I don't know whether there is also a file like pedgrp.dat for vehicles, so for now I hardcoded a list with vehicles to spawn.

There's a section in `default.ide` that has the spawn frequency and puts the vehicles into classes for where they should spawn.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,222 @@ bool Activities::GoTo::update(CharacterObject *character,
     return false;
 }
 
+bool Activities::DriveTo::update(CharacterObject *character,
```
danhedron: This one is quite long. It looks like there's two main things going on in the function, 1. Find the target position, and 2. steer towards it. These can be their own functions with descriptive names, which means you don't need as many comments.

ShFil119:
Experienced this:
https://gist.github.com/ShFil119/fbd8b13645a91e4a655a828ed171d0be

mole99:
@ShFil119 Thank you for testing! Yes vehicles can lock each other, but in that case it seems like they are in a dead end. I forgot to mention I haven't implemented that yet.

I'll come around IRC some time!

I thought this bug was solved, as it is difficult to reproduce, I'll take a look at this.

rwengine/src/ai/CharacterController.hpp
```diff
@@ -132,6 +142,14 @@ class CharacterController {
     void setMoveDirection(const glm::vec3& movement);
     void setLookDirection(const glm::vec2& look);
 
+    void setLane(int lane) {
+        m_lane = lane;
+    }
+
+    int getLane() {
```
ShFil119: const

rwengine/src/ai/CharacterController.hpp
```diff
@@ -182,6 +200,23 @@ struct GoTo : public CharacterController::Activity {
     }
 };
 
+struct DriveTo : public CharacterController::Activity {
+    DECL_ACTIVITY(DriveTo)
+
+    AIGraphNode* targetNode;
+    bool rampant; // Drive fast, ignore traffic rules @todo
+
+    DriveTo(AIGraphNode* targetNode, bool _rampant = false)
```
ShFil119: targetNode should by default initialized to `nullptr`

ShFil119: Maybe it's the culprit of crash.

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -143,34 +151,119 @@ std::vector<GameObject*> TrafficDirector::populateNearby(
     const auto& group = world->data->pedgroups.at(groupid);
     peds.insert(peds.end(), group.cbegin(), group.cend());
 
-    std::random_device rd;
-    std::default_random_engine re(rd());
-    std::uniform_int_distribution<> d(0, peds.size() - 1);
-    const glm::vec3 kSpawnOffset{0.f, 0.f, 1.f};
+    // Vehicles for normal traffic @todo create correct vehicle list
+    std::vector<uint16_t> cars = {90, 91, 92, 94, 95, 97, 98, 100, 101, 102, 103, 104, 105, 106, 107, 108, 109, 110,
```
ShFil119: It can be `constexpr static std::array`.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,222 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+    // That's the node the vehicle is actually targeting
+    // depending on the lane, we have to shift its position
+    AIGraphNode realTargetNode;
+
+    // A list of nodes we can choose from
+    std::vector<AIGraphNode*> potentialNodes = targetNode->connections;
+
+    // Make sure that we have a lastTargetNode
+    if (lastTargetNode == nullptr) {
+        for ( auto node : potentialNodes) {
```
ShFil119: const ref?

rwengine/src/ai/CharacterController.cpp
mole99: Yes, I see what you mean, do you think of `WorldCollisionDispatcher`?
But in that case the goal is to prevent a collision, which means I have to sense beforehand if a pedestrian or a vehicle is in front of me, right?

ShFil119: Maybe better bullet's raytest/raycast?
See:
http://www.bulletphysics.org/mediawiki-1.5.8/index.php/Using_RayTest

rwengine/src/ai/AIGraph.cpp
mole99: Is it possible to use outer variables for copy_if, because we need to consider the center and the radius?

dracc: Yes, the lambda function can read external variables. The `&` in `[&](){...`will catch all external variables by reference. But preferably, you'd specify the ones used.

rwengine/src/ai/CharacterController.hpp
mole99: Doesn't hurt to try.

rwengine/src/ai/CharacterController.cpp
mole99: I broke it down into more digestible parts, is it okay now?

rwengine/src/ai/AIGraph.cpp
mole99: Thanks, I hope I got it right.

dracc: Sure looks like it! 👍

rwengine/src/ai/CharacterController.cpp
danhedron: Ray testing is one option. Another is a contact test, using ``btCollisionWorld::contactTest`` to check a rectangular area ahead of the vehicle. This would need a bit of work to add a ``GameWorld`` method for doing the actual check, to keep the amount of Bullet that is exposed to objects as low as possible.

We don't know the exact behaviour of the original game (or I don't, at least) so I'm not dead set on a particular solution. Depending on the performance this might be acceptable, with a follow-up issue to make a note that it should be sped up ASAP.

rwengine/src/ai/AIGraph.cpp
```diff
@@ -112,14 +113,8 @@ void AIGraph::gatherExternalNodesNear(const glm::vec3& center,
...
-            for (AIGraphNode* node : external) {
-                if (glm::distance2(center, node->position) < radius * radius) {
-                    if (node->type == type)
-                    {
-                        nodes.push_back(node);
-                    }
-                }
-            }
+            copy_if(external.begin(), external.end(), back_inserter(nodes),
+                    [&center, &radius, &type](const AIGraphNode* node) { return node->type == type && glm::distance2(center, node->position) < radius * radius; });
```
danhedron: This line is too long (>80 columns). The ``.clang-format`` file in the repository specifies the code style that should be used, and can be applied automatically using clang format.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -156,13 +156,121 @@ bool Activities::GoTo::update(CharacterObject *character,
     return false;
 }
 
+glm::vec3 CharacterController::createRoadTarget(glm::vec3 target, glm::vec3 start, glm::vec3 end)
```
danhedron: ``calculateRoadTarget`` ?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2065,7 +2065,7 @@ void opcode_00ae(const ScriptArguments& args, const ScriptVehicle vehicle, const
     {
```
danhedron: I'd consider adding a ``VehicleObject::getDriver()`` that returns a ``CharacterObject*``. As there's also code in the obstacle avoiding method that performs similar checks, this would tidy up some of the casting so it's a bit easier to follow what's going on.

rwengine/src/ai/CharacterController.hpp
```diff
@@ -182,6 +215,23 @@ struct GoTo : public CharacterController::Activity {
     }
 };
 
+struct DriveTo : public CharacterController::Activity {
+    DECL_ACTIVITY(DriveTo)
+
+    AIGraphNode* targetNode;
+    bool rampant; // Drive fast, ignore traffic rules @todo
+
+    DriveTo(AIGraphNode* targetNode = nullptr, bool _rampant = false)
```
ShFil119: I forgot, you can use default member intitializer and set ctor as default.

rwengine/src/ai/DefaultAIController.cpp
```diff
@@ -91,6 +96,116 @@ void DefaultAIController::update(float dt) {
...
+                        targetNode, false));
+                }
+            }
+            else {
+                // We need to pick an initial node
+                auto& graph = getCharacter()->engine->aigraph;
+                AIGraphNode* node = nullptr;
+                float mindist = std::numeric_limits<float>::max();
+
+                for (auto n : graph.nodes) {
```
ShFil119: Also const ref?

rwengine/src/ai/CharacterController.cpp
ShFil119: btw it can be const.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,248 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+        if ( (*i) == lastTargetNode || maxLanes <= 0) {
+            i = potentialNodes.erase(i);
+        }
+        else {
+            ++i;
+        }
+    }
+
+    // That's a dead end, try to turn around
+    if (potentialNodes.size() == 0) {
```
ShFil119: empty

rwengine/src/ai/DefaultAIController.cpp
```diff
@@ -91,6 +96,116 @@ void DefaultAIController::update(float dt) {
...
+
+                    // Assign the next target node, either it is already set, 
+                    // or we have to find one by ourselves
+                    if (nextTargetNode != nullptr) {
+                        targetNode = nextTargetNode;
+                        nextTargetNode = nullptr;
+                    }
+                    else {
+                        float mindist = std::numeric_limits<float>::max();
+                        for ( auto node : lastTargetNode->connections ) {
```
ShFil119: Also?

rwengine/src/ai/AIGraph.cpp
```diff
@@ -1,6 +1,7 @@
 #include "ai/AIGraph.hpp"
 
 #include <cstddef>
+#include <algorithm>
```
ShFil119: order

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,248 @@ bool Activities::GoTo::update(CharacterObject *character,
     return false;
 }
 
+glm::vec3 CharacterController::createRoadTarget(glm::vec3 target, glm::vec3 start, glm::vec3 end)
+{
+    // @todo set the real value
+    const float roadWidth = 5.f;
```
ShFil119: `constexpr static float`

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -57,6 +59,14 @@ std::vector<AIGraphNode*> TrafficDirector::findAvailableNodes(
             }
         }
 
+        for (auto& obj : world->vehiclePool.objects) {
```
ShFil119: const?

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,248 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+bool CharacterController::checkForObstacles()
+{
+    // We can't drive without a vehicle
+    VehicleObject* vehicle = character->getCurrentVehicle();
+    if (vehicle == nullptr) {
+        return false;
+    }
+
+    // The minimal distance we test for objects
+    float minColDist = 20.f;
```
ShFil119: static constexpr

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,248 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+        if (obj.second != vehicle) {
+            // Only check vehicles that are near our vehicle
+            if (glm::distance(vehicle->getPosition(), obj.second->getPosition()) <= minColDist) {
+                // Check if the vehicle is in front of us and in our way
+                if ( vehicle->isInFront(obj.second->getPosition()) > 0.f 
+                    && vehicle->isInFront(obj.second->getPosition()) < 10.f 
+                    && glm::abs(vehicle->isOnSide(obj.second->getPosition())) < 2.5f ) {
+                    // Check if the road has more than one lane
+                    // @todo we don't know the direction of the road, so for now, choose the bigger value
+                    int maxLanes = targetNode->rightLanes > targetNode->leftLanes ? targetNode->rightLanes : targetNode->leftLanes;
```
ShFil119: It needs to reformatted.

rwengine/src/ai/DefaultAIController.cpp
```diff
@@ -91,6 +96,116 @@ void DefaultAIController::update(float dt) {
...
+                            float distance = getCharacter()->getCurrentVehicle()->isInFront(node->position);
```
ShFil119: `distance`, `lastDistance`, `d`(139 and 191) can be const

rwengine/src/ai/CharacterController.cpp
```diff
@@ -153,6 +156,248 @@ bool Activities::GoTo::update(CharacterObject *character,
...
+    // @todo set the real value
+    const float roadWidth = 5.f;
+
+    glm::vec3 up = glm::vec3(0.f, 0.f, 1.f);
+    glm::vec3 dir = glm::normalize(start - end);
+
+    // Calculate the strafe vector
+    glm::vec3 strafe = glm::cross(up, dir);
+
+    glm::vec3 laneOffset = strafe * ( roadWidth / 2 + roadWidth * (float)(getLane() - 1) );
```
ShFil119: It's better to use `static_cast`.

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -143,34 +151,119 @@ std::vector<GameObject*> TrafficDirector::populateNearby(
...
+            int maxLanes = spawn->rightLanes < spawn->leftLanes ? spawn->rightLanes : spawn->leftLanes;
+	
+            // This street has no lanes, continue
+            if( maxLanes <= 0 ) {
+                continue;
+            }
+
+            // Choose a random lane
+            int lane = std::uniform_int_distribution<>(1, maxLanes)(random);
+            glm::vec3 laneOffset = strafe * ( 2.5f + 5.f * (float)(lane - 1) );
```
ShFil119: const

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -913,3 +923,64 @@ void VehicleObject::grantOccupantRewards(CharacterObject* character) {
...
+    // The point we need to test
+    glm::vec3 testPoint;
+
+    testPoint = point - getPosition();
+
+    // The two endpoints of the line
+    glm::vec3 v1;
+    glm::vec3 v2;
+
+    glm::vec3 up = glm::vec3(0.f, 0.f, 1.f);
```
ShFil119: Here `up`, `dir` and `vecTemp`.

rwengine/src/script/ScriptFunctions.hpp
```diff
@@ -240,22 +240,20 @@ inline void getClosestNode(const ScriptArguments& args, ScriptVec3& coord, AIGra
...
-            if (dist < closest) {
-                closest = dist;
-                xCoord = node->position.x;
-                yCoord = node->position.y;
-                zCoord = node->position.z;
-            }
+        // This is how the original game calculates distance,
+        // weighted manhattan-distance where the vertical distance
+        // has to be 3x as close to be considered.
+        float dist  = std::abs(coord.x - node->position.x);
```
ShFil119: It can be calculated in one long "operation". And set as const.

rwgame/RWGame.cpp
```diff
@@ -755,6 +755,23 @@ void RWGame::renderDebugPaths(float time) {
         debug.drawLine(position, position + left, color);
     }
 
+    // Draw the targetNode if a character is driving a vehicle
+    for (auto& p : world->pedestrianPool.objects) {
+        auto v = static_cast<CharacterObject*>(p.second);
+
+        btVector3 color(1.f, 1.f, 0.f);
```
ShFil119: static const?

rwgame/RWGame.cpp
```diff
@@ -755,6 +755,23 @@ void RWGame::renderDebugPaths(float time) {
...
+
+        if (v->controller->targetNode && v->getCurrentVehicle()) {
+            glm::vec3 pos1 = v->getPosition();
```
ShFil119: Here and below const.

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -913,3 +923,64 @@ void VehicleObject::grantOccupantRewards(CharacterObject* character) {
...
```
ShFil119: Please check this function also.

rwengine/src/ai/CharacterController.cpp
mole99: @danhedron When merged I will of course create an issue containing all future tasks.

ShFil119: @danhedron, do you think improvements of algorithm for checking should be included in this pull?
 (At least on my laptop (i5 2520m hd3000) fps drop is quite big)

mole99: I didn't know the performance impact is that great, but merging without collision avoidance would mean accidents on every intersection!
It is actually rather easy to implement a raytest, as there is already the class `VehicleRaycaster` in `VehicleObject`. The problem however is, that the function returns a `btRigidBody` and no `GameObject`, which I would need for lane changing.
Unfortunately I won't get any work done this weekend.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -143,6 +144,255 @@ bool Activities::GoTo::update(CharacterObject *character,
     return false;
 }
 
+glm::vec3 CharacterController::calculateRoadTarget(const glm::vec3 &target,
+                                                   const glm::vec3 &start,
+                                                   const glm::vec3 &end) {
+    // @todo set the real value
+    static constexpr float roadWidth = 5.f;
+
+    const glm::vec3 up = glm::vec3(0.f, 0.f, 1.f);
```
ShFil119: Not sure If I overlooked it earlier, but `up` can also `static`.

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -139,34 +147,125 @@ std::vector<GameObject*> TrafficDirector::populateNearby(
...
-    int counter = availablePeds;
-    // maxSpawn can be -1 for "as many as possible"
-    if (maxSpawn > -1) {
-        counter = std::min(availablePeds, maxSpawn);
-    }
+    auto availablePedsNodes = findAvailableNodes(AIGraphNode::Pedestrian, camera, radius);
+
+    // We have not reached the limit of spawned pedestrians
+    if (availablePeds > 0) {
+        const glm::vec3 kSpawnOffset{0.f, 0.f, 1.f};
```
ShFil119: static

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -139,34 +147,125 @@ std::vector<GameObject*> TrafficDirector::populateNearby(
...
+            created.push_back(ped);
+        }
+    }
+
+    auto availableVehicleNodes = findAvailableNodes(AIGraphNode::Vehicle, camera, radius);
+
+    // We have not reached the limit of spawned vehicles
+    if (availableCars > 0) {
+      
+        const glm::vec3 kSpawnOffset{0.f, 0.f, 1.f};
```
ShFil119: static

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -905,3 +919,66 @@ void VehicleObject::grantOccupantRewards(CharacterObject* character) {
...
+    // The point we need to test
+    glm::vec3 testPoint;
+
+    testPoint = point - getPosition();
+
+    // The two endpoints of the line
+    glm::vec3 v1;
+    glm::vec3 v2;
+
+    const glm::vec3 up = glm::vec3(0.f, 0.f, 1.f);
```
ShFil119: static

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -905,3 +919,66 @@ void VehicleObject::grantOccupantRewards(CharacterObject* character) {
...
```
ShFil119: static

rwengine/src/ai/TrafficDirector.cpp
```diff
@@ -139,34 +147,125 @@ std::vector<GameObject*> TrafficDirector::populateNearby(
...
+            // This street has no lanes, continue
+            if( maxLanes <= 0 ) {
+                continue;
+            }
+
+            // Choose a random lane
+            const int lane =
+                std::uniform_int_distribution<>(1, maxLanes)(random);
+            const glm::vec3 laneOffset =
+                strafe * (2.5f + 5.f * (float)(lane - 1));
```
ShFil119: It's better to use `static_cast`.

rwengine/src/ai/CharacterController.cpp
ShFil119: Sorry, problem is also on master. I'll make issue, earlier I's getting 60fps on release, now 20-30. :(

mole99: Nevermind, but I will still try to implement raytest, it should be done right at the first try :+1:

ShFil119: I think it's better to put it off. And do it in next pull (when this will be merged).

mole99: Then it's ready to merge.

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -2057,9 +2057,16 @@ void opcode_00ad(const ScriptArguments& args, const ScriptVehicle vehicle, const
 */
 void opcode_00ae(const ScriptArguments& args, const ScriptVehicle vehicle, const ScriptDrivingMode arg2) {
     RW_UNIMPLEMENTED_OPCODE(0x00ae);
-    RW_UNUSED(vehicle);
     RW_UNUSED(arg2);
     RW_UNUSED(args);
+
+    // Check whether we have a driver
+    if (vehicle->getOccupant(0) != nullptr)
```
husho: ```getDriver``` ?

mole99: Sorry, I forgot to change that.

danhedron:
This needs a rebase before it can be merged.

@mole99 Are you able to catch up your changes?

mole99:
This took a while, but it should be rebased now.
@danhedron Does it look acceptable? Concerning the future tasks, do you want to formulate the issues or should I write them?

danhedron:
I'll leave the implementation accuracy issues to you. I will update the performance issue.

PR #540, opened 2018-07-06
Remove unneeded FileHandle (aka shared_ptr)

ShFil119:
Been trying to convert shared_ptr to unique_ptr, but sometimes even there's no need for pointers.

Tested with address sanitizers in game, but brakes one data test:
https://gist.github.com/ShFil119/c82d46c5f548a5fd97d50194749c7770
(not sure how to fix/refactor it)
Edit:
Fixed

### DIFF
--- a/tests/test_FileIndex.cpp
+++ b/tests/test_FileIndex.cpp
@@ -1,4 +1,5 @@
 #include <boost/test/unit_test.hpp>
+#include <platform/FileHandle.hpp>
 #include <platform/FileIndex.hpp>
 #include "test_Globals.hpp"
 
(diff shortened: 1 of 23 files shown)


tests/test_FileIndex.cpp
```diff
@@ -59,14 +60,14 @@ BOOST_AUTO_TEST_CASE(test_indexArchive) {
...
-        BOOST_CHECK(handle == nullptr);
+        BOOST_CHECK(handle.data == nullptr);
     }
 
     index.indexArchive("models/gta3.img");
 
     {
         auto handle = index.openFile("landstal.dff");
-        BOOST_CHECK(handle != nullptr);
+        BOOST_CHECK(handle.data != nullptr);
```
ShFil119: Here fails.

ShFil119: Fixed.

ShFil119:
Should be ok now.

Issue #544, opened 2018-07-08
OpenGL::GL won't link


mole99:
The current master won't build on my system anymore, it fails to link OpenGL::GL.

I have debian installed as operating system:

`Linux debian-dell 4.9.0-6-amd64 #1 SMP Debian 4.9.88-1+deb9u1 (2018-05-07) x86_64 GNU/Linux`

The output is as follows:

cmake - https://gist.github.com/mole99/5a67a77fe4571b1e2a7a26226a9ae4aa
make - https://gist.github.com/mole99/180ba08c76134d0477ad683ee0b72e40

Changing line 56 in `/rwlib/CMakeLists.txt` from `OpenGL::GL` to `GL` solves the issue on my system, but I don't know if it breaks something for someone else.

progician:
What is you CMake version? The imported target was introduced in CMake 3.8.

progician:
Actually, while OpenRW is using OpenGL::GL imported targets introduced in CMake 3.8, the cmake minimum version is set to 3.2.2 so this could happen easily.

mole99:
```
cmake --version
cmake version 3.7.2
```

I see, my version is too low, when were the cmake files changed? OpenRW successfully compiled the last time I tried.

progician:
it looks like it changed int this commit: https://github.com/rwengine/openrw/commit/9231ecbcb00037979d14013b5fe1f5c60cfee89a

Lihis:
Tested on Debian release `stretch` and got the same error. I assume @mole99 is on Debian stretch too based on his cmake version.

mole99:
Yes, I forgot to mention I'm on Debian stretch. To solve this issue for me I installed a newer cmake version via the stretch-backports.

We can close this issue, when the minimum cmake version gets updated.

PR #552, opened 2018-07-17
Update cmake minimum required version to 3.8

NFSMONSTR:
OpenRW uses OpenGL::GL imported targets that introduced in CMake 3.8, so we [need](https://github.com/rwengine/openrw/issues/544#issuecomment-403313168) to update `cmake_minimum_required` to 3.8.
Also as said @madebr in IRC the findBoost package also has imported targets since that cmake version.

For Debian Stretch users:
In Debian repositories cmake version still lower than 3.8, so to build OpenRW you will [need](https://github.com/rwengine/openrw/issues/544#issuecomment-403395680) to install newer version of cmake from stretch-backports

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2.2)
+cmake_minimum_required(VERSION 3.8)
 
 project(OpenRW)
 


madebr:
LGTM

PR #559, opened 2018-07-27
Remove old style casts

ShFil119:
Cleanup. ;) All old casts are removed.

To do:
- [x] reformat
- [x] if possible, remove now revealed nasty code. (done?)
- [x] add flag `-Wold-style-cast` on CI, actually there's problem `cc1: warning: command line option ‘-Wold-style-cast’ is valid for C++/ObjC++ but not for C [enabled by default]`

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -201,13 +201,17 @@ bool CharacterController::checkForObstacles()
     // Try to stop before pedestrians
     for (const auto &obj : character->engine->pedestrianPool.objects) {
         // Verify that the character isn't the driver and is walking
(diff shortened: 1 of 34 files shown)


rwengine/src/ai/CharacterController.cpp
```diff
@@ -201,7 +201,7 @@ bool CharacterController::checkForObstacles()
     // Try to stop before pedestrians
     for (const auto &obj : character->engine->pedestrianPool.objects) {
         // Verify that the character isn't the driver and is walking
-        if (obj.second != character && ((CharacterObject*)obj.second)->getCurrentVehicle() == nullptr) {
+        if (obj.second != character && reinterpret_cast<CharacterObject*>(obj.second)->getCurrentVehicle() == nullptr) {
```
danhedron: The base to derived class conversion should be static_cast.

danhedron:
LGTM

PR #566, opened 2018-08-02
Various vehicle improvements

paulkocialkowski:
This is an attempt at making vehicles more usable in the game, with improvements to various aspects of driving.

### DIFF
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -310,7 +310,10 @@ void VehicleObject::tickPhysics(float dt) {
         float velocity = collision->getBulletBody()->getLinearVelocity().length();
         float velocityForward = physVehicle->getCurrentSpeedKmHour() / 3.6f;
         float velocityMax = velFac / 9.f;
-        float engineForce = info->handling.acceleration * throttle * velFac;
+        // The engine force is calculated from the acceleration and max velocity
+        // of the vehicle, with a specific coefficient to make it adapted to
+        // Bullet physics and avoid reaching top speed too fast.
+        float engineForce = info->handling.acceleration * throttle * velFac / 1.2f;
         float brakeF = getBraking();
         // Mass coefficient, that quantifies how heavy a vehicle is and excludes
         // light vehicles.
--- a/rwengine/src/objects/VehicleObject.cpp
+++ b/rwengine/src/objects/VehicleObject.cpp
@@ -104,7 +104,20 @@ VehicleObject::VehicleObject(GameWorld* engine, const glm::vec3& pos,
 
     float travel = fabs(info->handling.suspensionUpperLimit -
                         info->handling.suspensionLowerLimit);
-    tuning.m_frictionSlip = 3.f;
+    float maxVelocity = info->handling.maxVelocity;
+    float accelerationFloor = std::max(info->handling.acceleration, 30.f);
+    float massFloor = std::min(info->handling.mass, 3000.f);
+    // The friction slip represents the friction coefficient between each wheel
+    // and the ground. The higher the coefficient, the more slippery the contact
+    // gets between the wheel and the ground. Fast vehicles (with a significant
+    // acceleration and max speed for a low mass) tend to need a high
+    // coefficient (> 5.f) to allow properly turning when steering. On the other
+    // hand, slow vehicles tend to need a lower coefficient or they will be too
+    // reactive when turning. This (purely empirical) formula is an attempt to
+    // apply this idea, with floors to avoid too much divergence.
+    // For some reason, the calculation with handling info gives us a value in
+    // the right range, with an offset to set a sane base.
+    tuning.m_frictionSlip = 1.8f + maxVelocity * accelerationFloor / massFloor;
     tuning.m_maxSuspensionTravelCm = travel * 100.f;
 
     physVehicle =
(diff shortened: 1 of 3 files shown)


paulkocialkowski:
@Lihis: Thanks for the test!

Actually, there was no explicit top speed before (only the limitation given by Bullet Physics) which caused a wide range of issues. So with speed capping in place, the top speed indeed feels reduced. Note that there is an adaptation coefficient for translating the handling info max velocity to our physics, so that can be tweaked to get higher top speeds.

Good catch on the issue when starting up without steering, it's my mistake: I had forgotten to check that there is no throttle applied when zero-ing velocity on the x axis for stopped vehicles in the last commit. I've updated it now and tested that it goes fine.

mole99:
Awesome, while testing I have not managed to flip my car during normal driving!

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -307,11 +323,78 @@ void VehicleObject::tickPhysics(float dt) {
...
+        }
 
         if (handbrake) {
             brakeF = 5.f;
+        } else if (throttle < 0.f && velocityForward > velocityMax / 8.f) {
+            engineForce = 0.f;
+            brakeF = 2.f * std::min(1.f + kM, 4.f);
+        }
+
+        if (isStopped() && throttle == 0.f) {
```
ShFil119: I guess you should use something like : `throttle <= 0.1f`

paulkocialkowski: I don't think that would be right. Throttle is set to an integer value (either -1.f, 0.f or 1.f) so there is about no need for setting a threshold. In addition, what we're really interested in here is the velocity, which is checked by isStopped. The throttle condition is really just a way to allow the car to start again (it can hardly happen if the x coordinate of the velocity vector remains 0, since it makes it hard for the two other coordinate to compensate until reaching the point where isStopped() == false). When there is no driver interaction with the vehicle, throttle is effectively equal to zero.

ShFil119: I was thinking more about problem: `comparing floating point with == or != is unsafe`.

dracc: Integer value, no controller support?

danhedron:
These changes help a lot.

My only change would be to use a threshold on the throttle value, so it continues to function correctly into the future.

rwengine/src/objects/VehicleObject.cpp
paulkocialkowski: Okay, fair enough, I'll change that.

paulkocialkowski:
I implemented the threshold on the throttle value!

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -385,6 +385,18 @@ void VehicleObject::tickPhysics(float dt) {
             brakeF = 2.f * std::min(1.f + kM, 4.f);
         }
 
+        if (isStopped() && std::abs(throttle) < 0.1f) {
+            btVector3 v = collision->getBulletBody()->getLinearVelocity();
+            v.setX(0.f);
```
mole99: I think you also want to set the y axis to zero, or else the vehicle is possible to drift in one direction.

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -385,6 +385,18 @@ void VehicleObject::tickPhysics(float dt) {
...
+
+            collision->getBulletBody()->setLinearVelocity(v);
+
+            for (int w = 0; w < physVehicle->getNumWheels(); ++w) {
+                btWheelInfo& wi = physVehicle->getWheelInfo(w);
+                wi.m_rotation = 0.f;
```
mole99: The problem with setting the wheel angle to zero is, that it visibly snaps into that position. During the transition from Moving to Stopping, the current wheel angle should be kept.

paulkocialkowski: Ah right, I hadn't thought of/witnessed that.

rwengine/src/objects/VehicleObject.cpp
paulkocialkowski: Good point, thanks!

mole99: Glad I could help :+1:

mole99:
LGTM.

rwengine/src/objects/VehicleObject.hpp
```diff
@@ -7,6 +7,7 @@
 #include <memory>
 #include <string>
 
+#include <BulletDynamics/Vehicle/btRaycastVehicle.h>
```
ShFil119: You only use here btScalar.

rwengine/src/objects/VehicleObject.hpp
```diff
@@ -36,6 +37,7 @@ class VehicleObject final : public GameObject, public ClumpObject {
     float throttle{0.f};
     float brake{0.f};
     bool handbrake = true;
+    btScalar *wheelsRotation;
```
ShFil119: It asks for `std::unique_ptr`.

danhedron: Rather ``std::vector``

ShFil119: Good point. ;)

rwengine/src/objects/VehicleObject.hpp
paulkocialkowski: Fair enough and done.

rwengine/src/objects/VehicleObject.hpp
paulkocialkowski: Thanks, just pushed an updated version with std::vector.

PR #573, opened 2018-08-07
Refactor VisualFX and fix memory leaks

ShFil119:
Actually explosion works, because memory are leaking.

Pull fixes leaks and hidden problem of life time of explosion.
It's still far from optimal.

There's need to store handlers in two places: Pickup/ProjectileObject (ofc) and in vector for rendering.
There three options for dealing with problem.

My approach stores in Objects ref and unique_ptr in vector.
+ it's pretty elastic solution
- dealing with references

Vice versa:
+ less dealing with references
- dealing with unique ptr(maybe even worse than reference)

https://github.com/ShFil119/openrw/pull/40

Rendering when iterating on objects:
- increase dependency between renderer and objects.


ShFil119:
After some thoughts:
- it's easiest to keep unique_ptr/owner/dictator of life time of visual effect in vector in game world,
- if there's need to kill visual effect when object (here PickupObject) is destroyed, then we keep reference to it and call destroyEffect in dtor. https://github.com/rwengine/openrw/pull/573/files#diff-080d4bae0d3eaa0d52dff7f34a7c5773R164
- if there's no need to and effect should live its own live, then we don't keep reference and loop "https://github.com/rwengine/openrw/pull/573/files#diff-26697d1e896ea1037bd13c3ee148af81R528" do work for us. 

Please review, it's possible that I forgot about something. ;)

PR #574, opened 2018-08-08
Rename rwlib library to "core" to fit its new role

danhedron:
As we consider ``librw``, we need to move our rwlib out of the way to avoid confusion.

This changes our library to be called ``core``, as it will remain to house low level utilities and OS abstractions that OpenRW is still using.

Future work will look at which pieces of this library are likely to become redundant and can be removed.


madebr:
I would remove the `source` subfolder from the new library. (Same as the `src` subdirectory of rwengine)
And maybe name the new library `rwcore`? I love that prefix.

core/CMakeLists.txt
```diff
@@ -65,15 +58,15 @@ target_link_libraries(rwlib
         OpenGL::GL
     )
 
-set_target_properties(rwlib
+set_target_properties(core
         PROPERTIES
             OUTPUT_NAME rw
```
madebr: These lines can be removed. The output name `libcore`  (or `librwcore`) is fine. (Where librwlib was weird)

CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ endif()
 # Create a rw_interface TARGET that holds all compiler options
 include("${PROJECT_SOURCE_DIR}/cmake_configure.cmake")
 
-add_subdirectory(rwlib)
+add_subdirectory(core)
```
madebr: Only remark(s) is to change this to rwcore too. Otherwise LGTM.

PR #575, opened 2018-08-09
project: use GLEW (OpenGL Extension Wrangler)

madebr:
Use GLEW as OpenGL loader instead of the generated `rwlib/source/gl/gl_core_3_3.[ch]` files.

- [ ] Fails on mac due to undeclared identiefiers: https://travis-ci.org/madebr/openrw/jobs/413838559#L3715

librw uses GLEW


madebr:
This will be addressed later.

PR #579, opened 2018-08-10
wrap QOpenGLContext

madebr:
Get rid of the `OpenGLCompat.h` file for mac os by wrapping/hiding `QOpenGLContext`.


christoph-heiss:
LGTM, just tested this on macOS 10.13.6, did not notice any difference.

(also thanks for removing my hacky solution.)

PR #580, opened 2018-08-10
[Discussion] Remove setPosition() offset in CharacterObject

mole99:
Currently `setPosition()` in `CharacterObject` warps the character one unit higher than the actual hight.

I think this behavior was intended for testing, but isn't necessary anymore.

Removing the offset helps in a variety of situations, mostly where `GoTo()` was used:

- Pedestrians don't "jump" anymore when reaching an `AIGraphNode`
- During missions with various characters
- AI followers
- ...

I tested the change in debug mode and various missions and I found no issues.

Maybe some testers can confirm that everything works as intended.

### DIFF
--- a/rwengine/src/objects/CharacterObject.cpp
+++ b/rwengine/src/objects/CharacterObject.cpp
@@ -427,7 +427,7 @@ void CharacterObject::setPosition(const glm::vec3& pos) {
         if (pos.z <= -99.f) {
             realPos = engine->getGroundAtPosition(pos);
         }
-        btVector3 bpos(realPos.x, realPos.y, realPos.z + 1.0f);
+        btVector3 bpos(realPos.x, realPos.y, realPos.z);
         physCharacter->warp(bpos);
     }
     position = realPos;


danhedron:
The offset is because (as far as I'm aware) OpenRW is wrong about the position of characters. The game expects it relative to their feet and places them at such, but OpenRW is taking their position from the center of the body. The most obvious consequence of this was loading a save game resulted in the player falling through the ground.

The solution to the problem is to adjust the position of the collision and model to be relative to the character's feet.

mole99:
Thanks for the clarification, I haven't thought of that!

So the current problem is, that setPosition() uses an offset, but getPosition() not.

I'll close that pull request, maybe I'll come up with another solution.

mole99:
Sorry I think I closed this one too fast!

So it works, uh? But nevertheless it isn't the right behaviour.

I'll open a pull request, which doesn't fix the main problem, but adjusts the height in `GoTo()` which fixes the "jumping" and uses the correct values from the `btCapsuleShape()`.

I think if we want to do this right we would need to overwrite `getPosition()`?

danhedron:
Just to clarify, removing the offset fixes the jumping issue. It's only when a save game is loaded and the game tries to move the player, depending on the safehouse the player will fall through the ground. There are a few other hacks in the SCM that offset positions that will need to be un-done to fix the issue properly.

The correct solution is to change the character position to be relative to the model's feet. This could be done with a btCompoundShape to move the body up by 1/2 its height. The model's position will need to be adjusted also.

PR #586, opened 2018-08-12
Sfx - ScriptObject, metadata, opcodes, listener position

ShFil119:
Done:
- [x] Using path in sound manager
- [x] Sfx metadata
- [x] Grabbing listener position
- [x] Sound as object in script
- [x] opcodes
- [x] deal with formatting. ;)

Reviews are welcome. ;)

### DIFF
--- a/rwcore/loaders/LoaderSDT.hpp
+++ b/rwcore/loaders/LoaderSDT.hpp
@@ -62,8 +62,7 @@ public:
     ~LoaderSDT() = default;
 
     /// Load the structure of the archive
-    /// Omit the extension in filename
-    bool load(const rwfs::path& path);
+    bool load(const rwfs::path& sdtPath, const rwfs::path& rawPath);
 
     /// Load a file from the archive to memory and pass a pointer to it
     /// Warning: Returns nullptr if by any reason it can't load the file
--- a/rwengine/src/audio/SfxParameters.cpp
+++ b/rwengine/src/audio/SfxParameters.cpp
@@ -1,0 +1,160 @@
+#include "audio/SfxParameters.hpp"
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+
+namespace {
+/// Hardcoded array for translating script index into sfx index
+/// Valid for gta3.
+/// (In origin it also had been hardcoded)
+/// (-1 means unlimited)
+/// @todo some of them should return
+/// random(?) number in range
+/// see comment in second column
+/// @todo: preparing data for VC and SA
+constexpr SoundInstanceData sfxData[] = {
+    /// for opcode 018c
+    {78, 2857 /*-2871*/, -1},  /// male pain soft
+    {79, 2857 /*-2871*/, -1},  /// male pain loud
+    {80, 2290 /*-2300*/, -1},  /// female pain soft
+    {81, 2290 /*-2300*/, -1},  /// female pain loud
+    {82, 161 /*-162*/, -1},    /// item pickup
+    {83, 161 /*-162*/, -1},    /// item pickup
+    {92, 147, 40},             /// gate start
+    {93, 147, 40},             /// gate stop
+    {94, 337, -1},             /// checkpoint
+    {97, 336, -1},             /// countdown timer 3
+    {98, 336, -1},             /// countdown timer 2
+    {99, 336, -1},             /// countdown timer 1
+    {100, 337, -1},            /// countdown finish
+    {106, 176, 50},            /// bullet hit ground
+    {107, 177, 50},
+    {108, 178, 50},
+    {110, 389, 80},  /// silent
+    {111, 389, 80},
+    {112, 283, 80},           /// payphone
+    {114, 151, 60},           /// glass break
+    {115, 151, 60},           /// glass break
+    {116, 150, 60},           /// glass damage
+    {117, 152 /*-155*/, 55},  /// glass shards
+    {118, 327, 60},           /// boxes destroyed
+    {119, 328, 60},           /// boxes destroyed
+    {120, 140 /*-144*/, 60},  /// car collision
+    {121, 29, 60},            /// tire collision
+    {122, 167 /*-168*/, 20},  /// gun shell drop,
+                              /// dependent on collision
+                              /// the player is currently
+                              /// standing on
+    {123, 168, 20},           /// gun shell drop soft
+    /// for opcode 018d
+    {4, 390, 30},
+    {5, 390, 80},  /// Hepburn Heights
+                   /// southwest tower
+    {6, 391, 30},
+    {7, 391, 80},  /// Hepburn Heights
+                   /// north tower
+    {8, 392, 30},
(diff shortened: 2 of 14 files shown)


rwengine/src/audio/SoundManager.hpp
```diff
@@ -26,14 +28,14 @@
 /// instances simultaneously without duplicating raw source).
 class SoundManager {
 public:
-    SoundManager();
+    SoundManager(GameData* _data);
```
danhedron: Member variables should start with an underscore.

rwengine/src/audio/SoundSource.cpp
```diff
@@ -254,17 +254,14 @@ static int read_packet(void* opaque, uint8_t* buf, int buf_size) {
     return buf_size;
 }
 
-void SoundSource::loadSfx(const rwfs::path& path, LoaderSDT& sdt, size_t index, bool asWave) {
+void SoundSource::loadSfx(LoaderSDT& sdt, size_t index, bool asWave) {
```
danhedron: Can the Loader be const here? Just wondering.

rwgame/RWGame.cpp
```diff
@@ -601,6 +601,13 @@ void RWGame::render(float alpha, float time) {
         viewCam.frustum.fov *= viewCam.frustum.aspectRatio;
     }
 
+    // Update Listener parameters
+    // @todo ShFil119 it should be improved
+    world->sound.setListenerPosition(currentCam.position);
```
danhedron: Wrap in `SoundManager::updateListenerTransform` ?

rwengine/src/audio/SoundManager.hpp
ShFil119: Which version do you prefer for member `_data`, `m_data`, `data_`?

danhedron: `_data`

rwengine/src/audio/SoundSource.cpp
ShFil119: Cannot.

ShFil119:
Up. :v:

mole99:
I tested this PR on various platforms:

Windows 10 - works without any problems
Debian Stretch (stable) - no sound (even before this PR) - ffmpeg 3.2.12-1
Debian Buster (testing) - cutscenes have sound, also during missions, but no sfx - ffmpeg 4.0.2-1

Also when pressing ESC in buster (maybe also stretch, haven't tested), the game crashes.

Let me know if you need more specific information, sadly I'm currently not in reach of debian stretch!

ShFil119:
About no sound in Debian Stretch please open issue with log from debug build.

I'm using Arch linux with 4.0.2-1, I will try to seek source of additional problems. ;)

mole99:
Will do, thanks for that! Arch Linux, I tried that too once ;)

rwengine/src/audio/SoundSource.cpp
ShFil119: Why? I need raw_data  to live enough for feeding ffmpeg, it requires life time longer than function's one, so I store it in "body" of sdt.

rwengine/src/audio/SfxParameters.cpp
```diff
@@ -0,0 +1,178 @@
...
+    {102, 157, 50}           /// Callahan Bridge fire
+};
+}  // namespace
+
+int convertScriptIndexIntoSfx(int scriptId) {
+    const auto finded = std::find_if(std::begin(sfxData), std::end(sfxData),
+                                     [&scriptId](const SoundInstanceData data) {
+                                         return data.id == scriptId;
+                                     });
+    return finded->sfx;
```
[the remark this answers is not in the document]
ShFil119: yeah, should be `found`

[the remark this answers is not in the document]
ShFil119: Finding nothing shouldn't happen, but I guess I should add default case.

rwengine/src/audio/SoundSource.cpp
madebr: I was refering to this "Little Britain" meme
https://www.youtube.com/watch?v=AJQ3TM-p2QI

rwengine/src/audio/SfxParameters.hpp
```diff
@@ -0,0 +1,16 @@
+#ifndef _RWENGINE_SFX_PARAMETERS_HPP_
+#define _RWENGINE_SFX_PARAMETERS_HPP_
+
+#include <cstddef>
+
+/// Nr of existing sfx sounds in gta3
+/// @todo preparing data for VC and SA
+constexpr size_t kNrOfAllSfx = 3032;
```
danhedron: Is this used anywhere?

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -4280,10 +4281,9 @@ void opcode_018b(const ScriptArguments& args, const ScriptBlip blip, const Scrip
     @arg sound 
 */
 void opcode_018c(const ScriptArguments& args, ScriptVec3 coord, const ScriptSoundType sound) {
-    RW_UNIMPLEMENTED_OPCODE(0x018c);
-    RW_UNUSED(coord);
-    RW_UNUSED(sound);
-    RW_UNUSED(args);
+    auto world = args.getWorld();
+    auto name = world->sound.createSfxInstance(convertScriptIndexIntoSfx(sound));
+    world->sound.playSfx(name, coord, false, getScriptRange(sound));
```
danhedron: You call `convertScriptIndexIntoSfx` and `getScriptRange` here, where they both do a linear search to find the relevant SFX.

It would be better to have a single function like `SoundInstanceData* findScriptSoundSFXData(size_t sound)` so you can avoid doing the search twice, and more explicitly handle the invalid sound index case.

madebr:
@mole99 the sfx file in my data tree has/had a similar naming
https://github.com/madebr/openrw-irc/blob/6e5ce71e802973099868311b23cca936fb4854fe/2018-06-21.log#L100-L126

ShFil119:
What is your source of assets? I'll try to add checking for both cases. ;)

rwcore/loaders/LoaderSDT.hpp
```diff
@@ -63,7 +63,7 @@ class LoaderSDT {
 
     /// Load the structure of the archive
     /// Omit the extension in filename
```
danhedron: Comment is unneeded now

mole99:
Awesome, it works now even if it's called SFX.RAW :+1: 

Just one small change:
In `LoaderSDT::load(...)` the variable `path` isn't declared anymore, I think you want to use `sdtPath`.

ShFil119:
Fixed. :gear:

mole99:
LGTM.

mole99:
Just for completion: enabling av_register_all() successfully fixed the no-sound issue in Debian Stretch.

PR #594, opened 2018-08-15
Conan updates

madebr:
- Do not use `rwdep:xxx` targets
- Use recent and prebuilt  conan packages
- cmake + ci fixes

Issues:
- building rwviewer in debug mode fails: https://github.com/bincrafters/community/issues/403 (workaround  present in this pr)
- opus through ffmpeg contains a main function, which will be picked up if it is first passed to the linker: https://github.com/bincrafters/community/issues/404 (workaround present in this pr)


madebr:
@danhedron I think the problem will only be resolved once I create a minimal reproducer. Because the problem might be caused by rwviewer. Let's review again next week.

madebr:
@danhedron I fixed the failed rwviewer build in the latest commit.

madebr:
@ShFil119 I cannot find any reference to `branched` on [their docker page](https://hub.docker.com/_/fedora/).

PR #600, opened 2018-08-19
Cleanup Game Renderer

danhedron:
This is just an initial pass to take care of the OpenGL usage that was interfering with removing OpenGL from elsewhere, and make the code easier to look at.

- Remove unused code
- Remove bare glDraw calls
- Remove unused opengl objects
- Refactored some code into more readable methods

Follow up work:
- Remove raw OpenGL from MapRenderer

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -32,14 +32,6 @@ constexpr uint32_t kMissingTextureBytes[] = {
     0xFFFF00FF, 0xFF0000FF, 0xFFFF00FF, 0xFF0000FF,
 };
 
-struct WaterVertex {
-    static const AttributeList vertex_attributes() {
-        return {{ATRS_Position, 2, sizeof(WaterVertex), 0ul}};
-    }
-
-    float x, y;
-};
-
 /// @todo collapse all of these into "VertPNC" etc.
 struct ParticleVert {
     static const AttributeList vertex_attributes() {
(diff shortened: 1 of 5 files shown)


rwengine/src/render/GameRenderer.cpp
```diff
@@ -573,126 +520,35 @@ void GameRenderer::drawColour(const glm::vec4& colour, glm::vec4 extents) {
...
+    constexpr float cinematicExperienceSize = 0.15f;
+    renderer->useProgram(ssRectProg.get());
+    renderer->setUniform(ssRectProg.get(), "colour", glm::vec4{0.f, 0.f, 0.f, 1.f});
+    renderer->setUniform(ssRectProg.get(), "size", glm::vec2{1.f, cinematicExperienceSize});
+    renderer->setUniform(ssRectProg.get(), "offset", glm::vec2{0.f,-1.f * (1.f - cinematicExperienceSize)});
+    Renderer::DrawParameters wdp;
+    wdp.depthMode = DepthMode::OFF;
+    wdp.blendMode = BlendMode::BLEND_NONE;
+    wdp.count = ssRectGeom.getCount();
+    wdp.textures = {0};
```
ShFil119: nit: generally we do here a lot of dynamic allocations. Can we replace type of `textures` aka `std::vector` with templated `std::array`?

danhedron: That is possible, but it would fix the number of textures we could use at once.

It's a reasonable change to make, as we don't have that many textures to worry about.

ShFil119: Objects reusing is also an option, but I guess It'll require even more work.

danhedron: Yes. I prefer using the fixed size array to explicitly set the texture state for the draw.

PR #603, opened 2018-08-22
Prepare tests for Sound and SoundBuffer

ShFil119:
Also refactored SoundBuffer a bit.


tests/test_Sound.cpp
```diff
@@ -0,0 +1,105 @@
...
+//    BOOST_REQUIRE(sound.isPlaying() == true);
+
+//    sound.pause();
+//    BOOST_REQUIRE(sound.isPaused() == true);
+
+//    sound.stop();
+//    BOOST_REQUIRE(sound.isStopped() == true);
+//}
+
+BOOST_AUTO_TEST_CASE(checkSettingPosition) {
```
danhedron: Test names should state the expectation explicitly. You already have one expectation per test (good stuff), but the names could be clearer, for example this test could be "sound_sets_openal_source_position" so if it fails in future it's obvious what the test was for before anyone even looks at the code.

PR #605, opened 2018-08-26
rwengine: delete OpenAL buffers and sources before OpenAL shutdown

madebr:
OpenAL printing
```
AL lib: (WW) FreeContext: (0x614000052440) Deleting 22 Sources
AL lib: (WW) FreeDevice: (0x62d0001f4400) Deleting 22 Buffers
```
at program shutdown is actually a warning that not all sources and
buffers were deleted.

### DIFF
--- a/rwengine/src/audio/SoundBuffer.cpp
+++ b/rwengine/src/audio/SoundBuffer.cpp
@@ -15,6 +15,11 @@ SoundBuffer::SoundBuffer() {
     alCheck(alSourcei(source, AL_LOOPING, AL_FALSE));
 }
 
+SoundBuffer::~SoundBuffer() {
+    alCheck(alDeleteSources(1, &source));
+    alCheck(alDeleteBuffers(1, &buffer));
+}
+
 bool SoundBuffer::bufferData(SoundSource& soundSource) {
     alCheck(alBufferData(
         buffer,
--- a/rwengine/src/audio/SoundManager.cpp
+++ b/rwengine/src/audio/SoundManager.cpp
@@ -38,13 +38,7 @@ SoundManager::SoundManager(GameWorld* engine) : _engine(engine) {
 }
 
 SoundManager::~SoundManager() {
-    // De-initialize OpenAL
-    if (alContext) {
-        alcMakeContextCurrent(nullptr);
-        alcDestroyContext(alContext);
-    }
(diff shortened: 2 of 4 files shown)


rwengine/src/audio/SoundManager.cpp
```diff
@@ -38,6 +38,10 @@ SoundManager::SoundManager(GameWorld* engine) : _engine(engine) {
 }
 
 SoundManager::~SoundManager() {
+    sounds.clear();
```
ShFil119: Here and below is unneeded. (Map does on its own.)

rwengine/src/audio/SoundBuffer.cpp
```diff
@@ -15,6 +15,14 @@ SoundBuffer::SoundBuffer() {
     alCheck(alSourcei(source, AL_LOOPING, AL_FALSE));
 }
 
+SoundBuffer::~SoundBuffer() {
+    if (this->isPlaying()) {
```
ShFil119: Here and below `this`I think is redundant.

rwengine/src/audio/SoundManager.cpp
madebr: You need to delete the openal buffers inside the sound objects before openAL closes.

ShFil119:
Thanks for patch, forgotten when refactoring. (It's easy to hit openAL's limit 256 buffers now)

rwengine/src/audio/SoundManager.cpp
ShFil119: Or maybe even better - wrapping in function?

madebr: Removing these 3 clears, will still lead to the warning at program shutdown.

ShFil119: I mean removing `sfx.clean()` and moving rest into new function.

rwengine/src/audio/SoundBuffer.cpp
madebr: You're right, I can remove the test and stop.
I don't know whether this is defined behavior?

madebr:
@ShFil119  I've included your suggestions.

danhedron:
Needs rebase

PR #614, opened 2018-08-28
rwviewer: load gamedata before passing it to gameworld

madebr:
Load GameData before passing it to GameWorld.
GameWorld needs loaded GameData because SoundManager's constructor accesses it.


madebr:
Same fix as https://github.com/rwengine/openrw/pull/601 but for the rwviewer.

PR #621, opened 2018-08-31
Use microprofile for profiling

danhedron:
screenshot from 2018-08-31 20-39-29
screenshot from 2018-08-31 20-39-57

Replace the home-grown profiler (bad) with microprofile (good).

Microprofile has a few advantages:-

- Information is rendered out of process, so it doesn't interfere with itself.
- It has a better feature set (explicit colors, saving captures).
- It's probably faster than the profiler I wrote.
- It has support for GPU side timing (unused at the moment).

The OpenGL debugging is removed from the `RW_PROFILER` path, as it casuses CPU-side waits in places they wouldn't otherwise occur.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "external/microprofile"]
```
madebr: This will certainly be useful!
I would put external dependencies in a sub-directory of `external`.
`librw` will also become a dependency and can then be put alongside `microprofiler`.

I propose the following directory structure:
```
external
├── CMakeLists.txt
├── microprofiler
│    ├── microprofiler/  # git submodule
│    └── CMakeLists.txt
├── librw
│    ├── librw/          # git submodule
│    └── CMakeLists.txt
└── librwgta
     ├── librwgta/       # git submodule
     └── CMakeLists.txt
       
```
I did something similar here: https://github.com/madebr/openrw/tree/librw/external

external/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+
+if (ENABLE_PROFILING)
+
+add_library(microprofile STATIC
+  microprofile/microprofile.cpp
```
madebr: consider adding:
```
microprofile/microprofile.h
microprofile/microprofile_html.h
```
So all IDE's will display these files as well.

external/microprofile/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+add_library(microprofile STATIC
+    microprofile/microprofile.h
+    microprofile/microprofile_html.h
+    microprofile/microprofile.cpp
+)
+
+target_include_directories(microprofile
```
madebr: I just tried to compile this branch and got this error:
```
/usr/bin/ld: ../external/microprofile/libmicroprofile.a(microprofile.cpp.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//usr/lib64/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
Adding
```
find_package(Threads)
target_link_libraries(microprofile
    PUBLIC
        Threads::Threads
)
```
fixed this.

madebr: This is still required.
This link error does not show up on ci because `ENABLE_PROFILERS=address` on ci.
If I enable this profiler, the link error goes away.

The link error happens for rwviewer, since that one does not link to SDL2 (which already links to pthread).

PR #629, opened 2018-09-02
Convert AnimationBone and Animator to unique ptr

ShFil119:


ShFil119:
Worth noted is that - after this pull it will be easy to write code which clears resources after cutscenes.

rwengine/src/ai/CharacterController.cpp
```diff
@@ -669,7 +669,7 @@ bool Activities::UseItem::update(CharacterObject *character,
     auto world = character->engine;
     const auto &weapon = world->data->weaponData.at(itemslot);
     auto &state = character->getCurrentState().weapons[itemslot];
-    auto animator = character->animator;
+    auto& animator = character->animator;
```
Lihis: Inconsistent placement of `&`.

rwengine/src/objects/CharacterObject.cpp
```diff
@@ -316,12 +316,12 @@ void CharacterObject::changeCharacterModel(const std::string& name) {
     auto newmodel = engine->data->loadClump(modelName + ".dff");
 
     if (animator) {
```
danhedron: The if can be removed and just call `animator.reset()` instead.

ShFil119:
Updated. :)

PR #631, opened 2018-09-08
Game objects as unique ptrs

ShFil119:
Tweaked a bit `test_GameWorld.cpp` actual version would delete world twice.

### DIFF
--- a/rwengine/src/engine/GameWorld.cpp
+++ b/rwengine/src/engine/GameWorld.cpp
@@ -76,9 +76,13 @@ GameWorld::GameWorld(Logger* log, GameData* dat)
 }
 
 GameWorld::~GameWorld() {
-    for (auto& p : allObjects) {
-        delete p;
-    }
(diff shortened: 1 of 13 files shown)


rwengine/src/engine/GameWorld.cpp
```diff
@@ -76,9 +76,13 @@ GameWorld::GameWorld(Logger* log, GameData* dat)
 }
 
 GameWorld::~GameWorld() {
-    for (auto& p : allObjects) {
-        delete p;
-    }
+    // Bullet requires to remove each object before all physic world
```
danhedron: This should already happen, because the object pool members in GameWorld are before dynamicsWorld.

ShFil119: I remember some some asan issue. Will retest it. ;)

ShFil119: ahh, there's problem with `TrafficDirector` in test `test_available_nodes`.

Issue #598, opened 2018-08-18
Memory ownership in class AIGraph should be refactored


ShFil119:
Code is quite "nasty" to read and use. It's easy to make memory leak, see tests.

Using RAII should make code easier to understand how to use.

ShFil119:
Fixed by #632

PR #632, opened 2018-09-08
Convert nodes to unique ptrs

ShFil119:
Additionally avoided copying vector in `resetToAINode` and added `const` in range loops.
Refs: #598


ShFil119:
Resolved conflicts. :gear: :gear: :gear:

Issue #428, opened 2018-04-30
OpenRW uses quite a bit of RAM


darkf:
Just opening it up to the main menu uses 876 MiB of RAM, according to Windows Task Manager. If OpenRW is loading everything up front, it should probably stream it instead.

ShFil119:
It's weird, in debug mode on arch linux openRW has mem usage ~230MB.
![mem](https://user-images.githubusercontent.com/7227492/39470631-05c039ee-4d3f-11e8-8e8a-06360c483bcc.png)

(Cutscenes are a bit problematic, after ending they don't release animations and music)

ShFil119:
@darkf, can you check master now?

PR #636, opened 2018-09-27
Cutscenes cleanup

ShFil119:
First commit replaces raw ptr with unique ptr.
Second one implements flushing resources (animations and sound) when cutscene ends.

Should hep with #428. Requesting reviews, especially from @darkf. ;)

### DIFF
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -527,13 +527,8 @@ void GameData::loadIFP(const std::string& name, bool cutsceneAnimation) {
 
     if (f.data) {
         if (LoaderIFP loader{}; loader.loadFromMemory(f.data.get())) {
-            if (cutsceneAnimation) {
-                animationsCutscene.insert(loader.animations.begin(),
-                                          loader.animations.end());
-            } else {
-                animations.insert(loader.animations.begin(),
-                                  loader.animations.end());
-            }
+            auto& dest = cutsceneAnimation ? animationsCutscene : animations;
+            dest.insert(loader.animations.begin(), loader.animations.end());
         }
     }
 }
--- a/rwengine/src/engine/GameData.cpp
+++ b/rwengine/src/engine/GameData.cpp
@@ -522,14 +522,18 @@ bool GameData::loadModel(ModelID model) {
     return true;
 }
 
-void GameData::loadIFP(const std::string& name) {
+void GameData::loadIFP(const std::string& name, bool cutsceneAnimation) {
     auto f = index.openFile(name);
 
     if (f.data) {
-        LoaderIFP loader;
-        if (loader.loadFromMemory(f.data.get())) {
-            animations.insert(loader.animations.begin(),
-                              loader.animations.end());
+        if (LoaderIFP loader{}; loader.loadFromMemory(f.data.get())) {
+            if (cutsceneAnimation) {
+                animationsCutscene.insert(loader.animations.begin(),
+                                          loader.animations.end());
+            } else {
+                animations.insert(loader.animations.begin(),
+                                  loader.animations.end());
+            }
         }
     }
 }
--- a/rwengine/src/engine/GameData.hpp
+++ b/rwengine/src/engine/GameData.hpp
@@ -160,7 +160,7 @@ public:
     /**
      * Loads an IFP file containing animations
      */
-    void loadIFP(const std::string& name);
+    void loadIFP(const std::string& name, bool cutsceneAnimation = false);
 
     /**
      * Loads data from an object definition dat.
@@ -289,9 +289,14 @@ public:
     std::vector<TextureAtlas*> atlases;
 
     /**
-     * Loaded Animations
+     * Loaded Animations (doesn't contain animations of cutscene)
      */
     AnimationSet animations;
+
+    /**
+     * Loaded Animations (only) for cutscene
+     */
+    AnimationSet animationsCutscene;
 
     /**
      * Pedestrian Animation Groups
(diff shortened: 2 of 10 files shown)


ShFil119:
Resolved conflict.

rwengine/src/engine/GameState.hpp
```diff
@@ -316,7 +316,9 @@ class GameState {
 
     bool skipCutscene = false;
     bool isIntroPlaying = false;
-    CutsceneData* currentCutscene = nullptr;
+
+    std::unique_ptr<CutsceneData> currentCutscene;
```
ShFil119: nit: with cpp 17 we can avoid dynamic allocation when using optional.

madebr: `optional` would indeed simplify a lot of things.
Boost also has an optional.

ShFil119:
As suggested on irc - instead on keeping bool in Animation, I 've added second map just for cutscene animations.

rwengine/src/engine/GameData.cpp
```diff
@@ -522,14 +522,19 @@ bool GameData::loadModel(ModelID model) {
...
-void GameData::loadIFP(const std::string& name) {
+void GameData::loadIFP(const std::string& name, bool cutsceneAnimation) {
     auto f = index.openFile(name);
 
     if (f.data) {
         LoaderIFP loader;
         if (loader.loadFromMemory(f.data.get())) {
-            animations.insert(loader.animations.begin(),
-                              loader.animations.end());
+            if (!cutsceneAnimation) {
```
madebr: I would swap the if/elses. Makes the code clearer.

```suggestion
            if (cutsceneAnimation) {
                animationsCutscene.insert(loader.animations.begin(),
                                          loader.animations.end())
            } else {
                            animations.insert(loader.animations.begin(),
                                  loader.animations.end());
            }
```

rwengine/src/engine/GameData.hpp
```diff
@@ -293,6 +293,11 @@ class GameData {
      */
```
madebr: Add some documentation about why there are 2 sets? And that they are disjoint? (are they?)

rwengine/src/script/modules/GTA3ModuleImpl.inl
```diff
@@ -7890,7 +7890,7 @@ void opcode_02e6(const ScriptArguments& args, const ScriptObject object, const S
     auto cutscene = args.getObject<CutsceneObject>(0);
     std::string animName = arg2;
     std::transform(animName.begin(), animName.end(), animName.begin(), ::tolower);
-    auto anim = args.getWorld()->data->animations[animName];
+    auto anim = args.getWorld()->data->animationsCutscene[animName];
```
madebr: An `AnimationSet` is a `std::map`. Using the `operator[]` will create a new entry in the map.
So use the `find` member function.

ShFil119:
I'll try to fix findings and use `std::optional` instead of `std::unique_ptr`.

ShFil119:
Up. ^^

rwengine/src/engine/GameData.cpp
```diff
@@ -522,14 +522,19 @@ bool GameData::loadModel(ModelID model) {
     return true;
 }
 
-void GameData::loadIFP(const std::string& name) {
+void GameData::loadIFP(const std::string& name, bool cutsceneAnimation) {
     auto f = index.openFile(name);
 
     if (f.data) {
         LoaderIFP loader;
         if (loader.loadFromMemory(f.data.get())) {
```
madebr: Let `LoaderIFP::loadFromMemory` accept `f.data` or `f`?

rwengine/src/engine/GameData.hpp
madebr: nit: second animation should be plural as well

rwengine/src/engine/GameWorld.cpp
```diff
@@ -750,15 +751,14 @@ void GameWorld::PhysicsTickCallback(btDynamicsWorld* physWorld,
 void GameWorld::loadCutscene(const std::string& name) {
     auto datfile = data->index.openFile(name + ".dat");
 
-    CutsceneData* cutscene = new CutsceneData;
+    auto cutscene = CutsceneData();
```
madebr: Curly braces?

ShFil119: It shouldn't make any difference. I'll try to avoid moving. ;)

rwengine/src/engine/GameData.cpp
ShFil119: In my opinion it's too much for this pull. (it changes design)

ShFil119:
Small up. :pager:

rwengine/src/engine/GameWorld.cpp
```diff
@@ -788,20 +784,32 @@ void GameWorld::startCutscene() {
 }
 
 void GameWorld::clearCutscene() {
+    eraseCutsceneObjects();
+    eraseCutsceneSound();
+    eraseCutsceneAnimations();
+
+    state->currentCutscene = std::nullopt;
+    state->isCinematic = false;
```
ShFil119: maybe we should prepare "universal" function to free camera?

madebr: If it would mean, less spaghetti code, fine by me :smile:

rwgame/states/IngameState.cpp
```diff
@@ -380,7 +380,7 @@ const ViewCamera& IngameState::getCamera(float alpha) {
     auto world = getWorld();
 
     if (state->currentCutscene && state->cutsceneStartTime >= 0.f) {
-        auto cutscene = state->currentCutscene;
+        auto& cutscene = state->currentCutscene;
```
madebr: since you're using a reference, make it `const`?

PR #640, opened 2018-10-01
New Screenshots and Improved direction in readme

danhedron:


ShFil119:
Maybe screenshots should be taken from window with higher resolution? I think about aliasing.

vflyson:
The first 2-3 screenshots seem to have an unusual (for the vanilla game) camera angle, I think they're perfect aside from that. In my opinion, it's important to make the good first impression for people who see this project for the first time, ergo making snapshots which would remind them of the original experience playing that game would describe better what the goal of OpenRW is.

danhedron:
The UI is not scalled at the moment, so a larger window with result in weird UI. I can fudge it with super-sampling to reduce the aliasing or add a switch so UI rendering scales with the window.

vflyson:
If it's not too much work I think it may be worth implementing that before taking snapshots for the updated README file; or taking the same snapshots in a 640x480 window and only scaling them down like 50% before uploading may also work, but you'd know better.

ShFil119:
nit: maybe we should post also link to cdash?

madebr:
https://github.com/danhedron/openrw/tree/update-readme
LGTM

PR #642, opened 2018-10-09
Improved function for destroying effects

ShFil119:
Old version (needlessly) iterates on all objects of vector.


ShFil119:
It doesn't change behavior of engine and dracc has taken a look at it. So I'm merging. ^^

Issue #425, opened 2018-04-29
Switching to C++17 as the default standard?


ShFil119:
It has some advantages for us:
- safe std::optional instead of union,
- easier setting constexpr for more data structures,
- some staff from boost is included in cpp17 (filesystem IIRC)

JayFoxRox:
We dicussed this in the past if I remember correctly.

My reasoning is typically: Upgrading dependencies (including toolchains / compilers) is fine and a good idea if only free software changes are necesary.

However, one issue I see is the portability. C++17 is still rather "new", even today. Some distributions might not ship it in their LTS and especially homebrew SDKs for other consoles won't properly work with it.

Another issue I have with more modern languages for projects like this, is that we diverge more and more from the original game logic as we overcomplicate things with fancy templates (/ macros) etc.

---

On the other hand, the RE-route has not been taken for OpenRW and for the majority of the code it will never happen anyway. So the added logic complexity is fine with me.
Furthermore, I doubt that many users are using OpenRW in it's current form anyway, even on PC.
By the time OpenRW is shippable or worth using, we'll probably have widespread C++17 adoption.
If anyone is really desperate to run OpenRW on a legacy non-C++17 platform, they can probably port C++ too.

So going with C++17 and doing extra work for porting, is probably easier than crippling OpenRW with weird kludges to do the same (even if there was probably an even more elegant design choice in the original games code to reach the same goal).
Most developers will also be running a modern platform with C++17 anyway.

dracc:
Fwiw I'm all for going C++17.
Lesser third party dependence is typically a good thing.

darkf:
It's somewhat well-supported AFAICT, at least the compiler aspect of it -- projects like Citra/Yuzu already use it, and Dolphin has optional support while C++17 compilers mature.

It will indeed drop `boost::filesystem`.

madebr:
We're no dinosaurs (maybe).

I would just make the switch if even a big library as qt is going for c++17 for qt6.
https://www.phoronix.com/scan.php?page=news_item&px=More-Qt6-2020-Details

JayFoxRox:
@madebr that news article says they'll be using C++17 in November 2020.

---

I'm still slightly worried about people on their LTS, but I guess they can add custom repos, like @Calinou said.
But so far, nobody is against C++17, so it will probably happen very soon for us.

In fact, I don't see anything which would prevent us from deciding this *right now* to avoid hearing the same arguments again. @danhedron ?

[...]

PR #644, opened 2018-10-16
C++17

ShFil119:
As problems:
- https://github.com/rwengine/openrw/commit/491e5303d0e23433f826ee3d98e1da294b191e21
- https://github.com/bincrafters/community/issues/486
were solved, we can switch to c++17. ;)


madebr:
See #425.
I'm in favor for c++17.

ShFil119:
Good point. I think for now we can stay with boost. 

But using gcc8+ shouldn't be big problem if we want to provide packages. Also it's available in official repo of 18.04 aka newest LTS.

madebr:
Let's not tie our hands on our backs needlessly and go c++17 all-in.
if `std::filesystem` is in the standard, it's there to stay. (big chance at least)
So use it.

PR #646, opened 2018-10-17
Cleanup random generators

ShFil119:


madebr:
As an aside, wouldn't it be interesting if the game would be deterministic?
https://youtu.be/kSKeWH4TY9Y?t=2430

ShFil119:
Yeah, we could/should use strategy pattern for controllers and prepare deterministic versions.

rwengine/src/engine/GameWorld.hpp
```diff
@@ -401,6 +396,26 @@ class GameWorld {
 
     PlayerController* getPlayer();
 
+    template <
+        typename T1, typename T2 = T1,
+        typename std::enable_if<std::is_integral<T1>::value>::type* = nullptr,
+        typename std::enable_if<std::is_integral<T2>::value>::type* = nullptr>
+    T1 getRandomNumber(T1 min, T2 max) {
```
husho: static?

rwengine/src/engine/GameWorld.hpp
```diff
@@ -401,6 +396,26 @@ class GameWorld {
...
+        return dist(randomNumberGen);
+    }
+
+    template <
+        typename T1, typename T2 = T1,
+        typename std::enable_if<std::is_floating_point<T1>::value>::type* =
+            nullptr,
+        typename std::enable_if<std::is_floating_point<T2>::value>::type* =
+            nullptr>
+    T1 getRandomNumber(T1 min, T2 max) {
```
husho: static?

rwengine/src/engine/GameWorld.hpp
```diff
@@ -418,6 +433,11 @@ class GameWorld {
      * Private data
      */
     std::unique_ptr<btOverlappingPairCallback> _overlappingPairCallback;
+
+    /**
+     * Randomness Engine
+     */
+    std::mt19937 randomNumberGen{std::random_device()()};
```
ShFil119: After some thoughts I think it's better to  use `default_random_engine`.

ShFil119:
@madebr, if we avoid seeding, won't happen that?

rwengine/src/engine/GameWorld.hpp
madebr: `static` is impossible due to the random source (`randomNumberGen`).
I would rename the function to `getRandomUniform`. But that's a detail.

rwengine/src/engine/GameWorld.hpp
```diff
@@ -401,7 +401,8 @@ class GameWorld {
         typename std::enable_if<std::is_integral<T1>::value>::type* = nullptr,
         typename std::enable_if<std::is_integral<T2>::value>::type* = nullptr>
     T1 getRandomNumber(T1 min, T2 max) {
-        std::uniform_int_distribution<T1> dist(min, max);
```
madebr: I think making the `std::uniform_int_distribution` static makes the code unclear.
It is also a premature optimization.
Do we know whether the construction cost of a `uniform_int_distribution` is high?
Because in the next line, it is copy constructed anyways.

madebr:
> @madebr, if we avoid seeding, that won't happen?

Making the engine deterministic is a big problem, maybe out-of-scope for this project.
It also needs a deterministic physics engine etc..
Even though, moving all the random operations to a central place is a step in the right direction.

rwengine/src/engine/GameWorld.hpp
ShFil119: reverted

ShFil119:
Rebased.

Issue #643, opened 2018-10-14
Build fails (use of deleted function)


vflyson:
OS: `Linux localhost 4.9.0-1-amd64 #1 SMP Debian 4.9.6-3 (2017-01-28) x86_64 GNU/Linux`
Compiler: `gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1)`

The log:
```
[ 65%] Building CXX object rwengine/CMakeFiles/rwengine.dir/src/render/GameRenderer.cpp.o
In file included from /usr/include/x86_64-linux-gnu/c++/6/bits/c++allocator.h:33:0,
                 from /usr/include/c++/6/bits/allocator.h:46,
                 from /usr/include/c++/6/memory:63,
                 from /home/vflyson/Projects/openrw/rwengine/src/objects/VehicleObject.hpp:6,
                 from /home/vflyson/Projects/openrw/rwengine/src/objects/VehicleObject.cpp:1:
/usr/include/c++/6/ext/new_allocator.h: In instantiation of ‘void __gnu_cxx::new_allocator<_Tp>::construct(_Up*, _Args&& ...) [with _Up = std::pair<const std::__cxx11::basic_string<char>, VehicleObject::Part>; _Args = {const std::pair<const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, VehicleObject::Part>&}; _Tp = std::pair<const std::__cxx11::basic_string<char>, VehicleObject::Part>]’:
/usr/include/c++/6/bits/alloc_traits.h:475:4:   required from ‘static void std::allocator_traits<std::allocator<_Tp1> >::construct(std::allocator_traits<std::allocator<_Tp1> >::allocator_type&, _Up*, _Args&& ...) [with _Up = std::pair<const std::__cxx11::basic_string<char>, VehicleObject::Part>; _Args = {const std::pair<const std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, VehicleObject::Part>&}; _Tp = std::pair<const std::__cxx11::basic_string<char>, VehicleObject::Part>; std::allocator_traits<std::allocator<_Tp1> >::allocator_type = std::allocator<std::pair<const std::__cxx11::basic_string<char>, VehicleObject::Part> >]’
[...]

PR #648, opened 2018-10-19
Replace insert with emplace

ShFil119:
See:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87658

Should help with #643.


ShFil119:
Code doesn't change behavior, so I think it's enough. I'll merge it.

PR #649, opened 2018-10-22
Pause all other sounds during a cutscene

### LINKED ISSUE #583: Before cutscene all other sounds should be stopped/paused
Sound of sfx, audio station can mix with that of cutscene.

qkolj:
Fixes #583 
I tested only the cutscene on the first mission (Luigi's Girls), but it should work for any other.


ShFil119:
Thanks for contribution. ^^ I'll review and test it tomorrow.

ShFil119:
Thanks. Tested, works fine. 

Todo: later verify if original also stops sfx generated by 018d and background. (Don't have windows to check. :( )

PR #652, opened 2018-10-31
Start using structured bindings

ShFil119:
It should increase readability a bit. ;)


rwengine/src/engine/Animator.cpp
```diff
@@ -24,21 +24,16 @@ void Animator::tick(float dt) {
         glm::quat rotation{1.0f,0.0f,0.0f,0.0f};
     };
 
-#if 0
```
ShFil119: I'm removing these snippets, because it doesn't  even compile and makes code more complex.

PR #656, opened 2018-11-08
Reduce nr of dynamic allocations

ShFil119:
Do you prefer using const ref or raw ptr?

### DIFF
--- a/rwengine/src/ai/CharacterController.cpp
+++ b/rwengine/src/ai/CharacterController.cpp
@@ -674,7 +674,7 @@ bool Activities::UseItem::update(CharacterObject *character,
     };
 
     auto world = character->engine;
-    const auto &weapon = world->data->weaponData.at(itemslot);
+    auto weapon = &world->data->weaponData.at(itemslot);
     auto &state = character->getCurrentState().weapons[itemslot];
     auto &animator = character->animator;
     auto shootcycle = find_cycle(weapon->animation1);
@@ -712,7 +712,7 @@ bool Activities::UseItem::update(CharacterObject *character,
 
             if (currenttime >= firetime && !fired) {
                 state.bulletsClip--;
-                Weapon::fireHitscan(weapon.get(), character);
+                Weapon::fireHitscan(weapon, character);
                 fired = true;
             }
             if (currenttime > loopend) {
@@ -739,7 +739,7 @@ bool Activities::UseItem::update(CharacterObject *character,
 
             if (currID >= firetime && !fired) {
                 state.bulletsClip--;
-                Weapon::fireProjectile(weapon.get(), character, power);
+                Weapon::fireProjectile(weapon, character, power);
                 fired = true;
             }
             if (animator->isCompleted(AnimIndexAction)) {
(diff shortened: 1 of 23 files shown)


rwengine/src/ai/CharacterController.cpp
```diff
@@ -677,8 +677,8 @@ bool Activities::UseItem::update(CharacterObject *character,
     const auto &weapon = world->data->weaponData.at(itemslot);
```
madebr: What if there is no weapon at itemslot?

ShFil119: According to standard, `at` should throw exception. It's connected with question first post. Ref cannot be empty - so if we use this option, then modified and invalid assets could crash game. (ofc pointers also require checks)

Question if is it a problem?

rwengine/src/objects/ProjectileObject.hpp
```diff
@@ -26,7 +26,7 @@ class ProjectileObject final : public GameObject {
 
     struct ProjectileInfo {
         ProjectileInfo(ProjectileType p_type, glm::vec3 p_direction,
```
madebr: Should we keep the `p_` prefixes? It is expected that the constructor takes projectile type, direction, velocity, time and weapon data.

madebr:
If I understand correctly, you are sacrificing run time performance (`O(log(N))` -> `O(N)`) to dynamic memory usage (`N x string length`).
Is it worth it? Do you know how large `N` is?

rwengine/src/ai/CharacterController.cpp
madebr: I guess not now.
Later on, malicious scm or a scm fuzzer can crash the game.

ShFil119:
I can return to `std::map`. (or use unordered_map)

If I start debug mode, then there are 168 dynamic allocations. Map will cause more jumping on memory and duplicating memory usage for key(it's already in second slot).

It's hard to frankly say what's is better `O(N)` with small constant(thx to contiguous memory) or `O(log(N))` with bigger constant.

madebr:
The `std::string` in the `std::vector` will not necessarily be continuous. `std::string`  will allocate sufficient memory on the heap for longer strings. (smaller strings will benefit from the small string optimization)
If you want to keep the string in the `std::vector` (with sufficient capacity) and no other heap allocations, you should use `char string[MAXSIZE]` or `std::array<char, MAXSIZE>`. (idk how to store strings the C++ way)

ShFil119:
Yeah, `std::string` is just ptr to memory. I was thinking overall about all other Plain Old Data. (but still there is one less jump on memory) ;)

Using char for runtime strings (I think) is downgrade.

ShFil119:
Refactored, now uses std::unordered_map. Also avoided duplication of id string.

rwengine/src/objects/VehicleInfo.hpp
```diff
@@ -103,9 +102,15 @@ struct VehicleInfo {
         bool empty() const {
             return front.empty() && back.empty();
         }
-    } seats;
-};
+    };
```
madebr: nit: You can do
```
struct Seats {
   ...
} seats;
```

PR #658, opened 2018-11-17
Quick fix for choppy performance at startup

danhedron:
Send all bullet rigid bodies belonging to buildings and objects to sleep when loading.


rwengine/src/objects/InstanceObject.cpp
```diff
@@ -26,33 +26,40 @@ InstanceObject::InstanceObject(GameWorld* engine, const glm::vec3& pos,
...
 
-        /// @todo store path information properly
-        if (modelinfo->type() == ModelDataType::SimpleInfo) {
-            auto simpledata = static_cast<SimpleModelInfo*>(modelinfo);
-            for (auto& path : simpledata->paths) {
-                engine->aigraph.createPathNodes(position, rot, path);
-            }
-        }
+    if (body)
+    {
```
ShFil119: nit: formatting

Issue #622, opened 2018-08-31
Bullet takes ~60% of cpu time


ShFil119:
To do:
- determine who is the culprit (collision?, objects callbacks?)
- find an approach to fix it.

;)

vflyson:
I'm unable to properly run the game due to this issue on the machine which I successfully used to run openrw less than 6 months ago.

madebr:
@vflyson Apple, Linux or Windows?
@ShFil119 Would a bisect be useful? (I think we need a benchmark test for that that returns metrics)

dracc:
On debug or release build? (Would that matter btw? :thinking_face:

madebr:
@dracc It does on Windows.

ShFil119:
On linux also.

Lihis:
On my machine, on linux, even the debug build is still playable tho.

ShFil119:
Playable in debug mode - yes. But if you want to develop, aka use asan, it becomes unusable.

vflyson:
@madebr `Linux hostname 4.9.0-1-amd64 #1 SMP Debian 4.9.6-3 (2017-01-28) x86_64 GNU/Linux`

ShFil119:
@vflyson what hardware?

madebr:
I think I can reproduce a similar slowdown on my system.
When I start a new game, the fps floats around 20 to 30.
If I jump in the water, respawn on the other side of the river and run around a little, a lot of vehicles are created (looking at the console).
At the same time, the fps drops to 10 and lower and does not recover.

FP4:
there doesnt seem to be setForceUpdateAllAabbs set to false on the btDiscreteDynamicsWorld which causes aabb and also motionStates to be calulated for all static objects, every simulation cycle, which is just ridiculous

dracc:
Post-#660, I am back at a stable 75fps in debug-mode. :+1:

danhedron:
Closing for now.

PR #660, opened 2018-11-18
setForcedUpdateAllAabbs to false; Reduces stepSimulation time by 35%.

danhedron:
This causes a drop from 6.5ms to 4.2ms on my machine.

Thanks to @FP4 for pointing this out in #622.

To Check:
- [ ] Static objects update their AABB when moved by the game.


ShFil119:
I guess task "Static objects update their AABB when moved by the game." is not covered in this pull. (?)
@danhedron, can you open issue about this? (to avoid forgetting)

PR #665, opened 2018-11-23
[improvement] Contiguous memory for `bones` container

ShFil119:
Should improve a bit performance.

### DIFF
--- a/rwengine/src/loaders/LoaderIFP.cpp
+++ b/rwengine/src/loaders/LoaderIFP.cpp
@@ -63,7 +63,7 @@ bool LoaderIFP::loadFromMemory(char* data) {
     ANPK* fileRoot = read<ANPK>(data, dataI);
     std::string listname = readString(data, dataI);
 
-    for (int a = 0; a < fileRoot->info.entries; ++a) {
+    for (auto a = 0u; a < fileRoot->info.entries; ++a) {
         // something about a name?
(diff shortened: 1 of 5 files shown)


rwengine/src/loaders/LoaderIFP.cpp
```diff
@@ -76,14 +76,16 @@ bool LoaderIFP::loadFromMemory(char* data) {
         DGAN* animroot = read<DGAN>(data, dataI);
         std::string infoname = readString(data, dataI);
 
+        animation->bones.reserve(static_cast<unsigned>(animroot->info.entries));
```
madebr: Make `INFO::entries` `std::uint32_t`?

rwengine/src/loaders/LoaderIFP.cpp
```diff
@@ -76,14 +76,16 @@ bool LoaderIFP::loadFromMemory(char* data) {
...
+
         for (int c = 0; c < animroot->info.entries; ++c) {
             size_t start = data_offs;
             CPAN* cpan = read<CPAN>(data, dataI);
             ANIM* frames = read<ANIM>(data, dataI);
 
             AnimationBone boneData{};
             boneData.name = frames->name;
-            boneData.frames.reserve(frames->frames);
+            boneData.frames.reserve(static_cast<unsigned>(frames->frames));
```
madebr: Likewise. `std::uint32_t`

PR #666, opened 2018-11-27
Add coverage reports

madebr:
This pr adds some targets to the build system to collect coverage and submits the results to codecov.io.
Travis-ci will build the application, run the tests and submit the results.

- The logic to collect the coverage and generate a report, is completely embedded in the cmake script. This allows developers to generate offline reports and switch coverage servers. The bash script provided by https://codecov.io is used only to submit the result to codecov's servers.
- The following targets are interesting for developers:
   - `lcov`: generate a total coverage report in `lcov/html/all_targets/`
   - `TARGET-lcov`: generate a coverage report for target `TARGET` in `lcov/html/TARGET`
   - `coverage_upload`: collect coverage and upload to codecov.io. When running this command on your pc, the environment variable `CODECOV_TOKEN` must be set. See [the docs](https://docs.codecov.io/docs/about-the-codecov-bash-uploader)
   - `lcov-geninfo-clean`: Reset the coverage counters.
- For now, I have enabled coverage collection on Ubuntu gcc and Fedora llvm. I had to revert to gcc-7 on ubuntu because there was an incompatibility between lcov and gcc-8.

Known issues:
- Collecting coverage in parallel (`make coverage_upload -j9`, `ninja coverage_upload`) does not work. https://github.com/linux-test-project/lcov/issues/37
So always run all coverage related targets with an additional `-j1` argument.
This also happens in the ci test scripts.
- Ubuntu has to revert to gcc-7 (instead of gcc-8). I believe this will be fixed soon as the coverage collection works fine on my personal Fedora gcc8 machine.
- Because we currently build the tests twice, the coverage is collected twice as well. This will cause the codecov bot to initially report a wrong (higher) coverage.


rwcore/CMakeLists.txt
```diff
@@ -66,7 +66,10 @@ target_link_libraries(rwcore
         OpenGL::GL
     )
 
-openrw_target_apply_options(TARGET rwcore)
+openrw_target_apply_options(TARGET rwcore
+    COVERAGE
+    COVERAGE_EXCEPT gl/gl_core_3_3.c gl/gl_core_3_3.h
```
ShFil119: Maybe also tests?

madebr: That file is gigantic and full of functions we don't use.

ShFil119: I thought about directory `tests`. Now it's treated as normal code and wrongly increase (overall) coverage.

madebr: I'm sorry. I misread your comment.
There is something to be told about adding/removing unit tests.

I am for including the unit tests because:
- currently, coverage is low in the unit tests because we lack game data on ci
- this is related to the `TESTS_NODATA` compile flag (that should disappear)
- coverage is _just a number_

cmake/ctest/script_ci.ctest
```diff
@@ -69,18 +72,8 @@ set(APPEND_RESULTS FALSE)
 set(TESTS_NODATA FALSE)
 set(RUN_TESTS FALSE)
 
-include("${CTEST_SCRIPT_DIRECTORY}/build.ctest")
+set(CODECOV_FLAGS "data")
```
danhedron: Do we want to run codecov on the data test compilation? as we never run it.

madebr: It does not matter much. Apart from a lower coverage percentage and time taken during ci.
It it a more accurate number.

madebr: It's an incentive for us/people to add mock data. The coverage will rise.

PR #667, opened 2018-12-01
Convert Menu instances to std::optional

### LINKED ISSUE #653: Instances of std::shared_ptr<Menu> can be replaced with std::optional<Menu>
As debug state is clear owner of these, there is no need for `std::shared_ptr`.

Furthermore (with available now c++17) it's possible to avoid dynamic memory allocation and reduce overhead.

ShFil119:
It fixes #653.

### DIFF
--- a/rwgame/MenuSystem.hpp
+++ b/rwgame/MenuSystem.hpp
@@ -1,8 +1,10 @@
 #ifndef _GAME_MENUSYSTEM_HPP_
 #define _GAME_MENUSYSTEM_HPP_
+
 #include <algorithm>
 #include <functional>
 #include <memory>
+#include <optional>
 #include <string>
 
 #include <glm/glm.hpp>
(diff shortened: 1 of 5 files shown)


ShFil119:
Or do you prefer `std::unique_ptr`?

rwgame/MenuSystem.hpp
```diff
@@ -78,10 +78,9 @@ class Menu {
      * @brief creates a menu from the given menu items
      * @return a shared pointer to the menu with the items
```
ShFil119: to fix

rwgame/states/DebugState.cpp
```diff
@@ -33,7 +33,7 @@ static void jumpCharacter(RWGame* game, CharacterObject* player,
     }
 }
 
-std::shared_ptr<Menu> DebugState::createDebugMenu() {
+std::optional<Menu> DebugState::createDebugMenu() {
```
madebr: Can't these function (and anywhere else) return a plain `Menu` (without the `std::optional`-wrapping)?

ShFil119: Can, was used that version earlier. I've changed because my ide says it can prevent copy elision on some compilers. (qt creator) Revert?

madebr: Thanks for the clarification. You don't have to revert. Ideally, the compiler should be smart enough. Would lto optimize this?

ShFil119: I don't have enough knowledge about lto to answer. :(

PR #668, opened 2018-12-01
Convert renderer to unique_ptr

ShFil119:


madebr:
I believe the `std::shared_ptr` abstraction was made as to allow different renderers.
Maybe use a `std::variant`?

danhedron:
The polymorphism here was to allow different render implementations, despite the OpenGL that leaked into other places.

rwgame/states/PauseState.cpp
```diff
@@ -33,7 +33,7 @@ void PauseState::tick(float dt) {
 void PauseState::draw(GameRenderer* r) {
```
madebr: Can this become a reference?

rwgame/states/LoadingState.cpp
```diff
@@ -32,7 +32,7 @@ void LoadingState::draw(GameRenderer* r) {
     // Display some manner of loading screen.
     TextRenderer::TextInfo ti;
     ti.text = kLoadingString;
-    auto size = r->getRenderer()->getViewport();
+    auto size = r->getRenderer().getViewport();
```
madebr: Can r become a reference as well?

rwgame/HUDDrawer.cpp
```diff
@@ -13,7 +13,7 @@
 void HUDDrawer::drawScriptTimer(GameWorld* world, GameRenderer* render) {
```
madebr: `render` reference as well?

madebr:
grep gave me this:
```
$ grep -rn "GameRenderer *\\*"
rwviewer/ViewerWindow.hpp:48:    void gameLoaded(GameWorld*, GameRenderer*);
rwviewer/ViewerWidget.hpp:64:    void gameLoaded(GameWorld* world, GameRenderer* renderer);
rwviewer/ViewerWidget.hpp:83:    GameRenderer* _renderer = nullptr;
rwviewer/views/ViewerInterface.hpp:26:    void gameLoaded(GameWorld*, GameRenderer*);
rwviewer/views/ViewerInterface.hpp:29:    void showData(GameWorld* world, GameRenderer* renderer) {
rwviewer/ViewerWidget.cpp:332:void ViewerWidget::gameLoaded(GameWorld *world, GameRenderer *renderer) {
```

ShFil119:
@madebr, I won't touch viewer in this pull, because renderer is initialized in function `gameLoaded` and even if there's possibility to initialize that variable in ctor, it would require "some" more work.

PR #672, opened 2018-12-10
Cleanup headers

ShFil119:
Should reduce (re)compilation time a bit.


ShFil119:
I'll drop this commit later.

rwgame/GameConfig.hpp
```diff
@@ -0,0 +1,284 @@
+#ifndef RWGAME_GAMECONFIG_HPP
```
ShFil119: @madebr, did I mess with rebasing?

PR #592, opened 2018-08-15
[WIP] Rampages

husho:
Revive of #504
- [x] Figure out why rampage pickups don't work
- [ ] Draw HUD
- [ ] Trigger OnCharacterDie event
- [ ] onVehicleWreck event
- [ ] Model IDs -1 ALL PEDS -2 ALL CARS handle this
- [ ] Segfault on HUD draw?
- [ ] Fixups

blocked by #595 and #594

PR #675, opened 2018-12-17
[WIP] Rampages

christoph-heiss:
Revive of #592.

Though I could start working on this. For now just rebased on current master.

- [ ] Draw HUD
- [ ] Trigger OnCharacterDie event
- [ ] onVehicleWreck event
- [ ] Model IDs -1 ALL PEDS -2 ALL CARS handle this

### DIFF
--- a/cmake_configure.cmake
+++ b/cmake_configure.cmake
@@ -229,7 +229,7 @@ function(openrw_target_apply_options)
         endif()
     endif()
 
-    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND CMAKE_TARGET STREQUAL "Xcode")
         set_property(
             TARGET "${ORW_TARGET}"
             APPEND
--- a/rwengine/src/engine/GameState.cpp
+++ b/rwengine/src/engine/GameState.cpp
@@ -130,6 +130,9 @@ void Rampage::tick(float dt) {
     if (killsRequired <= 0) {
         if (!special) {
             state->gameStats.rampagesPassed++;
+
+            // First rampage rewards 5,000$, second 10,000$ etc.
+            state->playerInfo.money += state->gameStats.rampagesPassed * 5000;
         }
 
         status = Passed;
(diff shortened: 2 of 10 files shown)


rwengine/src/engine/GameState.cpp
```diff
@@ -94,3 +96,85 @@ void GameState::showHelpMessage(const GameStringKey& id) {
...
+Rampage::Status Rampage::getStatus() const {
+    return status;
+}
+
+void Rampage::init(const std::string& gxtEntry, const int32_t weaponID,
+                   const int32_t time, const int32_t kills,
+                   const std::array<int32_t, 4>& modelsToKill,
+                   const bool headshot) {
+    // TODO: Document why
+    if (gxtEntry != "PAGE_00") {
```
husho: There are 2 missions (Trial By Fire, Uzi Money) that use ```init_rampage``` opcode for their own purposes. They pass custom onscreen message, and there is a check if message isn't default then it's special mission rampage initialization, so don't increment rampagesPassed stat, because that's not a real rampage.

christoph-heiss: Good to know, thank you - going to document that appropriately.
Nice to still see you around!

cmake_configure.cmake
```diff
@@ -209,7 +209,7 @@ function(openrw_target_apply_options)
         )
     endif()
 
-    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND CMAKE_TARGET STREQUAL "Xcode")
```
madebr: Does this still work?
```
ORWSRC=# paste path to your source here
mkdir /tmp/orwbin
cd /tmp/orwbin
CC=$(xcodebuild -find clang) CXX=$(xcodebuild -find clang++) cmake "$ORWSRC" "-GUnix Makefiles"
make -j9
```

rwengine/src/engine/GameState.hpp
```diff
@@ -241,6 +243,43 @@ struct ScriptContactData {
     uint32_t baseBrief;
 };
 
+class Rampage {
```
ShFil119: I think `Rampage` should get its own file.

PR #676, opened 2018-12-19
Refactor argument & configuration parsing + split rwgame target in librwgame and rwgame

madebr:
- refactor argument & configuration parsing:
  - the arguments and keys for the configuration file are now defined in `rwgame/RWConfig.inc`
  - this makes the initialization in main more transparent
  - argument parsing can also be tested now
- update coverage collection cmake script:
  - ignore interface libraries + generator expressions in source list
  - add function and branch coverage (when running lcov)
- split the rwgame cmake executable target in librwgame and rwgame
- let the tests link to librwgame

I think `RWConfig` deserves some explanation:
- There are 4 `RWConfigLayer`s, of different priority.
- There is a default layer, a configuration file layer, an argument layer and a manual settings layer (in increasing priority)
- This allows users to have a property in their configuration file, overriding it on the command line and not have it overwritten at save.
- This code can also serve as a base for implementing a settings menu. This settings menu should write to the settings layer.
- The code supports saving configuration file, but is not (yet) used in rwgame. (it is tested)


tests/test_Config.cpp
```diff
@@ -269,36 +269,40 @@ BOOST_AUTO_TEST_CASE(test_TempFile) {
     BOOST_CHECK(!rwfs::exists(path));
 }
 
-BOOST_AUTO_TEST_CASE(test_config_initial) {
+BOOST_AUTO_TEST_CASE(test_configparser_initial) {
```
ShFil119: Maybe extra `_` after `test_config`?

tests/test_Config.cpp
```diff
@@ -269,36 +269,40 @@ BOOST_AUTO_TEST_CASE(test_TempFile) {
...
 
     TempFile tempFile;
     tempFile.append(cfg);
 
-    GameConfig config;
-    config.loadFile(tempFile.path());
+    RWConfigParser cfgParser;
+    RWConfigLayer cfgLayer;
+    RWConfigParser::ParseResult parseResult;
+    std::tie(cfgLayer, parseResult) = cfgParser.loadFile(tempFile.path());
```
ShFil119: You can avoid using `std::tie` using structure bindings.

tests/test_Config.cpp
madebr: No, I test the configparser here (=configuration file parser). There are also tests for the argparser (=argument parser) in there.

tests/test_Config.cpp
madebr: That would have saved me a lot of typing :D

rwgame/RWConfig.cpp
```diff
@@ -0,0 +1,575 @@
...
+};
+
+template <>
+struct Translator<bool> {
+    using internal_type = std::string;
+    using external_type = bool;
+    boost::optional<external_type> get_value(const internal_type &str) {
+        boost::optional<external_type> res;
+        try {
+            res = std::stoi(stripComments(str)) != 0;
```
ShFil119: You can use `lexical_cast` for it: https://github.com/rwengine/openrw/blob/master/rwcore/rw/casts.hpp

rwgame/RWConfig.cpp
```diff
@@ -0,0 +1,575 @@
...
+};
+
+template <>
+struct Translator<int> {
+    using internal_type = std::string;
+    using external_type = int;
+    boost::optional<external_type> get_value(const internal_type &str) {
+        boost::optional<external_type> res;
+        try {
+            res = std::stoi(stripComments(str));
```
ShFil119: `lexical_cast`

rwgame/RWConfig.cpp
```diff
@@ -0,0 +1,575 @@
...
+};
+
+template <>
+struct Translator<float> {
+    using internal_type = std::string;
+    using external_type = float;
+    boost::optional<external_type> get_value(const internal_type &str) {
+        boost::optional<external_type> res;
+        try {
+            res = std::stof(stripComments(str));
```
ShFil119: `lexical_cast`

rwgame/RWConfig.cpp
madebr: I need to detect whether the conversion succeeded.

tests/test_Config.cpp
ShFil119: So camel case?

madebr: camel case will look good here.

tests/test_Globals.cpp
```diff
@@ -1,11 +1,13 @@
...
 
 #if RW_TEST_WITH_DATA
 std::string Global::getGamePath() {
-    GameConfig config;
-    config.loadFile(GameConfig::getDefaultConfigPath() / "openrw.ini");
-    return config.getGameDataPath().string(); //FIXME: use path
+    rwfs::path configPath = RWConfigParser::getDefaultConfigPath() / "openrw.ini";
+    RWConfigLayer cfgLayer;
+    RWConfigParser cfgParser;
+    std::tie(cfgLayer, std::ignore) = cfgParser.loadFile(configPath);
```
ShFil119: I've found that using `[[maybe_unused]]` you can suppress warnings.

madebr: Indeed. Although Visual Studio 2017 does not (currently) support using it in a structured for loop, as in `for([[maybe_unused} auto& [key, value] : map)`. (I've got a compiler error because of this)

rwgame/GameBase.cpp
```diff
@@ -105,6 +31,59 @@ GameBase::GameBase(Logger &inlog, int argc, char *argv[]) :
             [this]() {window.hideCursor();});
 }
 
+RWConfig GameBase::build_config(const std::optional<RWArgConfigLayer> &args) {
```
ShFil119: I'd stay with camel case for functions/methods names.

tests/test_Globals.cpp
madebr: I'll change this and test whether the loading was successful.

rwgame/GameBase.cpp
madebr: aight

rwgame/RWConfig.cpp
madebr: Unless you know of a way to detect conversion failure, I prefer to keep it as it is.

rwgame/GameBase.cpp
```diff
@@ -105,6 +28,59 @@ GameBase::GameBase(Logger &inlog, int argc, char *argv[]) :
...
+        configPath = *args->configPath;
+    } else {
+        configPath = RWConfigParser::getDefaultConfigPath() / "openrw.ini";
+    }
+
+    if ((!args) || (args && !args->noconfig)) {
+        RWConfigLayer configLayer;
+        RWConfigParser::ParseResult parseResult;
+        RWConfigParser configParser{};
+        std::tie(configLayer, parseResult) = configParser.loadFile(configPath);
```
danhedron: Structured binding instead of `std::tie`.

SonofUgly:
Fullscreen no longer works after this. Using -f or --fullscreen causes rwgame to fail to launch.

PR #678, opened 2018-12-24
Various conan/cmake updates 'n fixes

### LINKED ISSUE #595: Game is not responding most of the time on windows
I'm using 089ab587. When I start the game, it takes like 1 minute to get to the menu. After that it's impossible to click on anything. After a while it becomes a little more responsive [still it feels like 0.5fps], so I can click on "start new game".  After starting a new game, it becomes non responsive again, there is only sound, it almost does not render any frames, after alt-tabbing I can't bring game's window back. Task manager reports that the process "Not responding" and the game uses 1 core at 100%.

madebr:
- The `register` warnings in bullet are caused because the `__cplusplus` variable is not defined correctly: https://blogs.msdn.microsoft.com/vcblog/2018/04/09/msvc-now-correctly-reports-__cplusplus
  Fix this by adding the `/Zc:__cplusplus` compile flag
- update travis-ci.org to travis-ci.com
- tweaked cmake find scripts of SDL2 and ffmpeg a bit (they report version now + have conforming indents)
- tweaked the cmake wrapped targets a bit
- ~Merge `RW_FREEBSD`, `RW_NETBSD` and `RW_OPENBSD`: `RW_BSD`~
- set C and C++ standard explicitly as cmake properties. Do something similiar (but lower standard for microprofiler)
- I haven't tested, but I don't think the C++2017 support in Visual Studio 2015 is stellar. So do not generate projects for that Visual Studio version.
- updated conan versions of boost (1.68.0), bullet (2.87, they renamed it), openal (1.19.0), ffmpeg (4.0.2) and qt (5.12.0)
- Debugging on windows (in `Debug` build modus, not `RelWithDebInfo` modus) is currently slow because openrw, along with all dependencies, is built with the standard library containing debug symbols. So add an option `MSVC_NO_DEBUG_RUNTIME` to not use to debug runtime in any configuration type.
  I think this fixes #595.
  It would be nice if @husho can test this (by running the `create_vs_solution.py` script, building openrw and testing it)

Merry christmas! :santa:

### DIFF
--- a/cmake_configure.cmake
+++ b/cmake_configure.cmake
@@ -5,11 +5,6 @@ add_library(rw_checks INTERFACE)
 add_library(openrw::checks ALIAS rw_checks)
 target_link_libraries(rw_interface INTERFACE rw_checks)
 
-# target_compile_features(rw_interface INTERFACE cxx_std_14) is not supported by CMake 3.2
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang")
     target_compile_options(rw_interface
         INTERFACE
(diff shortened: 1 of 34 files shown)


krytarowski:
Please do not merge BSDs. There are various types of BSDs and they are no longer compatible and keep diverging for 20 years.

krytarowski:
If you want a shared ifdef, go for POSIX or UNIX, but not BSD. There are countless examples where one BSD is closer to Linux, other to Solaris, most new interfaces are done differently etc.

madebr:
@krytarowski Thanks for letting us know! I have dropped that patch.

cmake_configure.cmake
```diff
@@ -188,25 +197,36 @@ endforeach()
...
+                    "--mapping_file=${IWYU_MAPPING}"
+            )
+        endif()
+
+        set_target_properties("${ORW_TARGET}"
+            PROPERTIES
+                CXX_EXTENSIONS OFF
+                CXX_STANDARD 17
+                C_EXTENSIONS OFF
+                C_STANDARD 99
```
ShFil119: c11?

external/microprofile/CMakeLists.txt
```diff
@@ -25,4 +25,12 @@ openrw_target_apply_options(
     TARGET microprofile
 )
 
+set_target_properties(microprofile
+    PROPERTIES
+        CXX_EXTENSIONS OFF
+        CXX_STANDARD 11
+        C_EXTENSIONS OFF
+        C_STANDARD 99
```
ShFil119: also

madebr: I was trying to be conservative for microprofiler.

cmake_configure.cmake
madebr: Right now, the only C file are the gl bindings in rwcore.
This change will mean that openrw will be a c++17/c11 project.

external/microprofile/CMakeLists.txt
madebr: And it does not matter. Since microprofiler is a c++ only project.

PR #680, opened 2018-12-26
Further refactor of menu managment

ShFil119:
It also fixes weird behavior which can happen it debug mode on master. (freeze/crash) Caused by #667.

### DIFF
--- a/rwgame/MenuSystem.hpp
+++ b/rwgame/MenuSystem.hpp
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <functional>
+#include <initializer_list>
 #include <memory>
 #include <optional>
 #include <string>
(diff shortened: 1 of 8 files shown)


rwgame/MenuSystem.hpp
```diff
@@ -71,18 +71,14 @@ class Menu {
...
 
-    Menu(std::vector<MenuEntry> initial, int font = MenuDefaults::kFont,
+    Menu(std::initializer_list<MenuEntry>&& initial,
+         const glm::vec2& offset = glm::vec2(), int font = MenuDefaults::kFont,
          float size = 30.f)
-        : activeEntry(-1), font(font), size(size), entries(std::move(initial)) {
-    }
-
-    /**
-     * @brief creates a menu from the given menu items
```
dracc: Please consider adding some kind of comment back.

tests/test_Menu.cpp
```diff
@@ -27,9 +27,7 @@ BOOST_AUTO_TEST_CASE(menu_test_click) {
 
 BOOST_AUTO_TEST_CASE(menu_test_click_offset) {
     bool clickered = false;
```
madebr: nit: clicked?

rwgame/MenuSystem.hpp
ShFil119: Earlier to was wrapper to create ptr of `Menu` class, for optional container, plain constructor is enough.

ShFil119:
@madebr done. ;)

PR #682, opened 2019-01-03
Remove boost::filesystem; Fixes #670

### LINKED ISSUE #670: Remove boost filesystem lib and switch to c++17's
As you maybe know actually it's possible to use filesystem lib of c++17 or boost.
As we recently switched to c++17, now we can remove unneeded dependency and make cmake config a bit simpler.

Florin9doi:


tests/test_Config.cpp
```diff
@@ -92,7 +92,15 @@ class Temp {
 
 private:
     static rwfs::path getRandomFilePath(const rwfs::path &dirname) {
```
madebr: I would replace the generation of random paths with an implementation similar to the following:
```
    static std::string gen_random(size_t len) {
        constexpr std::string_view alphanum =
            "0123456789"
            "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
            "abcdefghijklmnopqrstuvwxyz";
        std::default_random_engine reng(std::random_device{}());
        std::uniform_int_distribution<size_t> dist(0u, alphanum.size());
        std::string res;
        res.reserve(len);
        std::generate_n(std::back_inserter(res), len, [&]() {
            return alphanum[dist(reng)];
        });
        return res;
    }
    static rwfs::path getRandomFilePath(const rwfs::path &dirname) {
        std::string name = "openrw_test_" + gen_random(16);
        return dirname / name;
    }
```
This is more c++ :smiley:

cmake_options.cmake
```diff
@@ -12,8 +12,8 @@ option(TESTS_NODATA "Build tests for no-data testing")
 set(FAILED_CHECK_ACTION "IGNORE" CACHE STRING "What action to perform on a failed RW_CHECK (in debug mode)")
 set_property(CACHE FAILED_CHECK_ACTION PROPERTY STRINGS "IGNORE" "ABORT" "BREAKPOINT")
 
-set(FILESYSTEM_LIBRARY "BOOST" CACHE STRING "Which filesystem library to use")
-set_property(CACHE FILESYSTEM_LIBRARY PROPERTY STRINGS "CXX17" "CXXTS" "BOOST")
+set(FILESYSTEM_LIBRARY "CXX17" CACHE STRING "Which filesystem library to use")
```
madebr: This option can go completely as `CXXTS` does not make sense in c++17.

rwcore/rw/filesystem.hpp
```diff
@@ -3,51 +3,23 @@
```
madebr: This file can be removed as well, I believe.
Or at least replaced by `namespace rwfs = std::filesystem`.

cmake_configure.cmake
```diff
@@ -116,7 +116,6 @@ if(USE_CONAN)
 endif()
 
 if(FILESYSTEM_LIBRARY STREQUAL "CXX17")
```
madebr: This logic can go.

tests/test_Config.cpp
madebr: I guess my suggestion contained a bug.
`std::random_device{}()` can return the same seed if a non-deterministic source is not available.
Try this:
```
    static std::default_random_engine& getRandomEngine() {
        static std::default_random_engine defaultRandomEngine = std::default_random_engine(std::random_device{}());
        return defaultRandomEngine;
    }
    static std::string gen_random(size_t len) {
        constexpr std::string_view alphanum =
            "0123456789"
            "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
            "abcdefghijklmnopqrstuvwxyz";
        std::uniform_int_distribution<size_t> dist(0u, alphanum.size());
        std::string res;
        res.reserve(len);
        std::generate_n(std::back_inserter(res), len, [&]() {
            return alphanum[dist(getRandomEngine())];
        });
        return res;
    }
```

scripts/docker/ubuntu_latest.docker
```diff
@@ -5,15 +5,14 @@ RUN apt-get update \
         build-essential \
         cmake \
         ninja-build \
-        gcc-7 \
-        g++-7 \
-        clang-6.0 \
+        gcc-8 \
```
madebr: `scripts/docker/conan_base.docker` needs the same treatment as this script.

tests/test_Config.cpp
```diff
@@ -7,16 +7,17 @@
 
 #include <fstream>
 #include <map>
+#include <random>
 
 #include <rw/debug.hpp>
-#include <rw/filesystem.hpp>
+#include <filesystem>
```
madebr: nit: you can `filesystem` to the other includes from the standard library.

cmake_options.cmake
madebr: It seems my comments were incorrect. Apple has not yet a `<filesystem>` library. It does have an experimental filesystem library though..

PR #686, opened 2019-01-13
Prevent weapon use while driving

vflyson:


danhedron:
Could you extend this to prevent all weapons except the Uzi being used? I assume this is hardcoded and not a flag somewhere.

vflyson:
Makes sense. I believe also the Detonator should be included in that exception, since I believe it's a weapon as well, but please correct me if I'm wrong.

vflyson:
I think there should be an exposed global state for lookLeft and lookRight from IngameState.cpp, this way I could easily make Uzi fire only if the player's doing a drive-by aiming, but I don't feel familiar/comfortable enough to modify the code that much, I feel like I'd better leave this PR small and leave it up to @danhedron or @ShFil119 to deal with those aspects of the engine.

PR #692, opened 2019-01-17
RWconfig fixes

madebr:
- fixes https://github.com/rwengine/openrw/pull/676#issuecomment-455022268
  bool argument options such as `--invert_y`, `--fullscreen`, ... did not work
- work around boostorg/program_options#72


rwgame/RWConfig.cpp
```diff
@@ -89,39 +81,62 @@ constexpr std::string_view arg_mask_to_key(std::string_view v) {
...
+}
+
+template <>
+std::optional<bool> vm_get_opt_argmask<bool>(const po::variables_map& vm, const std::string_view& argmask) {
+    const std::string key{arg_mask_to_key(argmask)};
+    auto it = vm.find(key);
+    if (it != vm.end()) {
+        return true;
+    }
+    return std::nullopt;
```
ShFil119: nit(?): `false`?
`&& *val;` of line 106 looks quite unneeded.

madebr: No. If it is not present, it can be overridden by a config or default value.

PR #693, opened 2019-01-18
Remove RW_TEST_WITH_DATA from test code

danhedron:
The `RW_TEST_WITH_DATA` flag causes problems:

- Doubles the number of builds done in CI.
- Increases configuration complexity.
- Hides errors when disabled.

This switches to always compiling these tests, disables them by default, and tags them with the "data-test" label.

These tests can still be run by passing `TEST_DATA=1` when configuring.


madebr:
My only gripe with this pr is that I find the syntax `./rwtests --run_test=@data-test` hard to remember.
I envision myself often looking up the exact arguments.
That's something I liked more in madebr/openrw#3: `./rwtests -- --nodata`

Looking at the function `add_label` on the [boost API](https://www.boost.org/doc/libs/1_69_0/libs/test/doc/html/boost/unit_test/master_test_suite_t.html#idp70033296-bb), I think both syntaxes can be made possible.

So, LGTM and if you agree, I'll rework the pr above with the above changes.

PR #696, opened 2019-01-20
Various (smaller changes)

ShFil119:
Commit 07972a7fa76462351689ddff6c2eb5f72446bef2 is temporary(?) solution, planning use weak_ptr + shared_ptrs. It's nicely specifying where weak_ptr should be used. Or if you prefer I can drop this commit. ;)

### DIFF
--- a/rwengine/src/core/Logger.hpp
+++ b/rwengine/src/core/Logger.hpp
@@ -1,6 +1,7 @@
 #ifndef _RWENGINE_LOGGER_HPP_
 #define _RWENGINE_LOGGER_HPP_
 
+#include <array>
 #include <initializer_list>
 #include <string>
 #include <utility>
@@ -13,7 +14,8 @@
  */
 class Logger {
 public:
-    enum MessageSeverity { Verbose, Info, Warning, Error };
+    enum MessageSeverity { Verbose = 0, Info, Warning, Error};
+    static constexpr std::array<char, 4> messageSeverityName{{'V', 'I', 'W', 'E'}};
 
     struct LogMessage {
         /// The component that produced the message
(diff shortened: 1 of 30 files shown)


rwengine/src/core/Logger.hpp
```diff
@@ -13,7 +13,7 @@
  */
 class Logger {
 public:
-    enum MessageSeverity { Verbose, Info, Warning, Error };
+    enum MessageSeverity { Verbose = 'V', Info = 'I', Warning = 'W', Error = 'E' };
```
madebr: I'ld just use a constexpr char-array with an int enum used as indices.

danhedron: The MessageSeverity enum should be in severity order, rather than alphabetical order.

rwengine/src/render/VisualFX.hpp
```diff
@@ -50,7 +50,7 @@ struct ParticleFX final : public VisualFX {
     float lifetime{-1.f};
 
     /** Texture name */
-    TextureData::Handle texture;
+    TextureData* texturePtr = nullptr;
```
madebr: nit: I don't really like the Hungarian notation (using -Ptr as suffix).
I don't think it is used (much) elsewhere.

rwengine/src/engine/GameWorld.cpp
```diff
@@ -116,9 +116,9 @@ bool GameWorld::placeItems(const std::string& name) {
     if (ipll.load(name)) {
         // Find the object.
         for (const auto& inst : ipll.m_instances) {
-            if (!createInstance(inst->id, inst->pos, inst->rot)) {
+            if (!createInstance(inst.id, inst.pos, inst.rot)) {
```
ShFil119: nit: passing just instance `InstanceData`

rwengine/src/render/VisualFX.hpp
ShFil119: Should be ok now.

PR #699, opened 2019-02-09
Fix segfault when failed to create shader object

JorgenPo:
When for some reason a shader object creation returns 0 then there likely will be an error when trying to allocate **sourceBuff** object (because glGetShaderiv with GL_SHADER_SOURCE_LENGTH will fail).
That cause a std::bad_alloc exception and a termination of the program. So I added a check for a shader object creation and now there will be an error message box containing clear message of what was happened.

### DIFF
--- a/rwengine/src/render/OpenGLRenderer.cpp
+++ b/rwengine/src/render/OpenGLRenderer.cpp
@@ -17,6 +17,12 @@ constexpr GLuint kUBOIndexDraw = 2;
 
 GLuint compileShader(GLenum type, const char* source) {
     GLuint shader = glCreateShader(type);
+
+    if (shader == 0) {
+        RW_ERROR("[OGL] Failed to create shader object");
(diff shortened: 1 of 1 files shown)


rwengine/src/render/OpenGLRenderer.cpp
```diff
@@ -17,6 +17,12 @@ constexpr GLuint kUBOIndexDraw = 2;
 
 GLuint compileShader(GLenum type, const char* source) {
     GLuint shader = glCreateShader(type);
+
+    if (shader == GL_ZERO) {
```
madebr: Big nitpick but I think it should be `0` instead of `GL_ZERO` according to `glCreateShader`'s [documentation](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glCreateShader.xhtml). `GL_ZERO` is an enum value that just happens to be 0.

PR #704, opened 2019-04-08
Refactor renderer (and prepare space for librw)

ShFil119:
To do (later?):
- remove openGL from DebugDraw and other places
 - rethink what should be in parent interface ("new GameRenderer")
- prepare that interface 
- refactor usages and "utilization" of polymorphism's pros


tsjost:
Oops this got accidentally automatically closed when I renamed the main branch; github is apparently refusing to reopen/change branch target as the original repo was removed.

Issue #429, opened 2018-04-30
Crash on exit


darkf:
(For reference I'm on Windows, have not tested on other platforms.)

When exiting the game (even before starting a new game, just via the menu) with either the menu or the titlebar close button, the game segfaults with a dereferenced null pointer.

Lihis:
Could not reproduce in Arch Linux, tested against master (464a1e3).

dracc:
Neither I can reproduce on Debian.

ShFil119:
Neither I can reproduce on Arch. 

But using Appveyor's build with wine yes.

madebr:
Cannot reproduce on Windows 10 x64 (neither using appveyor artifact or self-built one).
I can reproduce the crash on wine.

Robo-17:
I can reproduce on Win7 x64.

Matriks404:
I can reproduce it on Windows 10 x64 with self-built build, but not with AppVeyor ones.

ShFil119:
```
>	rwgame.exe!GameRenderer::~GameRenderer() 169	
 	rwgame.exe!RWGame::~RWGame() 133
 	rwgame.exe!main(int argc, const char * * argv) 33	
```
And there's:
```
glDeleteFramebuffers(1, &framebufferName);
```

PR #707, opened 2019-04-20
Close window in parent class (GameBase)

ShFil119:
Should fix: #429


madebr:
Shouldn't we also add `this->close()` to the destructor of `GameWindow`?

PR #708, opened 2019-04-22
Fix conanfile.py on case-insensitive filesystems

cornytrace:
When running `cd build && conan install ..` you get the following error:
`ERROR: Failed requirement 'qt/5.12.0@bincrafters/stable' from 'conanfile.py (openrw/master@None/None)'
ERROR: Requested 'qt/5.12.0@bincrafters/stable' but found case incompatible 'Qt'
Case insensitive filesystem can't manage this`
This pull request fixes the case of Qt so it will work.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -37,7 +37,7 @@ class OpenrwConan(ConanFile):
             'boost/1.68.0@conan/stable',
         ),
         'viewer': (
-            'qt/5.12.0@bincrafters/stable',
+            'Qt/5.12.0@bincrafters/stable',
         ),
         'tools': (
             'freetype/2.9.0@bincrafters/stable',


ShFil119:
Ci works with "qt".
`$ conan install .. -s arch=x86_64`
What returns it for your?

cornytrace:
Same error.
By the way, this is on Windows 10 x64.

cornytrace:
my `conan remote list` looks like this:
`conan-center: https://conan.bintray.com [Verify SSL: True]
bincrafters: https://api.bintray.com/conan/bincrafters/public-conan [Verify SSL: True]`
The strange thing is, that with Qt instead of qt some of the linux build fail...

cornytrace:
So the problem is, there exists both a Qt and a qt conan package. Windows based conan seems to error out because it finds Qt first, and doesn't even have the chance to look at qt.

cornytrace:
This looks like a conan issue, i'll move the issue to there

madebr:
conan-qt switched to lower case qt in december 2018: https://github.com/bincrafters/conan-qt/commit/8e7f9baa553920bec02a19af076d0b8260fa12b4

PR #709, opened 2019-04-28
Remove existing vehicles & peds when creating script objects

danhedron:
Unreferenced objects should be erased if a script is creating an object that would intersect with it.

This fixes the current issue with the intro where 8-ball and the Kuruma are duplicated.

### DIFF
--- a/rwengine/src/engine/GameWorld.cpp
+++ b/rwengine/src/engine/GameWorld.cpp
@@ -1006,31 +1006,9 @@ bool GameWorld::isRaining() const {
 void GameWorld::clearObjectsWithinArea(const glm::vec3 center,
                                        const float radius,
                                        const bool clearParticles) {
-    bool skipFlag = false;
-
     // Vehicles
     for (auto& obj : vehiclePool.objects) {
-        skipFlag = false;
-
-        // Skip if it's the player or owned by player or owned by mission
-        if (obj.second->getLifetime() == GameObject::PlayerLifetime ||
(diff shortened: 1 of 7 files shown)


rwengine/src/engine/GameWorld.cpp
```diff
@@ -1066,6 +1042,28 @@ void GameWorld::clearObjectsWithinArea(const glm::vec3 center,
     // explosions, remove all projectiles
 }
 
+std::vector<GameObject *>
+GameWorld::findOverlappingObjects(const glm::vec3 &center,
+                                  float radius) const {
+    std::vector<GameObject*> overlapping;
+
+    auto checkObjects = [&](const auto& objects) {
```
ShFil119: nit: `constexpr`?

rwengine/src/objects/VehicleObject.cpp
```diff
@@ -643,6 +643,21 @@ float VehicleObject::getVelocity() const {
...
+    }
+
+    for (auto& seat : seatOccupants) {
+        auto character = static_cast<CharacterObject*>(seat.second);
+
+        if (!character->canBeRemoved())
+            return false;
+    }
+
+    return true;
```
ShFil119: nit: it should be doable in one line, something like: `return GameObject::canBeRemoved() && !std::any_of(...)`

rwengine/src/engine/GameWorld.cpp
danhedron: not a constant expression

PR #710, opened 2019-05-01
RE missing stuff from VehicleObject

ShFil119:
Should be nice base for further reimplementing.


ShFil119:
I'll try prepare something working in librwgta and then push into openrw.

PR #713, opened 2019-05-13
Cleanup Atomic & Clump usage

danhedron:
- GameObjects now have either an Atomic or a Clump
- The transform of this object is updated to reflect where the object should be
- Accidentally fixed the player having the wrong model in the intro

### DIFF
--- a/rwengine/src/objects/GameObject.hpp
+++ b/rwengine/src/objects/GameObject.hpp
@@ -2,6 +2,7 @@
 #define _RWENGINE_GAMEOBJECT_HPP_
 
 #include <limits>
+#include <variant>
 
 #include <glm/gtc/quaternion.hpp>
 #include <glm/vec3.hpp>
@@ -28,10 +29,11 @@ class GameObject {
 
     BaseModelInfo* modelinfo_;
 
-    /**
-     * Model used for rendering
-     */
-    ClumpPtr model_ = nullptr;
+    using Model = std::variant<AtomicPtr, ClumpPtr>;
+    Model model_;
+
+    static const AtomicPtr NullAtomic;
+    static const ClumpPtr NullClump;
 
 protected:
     void changeModelInfo(BaseModelInfo* next) {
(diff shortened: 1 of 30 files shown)


rwengine/src/objects/GameObject.hpp
```diff
@@ -92,16 +87,30 @@ class GameObject {
...
+        }
+        return NullAtomic;
+    }
+
+    const ClumpPtr& getClump() const {
+        if (auto clump = std::get_if<ClumpPtr>(&model_))
+        {
+            return *clump;
+        }
+        return NullClump;
```
ShFil119: Maybe it'd better insert static instance of object into body of function? (above return)

rwengine/src/objects/GameObject.hpp
```diff
@@ -92,16 +87,30 @@ class GameObject {
...
 
-    /**
-     * Changes the current model, used for re-dressing chars
-     */
+     const AtomicPtr& getAtomic() const {
+        if (auto atomic = std::get_if<AtomicPtr>(&model_))
+        {
+            return *atomic;
+        }
+        return NullAtomic;
```
ShFil119: also

rwengine/src/objects/PickupObject.cpp
```diff
@@ -271,6 +271,12 @@ PickupObject::PickupObject(GameWorld* world, const glm::vec3& position,
 
     setEnabled(true);
     setCollected(false);
+
+    auto modelData = getModelInfo<SimpleModelInfo>();
+    auto atomic = modelData->getAtomic(0)->clone();
+    atomic->setFrame(std::make_shared<ModelFrame>());
```
ShFil119: Maybe atomic should have method for this. Logic is "outside".

rwengine/src/objects/GameObject.hpp
danhedron: Putting a local static here will generate some guard instructions, unlike a static member. [example](https://godbolt.org/z/M3K8RB).

PR #717, opened 2019-05-22
Use ImGui for debugging

danhedron:
To simplify the code (and make it nicer to use)

- cherry-pick some commits from @madebr which adds ImGui
- Replace the debug view modes with ImGui windows
- Replace the debug tools menu with ImGui

Long-term, with imgui we can:-
- Provide a bootstrap phase to input game path, if we can't auto-detect it.
- Replace rwviewer with a lighter-weight in-game version.

### DIFF
--- a/rwgame/RWImGui.cpp
+++ b/rwgame/RWImGui.cpp
@@ -1,5 +1,9 @@
 #include "RWImGui.hpp"
 
+#include <ai/CharacterController.hpp>
+#include <objects/CharacterObject.hpp>
+#include <objects/VehicleObject.hpp>
+
 #ifdef RW_IMGUI
 
 #include "RWGame.hpp"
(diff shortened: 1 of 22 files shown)


SonofUgly:
Looks nice. Only issue is that you can't exit out of the debug menu if you have a dropdown list open, you have to close it first.

rwgame/RWImGui.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include <glm/glm.hpp>
+
+#include <gl/gl_core_3_3.h>
+#include <glm/gtx/norm.hpp>
+
+namespace {
+void WindowDebugStats(RWGame& game) {
+    auto& io = ImGui::GetIO();
+
+    auto time_ms = 1000.0f / io.Framerate;
```
ShFil119: Imho 22-40 should be moved more into engine. (some opcodes uses it)

rwgame/RWImGui.cpp
```diff
@@ -0,0 +1,204 @@
...
+    static size_t times_index = 0;
+    static double time_average = 0, time_min = 0, time_max = 0;
+    times[times_index++] = time_ms;
+    if (times_index >= average_every_frame) {
+        times_index = 0;
+        time_average = 0;
+        time_min = std::numeric_limits<double>::max();
+        time_max = std::numeric_limits<double>::lowest();
+
+        for (double time : times) {
```
ShFil119: nit: it's float

rwgame/RWImGui.cpp
danhedron: Uses what?

ShFil119: Stats about frames. For example opcode 0078 needs delta time of avg frame.

PR #718, opened 2019-05-23
Fix choosing models for characters

ShFil119:
Before:
old
After:
fix_models


tsjost:
Oops this got accidentally automatically closed when I renamed the main branch; github is apparently refusing to reopen/change branch target as the original repo was removed.

Issue #71, opened 2016-05-24
Water cannot be seen through car windows


TacoDeBoss:
When viewed through a car window, water doesn't appear.

![screenshot_2016-05-23_22-25-17](https://cloud.githubusercontent.com/assets/2870898/15491359/43fa2df6-2135-11e6-892f-a3b7c7520efa.png)

JayFoxRox:
Can you create a better screenshot?
Press `M` while being ingame and then press `F` in that menu to get a free-flight camera (Controlled using mouse, WASD and shift)

TacoDeBoss:
@JayFoxRox Sorry, I've been doing some things the last couple of days.
![screenshot_2016-06-08_19-10-52](https://cloud.githubusercontent.com/assets/2870898/15915188/c1dee0c6-2dac-11e6-8d99-663279d7d5b9.png)

ShFil119:
Still happens on master. 2f8ae7fb0b5c9726c5a0a7d95c321d86ebf550a0

vflyson:
Seems like it's been resolved by #720

PR #720, opened 2019-08-17
Fix: https://github.com/rwengine/openrw/issues/71

alex-leleka:
Now water rendered before world objects.
https://github.com/rwengine/openrw/issues/71

### DIFF
--- a/rwengine/src/render/GameRenderer.cpp
+++ b/rwengine/src/render/GameRenderer.cpp
@@ -235,15 +235,15 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
 
     culled = 0;
 
-    renderObjects(world);
+	renderer->pushDebugGroup("Water");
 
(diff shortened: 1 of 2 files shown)


alex-leleka:
gta_waterfix

rwengine/src/render/GameRenderer.cpp
```diff
@@ -235,15 +235,15 @@ void GameRenderer::renderWorld(GameWorld* world, const ViewCamera& camera,
 
     culled = 0;
 
-    renderObjects(world);
-
-    renderer->pushDebugGroup("Water");
+	renderer->pushDebugGroup("Water");
```
ShFil119: You are breaking here code formatting. (4 spaces should be used instead tab)

Also (new) in line 240, 246.

alex-leleka: Fixed tabs mixing.

PR #724, opened 2019-09-29
Improve player vehicle exit logic

seeseemelk:
The player can now no longer accelerate whilst exiting a vehicle.
This prevents Claude from floating next to the vehicle if the player accelerates during the exit animation.
At the same time, this behaviour is closer to the exit vehicle behaviour in the actual game.


seeseemelk:
Accidentally changed master. Moved commits to new branch and reopened pull request

PR #726, opened 2019-09-30
Improve player vehicle exit logic

seeseemelk:
The player can now no longer accelerate whilst exiting a vehicle.
This prevents Claude from floating next to the vehicle if the player accelerates during the exit animation.
At the same time, this behaviour is closer to the exit vehicle behaviour in the actual game.

### DIFF
--- a/rwgame/states/IngameState.cpp
+++ b/rwgame/states/IngameState.cpp
@@ -259,7 +259,7 @@ void IngameState::tick(float dt) {
         } else if (glm::length2(movement) > 0.001f) {
             if (player->isCurrentActivity(
                     ai::Activities::EnterVehicle::ActivityName)) {
-                // Give up entering a vehicle if we're alreadying doing so
+                // Give up entering a vehicle if we're already doing so
                 player->skipActivity();
             }
         }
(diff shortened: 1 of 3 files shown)


rwgame/states/IngameState.cpp
```diff
@@ -259,14 +259,24 @@ void IngameState::tick(float dt) {
...
 
         if (player->getCharacter()->getCurrentVehicle()) {
             auto vehicle = player->getCharacter()->getCurrentVehicle();
             vehicle->setHandbraking(held(GameInputState::Handbrake));
+            if (player->isCurrentActivity(
+                    ai::Activities::ExitVehicle::ActivityName)) {
+                // The player cannot accelerate while exiting.
+                // He can, however, brake in the opposite direction of movement.
+                int velocitySign = vehicle->getVelocity() >= 0 ? 1 : -1;
+	            int movementSign = movement.x >= 0 ? 1 : -1;
```
Lihis: This and lines below have some tab characters, indent with spaces only.

Lihis (changes requested):

seeseemelk:
I replaced those tab character

Lihis (approved):

okias:
could you squash these commits?

Reviewed and Tested by.

PR #733, opened 2020-03-16
audio/SoundSource.hpp: Add missing include

pcercuei:
Fix build on Debian testing by adding a missing <vector> include.

Signed-off-by: Paul Cercueil <paul@crapouillou.net>


madebr:
@pcercuei
Thank you for the patches!!

PR #734, opened 2020-05-15
Improve file loading for .img files.

christoph-heiss:
Hi again!

So while working on openrw/debugging an issue, I really started to notice the significant loading time until being in-game.

I did some perf'ing and noticed that `LoaderIMG::load` was a pretty hot path and after digging a bit, I found out that `LoaderIMG::load` was called 3237(!) times with `models/gta3.img`.
That meant opening and reading the file and allocating memory for it each call.

I did some simple optimization, now each file is only loaded once by `LoaderIMG`, which resulted in a pretty nifty speedup. 
Also after these changes, `LoaderIMG::load` completely vanishes from the flamegraph.

On a hot file cache, loading on startup takes ~7600ms on `master` on my machine, after the changes it goes down to ~5100ms.
There might be even more loading time reduction on slower systems and/or spinning rust.

Only tested on Linux, so I’d much appreciate if someone could test this on Windows to check if really nothing breaks.

### DIFF
--- a/rwcore/loaders/LoaderIMG.cpp
+++ b/rwcore/loaders/LoaderIMG.cpp
@@ -1,53 +1,87 @@
-#include "loaders/LoaderIMG.hpp"
+#include <loaders/LoaderIMG.hpp>
 
-#include <boost/algorithm/string/predicate.hpp>
-#include <cstdio>
+#include <cassert>
+#include <cctype>
+#include <cstring>
+#include <algorithm>
 
-#include "rw/debug.hpp"
+#include <rw/debug.hpp>
+
+namespace {
+
+constexpr size_t kAssetRecordSize{2048};
+
+void to_lowercase_inplace(char* name) {
+    size_t len = std::strlen(name);
+
+    std::transform(
+        name, name+len, name,
+        [](char ch) -> char { return std::tolower(ch); }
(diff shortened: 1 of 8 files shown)


mole99:
I tested this PR on Linux and got a nice speedup:

| Master | PR |
|:-:|:-:|
| 3283 ms | 2509 ms |
| 3280 ms | 2460 ms |
| 3272 ms | 2500 ms |

That's a decrease of around 24% in loading time. Awesome!

madebr:
@christoph-heiss 
These changes will make the `LoaderIMG` thread unsafe.
So maybe add a note at the top of the `LoaderIMG` class that the class is not thread safe and then this is good to merge.

christoph-heiss:
> So maybe add a note at the top of the LoaderIMG class that the class is not thread safe and then this is good to merge.

Good suggestion, just pushed another commit which adds an appriopriate note to the class.

And if sometimes in the future the usecase arrives that we need to use this from multiple threads, making it MT-safe is fairly easy.

madebr:
Thanks @christoph-heiss !

> Good suggestion, just pushed another commit which adds an appriopriate note to the class.

This suggestion was made by @danhedron on the irc channel.

PR #735, opened 2020-05-16
Fix Travis-CI builds and tests/Config

christoph-heiss:
After sending in my latest PR, I noticed that our travis-ci was failing for three builds.
I did some investigating to fix those issues, pretty much just some regularly maintenance work was done.
One probably noteworthy thing is that I disabled some features of ffmpeg for the conan build - we don't use most of the codecs anyway and since ffmpeg needs to be build from source, this should speed up the CI rebuilds.
Especially RFC for the conan changes, it's my first time working with it.

Additionally, while compile-testing I noticed that `tests/test_Config.cpp` was failing to compile with `FILESYSTEM_LIBRARY` set to `CXX17` due to boost not adhering to the standards specification completely, and `unique_path` not being supported in CXX17's std::filesystem due to security concerns, so I replicated the functionality in some other way.

While working on this, I wondered if it still makes sense to support the `FILESYSTEM_LIBRARY=CXXTS` configuration? This implementation has been pretty much abandoned with recent compilers and even fails to compile on `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,13 +28,21 @@ matrix:
   - os: osx
     compiler: clang
     env: NAME="Apple macOS" NAME_SUFFIX="mac"
-    osx_image: xcode10.1
+    osx_image: xcode11.4
+    addons:
+      homebrew:
+        packages:
+          - boost
+          - cmake
+          - bullet
+          - ffmpeg
+          - glm
+          - openal-soft
+          - qt5
+          - sdl2
+          - jack
+          - freetype
     install:
-      - brew update
-      - /usr/bin/yes | pip2 uninstall numpy  # see https://github.com/travis-ci/travis-ci/issues/6688
-      - brew upgrade python
-      - brew upgrade
-      - brew install boost cmake bullet ffmpeg glm openal-soft qt5 sdl2 jack freetype
       - export PATH="/usr/local/opt/qt/bin:$PATH"
     script:
       - mkdir -p "$TRAVIS_BUILD_DIR/build"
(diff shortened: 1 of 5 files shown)


madebr:
> Especially RFC for the conan changes, it's my first time working with it.

These look good.
Modifying the `default_options` dict still allows consumers to override it.

> While working on this, I wondered if it still makes sense to support the `FILESYSTEM_LIBRARY=CXXTS` configuration? This implementation has been pretty much abandoned with recent compilers and even fails to compile on `master`.

The final goal is to drop boost::filesystem. (see #670 and #682)
So dropping workarounds for boost is completely fine.
Does Macos' apple-clang have a good filesystem implementation?

I'm totally fine with going full c++17 filesystem and dropping TS & boost.
Should the cmake script check for a working standard c++ library?

christoph-heiss:
Thanks for the quick review!
Okay, so CXXTS-fs can go - I will do a PR sometime soon to rip that.

> Does Macos' apple-clang have a good filesystem implementation?

I can check that. (Maybe also the various *BSDs we support, shouldn't forget these.)

> Should the cmake script check for a working standard c++ library?

Sounds like a nice idea, I'll try this. 

> I'm totally fine with going full c++17 filesystem and dropping TS & boost.

If we do, it will definitely raise the minimum compiler version. But that's probably fine since OpenRW probably already depends on a somewhat recent compiler.
Also, it would definitely benefit the maintainability a lot since all the fs-stuff is *insanely fragile*, seeing how often it breaks.

christoph-heiss:
Ping?
(I'd put all the fs-stuff discussed here in another PR for clear separation.)

madebr:
This pr looks good and is ok to merge.

But I think we have to face the fact that this project is dead.
@aap has succeeded in (almost) completely reverse engineering gta3 and has now begun with vice city.

See https://github.com/gtamodding/re3 for that work.

madebr (approved):

PR #740, opened 2021-06-23
Fix linux build error - size_t undefined - sleep_for() not a member of std::this_thread

Un1Gfn:


Un1Gfn:
`sleep_for()` from `<future>` emits the error

`sleep_for()` from `<thread>` does not

Output is playable after applying this patch

AUR [comment](https://aur.archlinux.org/packages/openrw-git/#comment-814129)

Original [patch](https://gist.github.com/Un1Gfn/a0e8e5e1be51b41c1af09eccf12d121a#file-0001-fix-linux-build-error-patch)

madebr (approved):

tsjost:
Great, thank you!

PR #746, opened 2023-11-01
Fixes compatibility with newer ffmpeg versions.

Dygear:
Fixes `Error: assigning to 'AVCodec *' from 'const AVCodec *' discards qualifiers`.

Building on macOS Sonoma 14.1 (23B74), using a M2 Max in an `arch -x86_64 zsh` environment.

Fixes the error by using the auto keyword to infer the correct type.

It should be noted that this *might* not be the best way to handle this. Indeed right after this, the compiler notes that the `codec` variable is never actually used in either location. Perhaps the author of the [original commit for this line](https://github.com/rwengine/openrw/commit/50c6eedf4f48bcbf749f7e0bc50b42be33c0e345), Filip Gawin, could explain what the intention was with this line. Removing this line completely also fixes the error, but I'm unable to tell if it has side effects.

### DIFF
--- a/rwengine/src/audio/SoundSource.cpp
+++ b/rwengine/src/audio/SoundSource.cpp
@@ -121,7 +121,7 @@ bool SoundSource::findAudioStream(const std::filesystem::path& filePath) {
     }
 
     audioStream = formatContext->streams[streamIndex];
-    auto codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
     return true;
 }
 
--- a/rwengine/src/audio/SoundSource.cpp
+++ b/rwengine/src/audio/SoundSource.cpp
@@ -121,7 +121,7 @@ bool SoundSource::findAudioStream(const std::filesystem::path& filePath) {
     }
 
     audioStream = formatContext->streams[streamIndex];
-    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+    auto codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
     return true;
 }
 
@@ -148,7 +148,7 @@ bool SoundSource::findAudioStreamSfx() {
     }
 
     audioStream = formatContext->streams[streamIndex];
-    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+    auto codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
 
     return true;
 }
--- a/rwengine/src/audio/SoundSource.cpp
+++ b/rwengine/src/audio/SoundSource.cpp
@@ -148,7 +148,7 @@ bool SoundSource::findAudioStreamSfx() {
     }
 
     audioStream = formatContext->streams[streamIndex];
-    auto codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
 
     return true;
 }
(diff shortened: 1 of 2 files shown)


rolandpoulter:
+1
had the same error in the description and this fixed it

rwengine/src/audio/SoundSource.cpp
```diff
@@ -121,7 +121,7 @@ bool SoundSource::findAudioStream(const std::filesystem::path& filePath) {
     }
 
     audioStream = formatContext->streams[streamIndex];
-    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
+    auto codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
```
Lihis, line 124: This shadows the [SoundSource::codec](https://github.com/rwengine/openrw/blob/0f83c16f6518c427a4f156497c3edc843610c402/rwengine/src/audio/SoundSource.hpp#L93).

I think the [SoundSource::codec](https://github.com/rwengine/openrw/blob/0f83c16f6518c427a4f156497c3edc843610c402/rwengine/src/audio/SoundSource.hpp#L93) should be changed to `const AVCodec*` as I'm pretty sure shadowing it only causes other issues.

Lihis: Yes I think the `SoundSource::codec` should be made `const`.

Lihis (changes requested):

rwengine/src/audio/SoundSource.cpp
Dygear: So `const codec` instead of `auto codec`?

Lihis: No because that still shadows the member variable.

The `SoundSource::codec`: https://github.com/rwengine/openrw/blob/0f83c16f6518c427a4f156497c3edc843610c402/rwengine/src/audio/SoundSource.hpp#L93

should be:

```cpp
const AVCodec* codec = nullptr;
```

Dygear: Do you want me to make that change and commit it into this pr?

Lihis: Yes commit it to this PR. And I think you can drop the original commit.

Dygear:
Ok took your suggestion @Lihis, hopefully this can be closed now once merged.

rwengine/src/audio/SoundSource.cpp
```diff
@@ -148,7 +148,7 @@ bool SoundSource::findAudioStreamSfx() {
...
```
Lihis, line 151: From here the `auto` should also be removed because it shadows the member variable:

```suggestion
    codec = avcodec_find_decoder(audioStream->codecpar->codec_id);
```

Lihis (changes requested):
Otherwise looks good

Dygear:
@Lihis Done

Lihis (approved):

tsjost:
Thanks for the fix and review!

This fixes the build error, but Start new game soft-hangs with audio still playing; not yet sure if this also has something to do with newer ffmpeg, or just something weird with my machine.

Lihis:
> Thanks for the fix and review!
> 
> This fixes the build error, but Start new game soft-hangs with audio still playing; not yet sure if this also has something to do with newer ffmpeg, or just something weird with my machine.

At least PR #747 fixes the hang for me.

Lihis:
> Thanks for the fix and review!
> 
> This fixes the build error, but Start new game soft-hangs with audio still playing; not yet sure if this also has something to do with newer ffmpeg, or just something weird with my machine.

@tsjost at least the PR #747 fixes the hang for me.

PR #750, opened 2024-11-26
ffmpeg 7.1 support

okias:
Works. There are issues with the tests, but that's unrelated to the fixes.

### DIFF
--- a/rwengine/src/audio/SoundSource.cpp
+++ b/rwengine/src/audio/SoundSource.cpp
@@ -10,13 +10,6 @@ extern "C" {
 #include <libswresample/swresample.h>
 }
 
-// Rename some functions for older libavcodec/ffmpeg versions (e.g. Ubuntu
-// Trusty)
-#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(55, 28, 1)
-#define av_frame_alloc avcodec_alloc_frame
-#define av_frame_free avcodec_free_frame
-#endif
-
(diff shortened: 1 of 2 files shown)


rwengine/src/audio/SoundSource.cpp
```diff
@@ -377,10 +372,31 @@ void SoundSource::decodeAndResampleFrames(const std::filesystem::path& filePath,
             while ((receiveFrame =
                         avcodec_receive_frame(codecContext, frame)) == 0) {
                 if (!swr) {
+#if !HAVE_CH_LAYOUT
```
dracc, line 375: Here you're doing
```c
#if !HAVE_CH_LAYOUT
...
#endif
#if HAVE_CH_LAYOUT
...
#else
```
which could be baked together into a single `#if` as far as I can tell.

Lihis:
> This PR fixes building on my machine running Debian Testing ("Trixie"). However, debug mode is silent and "Start New Game" gives some traffic noise but no visuals. Given that building is broken without these fixes, I'm not sure what the status is on current `master`.

The PR branch is missing e20356b05fd3ab0eb7ca376a901fa222173046bc, so rebasing against `main` should fix the starting of new game.

okias:
@dracc can you confirm it's regression against previous state without the PR?

dracc (approved):
Since the rebase this builds and runs as expected on my machine. Thank you!

PR #751, opened 2024-11-26
core: make findFilePath more robust

okias:
Do not crash, when asset is missing, just warn about it. Game can work well without missing wav and mp3 files.

Also, the crash itself looks a bit mysterious, so now the game prints name of the missing asset.

// this caused pretty headache because I didn't copied the audio files from CD2...

### DIFF
--- a/rwcore/platform/FileIndex.cpp
+++ b/rwcore/platform/FileIndex.cpp
@@ -39,12 +39,22 @@ void FileIndex::indexTree(const std::filesystem::path &path) {
 }
 
 const FileIndex::IndexedData *FileIndex::getIndexedDataAt(const std::string &filePath) const {
-    auto normPath = normalizeFilePath(filePath);
-    return &indexedData_.at(normPath);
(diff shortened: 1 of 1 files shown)


rwcore/platform/FileIndex.cpp
```diff
@@ -39,12 +39,22 @@ void FileIndex::indexTree(const std::filesystem::path &path) {
...
+    }
 }
 
 std::filesystem::path FileIndex::findFilePath(const std::string &filePath) const {
-    return getIndexedDataAt(filePath)->path;
+    auto idxData = getIndexedDataAt(filePath);
+    if (!idxData)
+	return filePath;
+
+    return idxData->path;
```
Lihis, line 57: Try to follow existing code style. We also have [.clang-format](https://github.com/rwengine/openrw/blob/f10a5a8f7abc79a0728847e9a10ee104a1216047/.clang-format) config to format the code with `clang-format`.

Lihis (changes requested):

PR #757, opened 2024-12-11
rwcore: we don't support loading of DXT1, 3, or 5 assets

okias:
Shouldn't be hard to implement, thou warn, as in GTA 3 assets texNative.rasterformat may be 0x0300, which imply RGBA4444, but also DXT3 with dxttype flag set.


okias:
I implemented simple DXT support in !758 . Closing

Issue #379, opened 2018-01-30
Dynamic weather conditions


danhedron:
The weather system currently does not change weather.

The original game has a hard-coded set of weather conditions that are cycled through [seen here](https://www.gtamodding.com/wiki/Time_cycle#Weather_lists).

The behaviour of some opcodes related to weather need to be fixed to set the forced weather condition, rather than interrupting the dynamic weather.

Weather conditions are part of the Basic state, which is also loaded from save files.

Further information is appreciated.

Lihis:
In IRC @JayFoxRox brought up that opcode 01B5 is used only once in GTA III; in "Blowfish" mission. I did some testing and effect of the 01B5 can been seen only with correct timing. The 01B5 is triggered to force weather to sunny when you park the Trasher in the marker between the two canisters and weather is released with opcode 01B7 after the factory has been blown up or after failing the mission.

Consider consider following weather scenario:

12:00 PM: Sunny
13:00 PM: Sunny (starts transition to foggy)
14:00 PM: Foggy
15:00 PM: Foggy (starts transition to sunny)
16:00 PM: Sunny

To see the effect of the 01B5 you should:

13:00 PM: Sunny (starts transition to foggy)
13:57 PM: `Park Trasher to marker between the two canisters`
14:00 PM: Foggy (starts transition to sunny)
14:05 PM: `Trigger the bomb and leave the Trasher`
15:00 PM: Sunny (starts transition to foggy)
16:00 PM: Foggy (starts transition to sunny)
17:00 PM: Sunny

I have started to implement (initial version) of dynamic weather, I'll try to prep PR in following days.

Lihis:
As @JayFoxRox kindly noted that we have more game breaking issues that would need attention, I'm postponing my work (which still needs some research) on this until this issue becomes relevant. Current implementation at https://github.com/Lihis/openrw/pull/2.

PR #759, opened 2025-01-05
Initial dynamic weather support

Lihis:
This has been work-in-progress (see Lihis/openrw#2) for a while and I will not have time to finish any time soon so I think it might be better to get this merged in its current state so someone can improve it at some point.

I have not tested this against latest `main` but I'll try to do it within few weeks or so.

Copy-paste from PR in my fork of the repo:

------

Tasklist and some details:

- [x] Weather cycles according the [weather list](https://gtamods.com/wiki/Time_cycle#Weather_lists) for GTA III.
- [ ] Verify is the above weather list same for:
  - [ ] v1.1
  - [ ] Steam version
- [ ] Rain droplets in rainy weather
- [ ] Weather can be controlled by opcodes and it behaves according them:
  - [ ] [01B5](https://gtamods.com/wiki/01B5): Force next weather (see https://github.com/rwengine/openrw/issues/379#issuecomment-386902620). Weather does not cycle until released (via 01B7).
  - [x] [01B6](https://gtamods.com/wiki/01B6): Force weather now; weather is immediately set to requested weather. Weather does not cycle until released (via 01B7).
  - [x] [01B7](https://gtamods.com/wiki/01B7): Release weather; continue weather cycle from where it was.
  - [0251](https://gtamods.com/wiki/0251): Store weather; unused.
  - [0251](https://gtamods.com/wiki/0252): Restore weather; unused.
- [ ] New game:
  - [ ] Weather cycle starts from array index 0?
- [ ] Load game:
  - [ ] Clean save without weather cheats; weather cycle continues from the saved state.
  - [ ] "Dirty" save where a weather cheat used; weather does not cycle, weather is kept in the forced state.
- [ ] Write simple SCM for testing opcode 01B5.
- [ ] Did miss something? Yes/No: TBD

### DIFF
--- a/rwengine/src/data/Weather.hpp
+++ b/rwengine/src/data/Weather.hpp
@@ -40,6 +40,14 @@ public:
                       float a, float tod);
 
     std::vector<Entry> entries;
+
+    // Taken from: https://www.gtamodding.com/wiki/Time_cycle#Weather_lists
+    // TODO: This weather list applies only for GTA III
+    const uint16_t WeatherList[64] = {
+            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2, 2, 1, 0,
+            0, 0, 1, 3, 3, 1, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 2,
+            2, 1, 2, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3, 3, 0, 0, 0, 2, 1
+    };
 };
 
 #endif
--- a/rwgame/RWGame.cpp
+++ b/rwgame/RWGame.cpp
@@ -533,6 +533,7 @@ void RWGame::tick(float dt) {
     static float clockAccumulator = 0.f;
     static float scriptTimerAccumulator = 0.f;
     static ScriptInt beepTime = std::numeric_limits<ScriptInt>::max();
+    static uint8_t prevGameHour = state.basic.gameHour;
     if (currState->shouldWorldUpdate()) {
         world->chase.update(dt);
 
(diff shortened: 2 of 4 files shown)


rwengine/src/data/Weather.hpp
```diff
@@ -40,6 +40,14 @@ class Weather {
                       float a, float tod);
 
     std::vector<Entry> entries;
+
+    // Taken from: https://www.gtamodding.com/wiki/Time_cycle#Weather_lists
+    // TODO: This weather list applies only for GTA III
+    const uint16_t WeatherList[64] = {
```
okias, line 46: maybe the array could include the fact it's GTA3 specific? prefix or postfix with game name or game codename? What do you think?

rwgame/RWGame.cpp
```diff
@@ -554,6 +555,22 @@ void RWGame::tick(float dt) {
             clockAccumulator -= 1.f;
         }
 
+        if (prevGameHour != state.basic.gameHour) {
```
okias, line 558: what about making it inline function, so the `tick` doesn't get so huge when all the rules get implemeneted?

okias (commented):
look like a good start

rwengine/src/data/Weather.hpp
Lihis: For now I would just leave the TODO there; the consensus has been (or was couple years back when the project was more active) that we should prioritize getting GTA III into playable state.

I hope that someone who implements the VC/SA part, refactors this properly.

okias: I would love to spend time (at some point) to getting miami working.
May I propose special mark in comment? maybe `gta3-specific`?

 Would be nice to have +- track (`grep`able) of all the places, where the engine needs to be adjusted.

PR #761, opened 2025-01-08
Migrate Travis CI to GitHub Actions

### LINKED ISSUE #760: Continuous Integration broken
It looks that Travis CI has ran out of credits so we're lacking CI builds for Linux.

Can we continue to use Travis CI without huge financial requirements or do we need to look for replacement?

Some replacements which comes to my mind would be:

- GitHub Actions
    - Might be possible to do Mac and Windows builds so we could drop the AppVeyor and have everything in one place
- CircleCI
    - 6000 minutes on free plan. Some years ago they did not have Windows support so current status (and Mac support) should be checked

-----

AppVeyor seems broken too but I would keep this issue only about Travis CI as AppVeyor "works"; yes the build does not pass but it does not need credit card to make it pass.

Lihis:
Some notes:

- Code coverage still needs work to be enabled
- MacOS pipeline needs to be done

Closes #760


okias:
```
Acked-by: David Heidelberg <david@ixit.cz>
```
quickly checked, haven't found anything wrong :wink:

Lihis:
> I'll create issues for getting code coverage enabled and making build for MacOS once the PR is merged.
> 
> PR #750 is needed for the compile to succeed; tho only on Ubuntu all tests passes, Arch Linux and Fedora has some failures which needs to be fixed.

Issues created:

- #763 
- #764 
- #765

